PR #4, opened 2020-11-11
Load dataset using rasterio. Add info method.

adehecq:
Add steps to __init__ to load the rasterio dataset and copy specific attributes. Add the info method.

### DIFF
--- a/GeoUtils/raster_tools.py
+++ b/GeoUtils/raster_tools.py
@@ -1,10 +1,39 @@
+import numpy as np
 import rasterio as rio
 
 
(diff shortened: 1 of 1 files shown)


iamdonovan (dismissed):

iamdonovan (approved):

atedstone (changes requested):
Please add an rst docstring to __init__ and info methods. ;)

atedstone (approved):
Alright, I'll let this one through for now. But we still need those docstrings! :-P

PR #6, opened 2020-11-12
Iamdonovan

iamdonovan:
### raster_tools.py:
- Raster.load() to read from disk
- docstrings for **Raster**.__init__(), **Raster**.info(), and **Raster**.load()
- option to calculate statistics for each band in **Raster**.info()
- started **SatelliteImage** class

### vector_tools.py:
- started **Vector** class

### __init__.py:
added import statements for both raster_tools and vector_tools


atedstone (approved):

adehecq (approved):
- self.nbands is redundant with self.count
- shall we use self.r or self.data (the former being shorter)?

PR #8, opened 2020-11-12
Nicer implementation of info printing, plus some other minor bits

atedstone:
Add __str__ function so that print(myRaster) yields the output of myRaster.info().
myRaster.info() now provides a string (rather than printing directly).

Added a dummy __repr__ function, which definitely needs revisiting once we've fleshed out the rest of the class.

### DIFF
--- a/GeoUtils/raster_tools.py
+++ b/GeoUtils/raster_tools.py
@@ -16,24 +16,40 @@ class Raster():
         self.ds = ds
         
         # Copy most used attributes/methods
+        self._saved_attrs = saved_attrs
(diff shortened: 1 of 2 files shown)


iamdonovan (changes requested):
can you update this with the current version of info(), with stats implemented?

atedstone:
merged the conflicts now. I promise to work next on something that shouldn't conflict!

iamdonovan (approved):
beautiful.

adehecq (approved):

PR #10, opened 2020-11-12
Save to file

atedstone:
Add function to save raster to file.

### DIFF
--- a/GeoUtils/raster_tools.py
+++ b/GeoUtils/raster_tools.py
@@ -122,5 +122,45 @@ class Raster():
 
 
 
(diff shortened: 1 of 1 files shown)


adehecq (approved):
Would be good to add the options to:
- set (update) the nodata value
- add metadata
- have different creation options, passed as dict (e.g. {'COMPRESS':'LZW', 'BLOCKXSIZE':256}) etc)

atedstone:
@adehecq all great ideas for addition, I'll try to get to work on them shortly.

iamdonovan (approved):
I even got it to work on my machine!

PR #11, opened 2020-11-12
Crop

iamdonovan:
## changes

- re-named saved_attrs to default_attrs
- changed argument __init__(saved_attrs=saved_attrs) to attrs=None. If no additional attributes are passed, will only load default attributes.
- created _read_attributes(), which will copy self._saved_attrs as before, but will also read additional attributes.
- fleshed out crop(), which will now update the object to the new geometry.

## still to-do
Need to double-check how rasterio.mask.mask is calculating the new boundaries.

### DIFF
--- a/GeoUtils/raster_tools.py
+++ b/GeoUtils/raster_tools.py
@@ -51,7 +51,7 @@ class Raster(object):
         # read the file as a rasterio dataset
         self.ds = self.memfile.open()
 
-        self._read_attrs(attrs)
+        self._read_attrs(list(attrs))  # cast as a list to prevent iteration weirdness
 
         if load_data:
             self.load(bands)
@@ -83,6 +83,7 @@ class Raster(object):
             for attr in default_attrs:
                 if attr not in attrs:
                     attrs.append(attr)
+            self._saved_attrs = attrs
 
         for attr in attrs:
             setattr(self, attr, getattr(self.ds, attr))
--- a/GeoUtils/raster_tools.py
+++ b/GeoUtils/raster_tools.py
@@ -174,7 +174,7 @@ class Raster(object):
         if isinstance(cropGeom, Raster):
             xmin, ymin, xmax, ymax = cropGeom.bounds
         elif isinstance(cropGeom, vt.Vector):
-            pass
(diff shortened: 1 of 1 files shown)


adehecq (approved):

adehecq:
You may want to allow tuples as well in method crop, rather than just lists.

GeoUtils/raster_tools.py
```diff
@@ -70,6 +74,36 @@ def __str__(self):
...
 
+    def _read_attrs(self, attrs=None):
+        # Copy most used attributes/methods
+        if attrs is None:
+            self._saved_attrs = default_attrs
+            attrs = default_attrs
+        else:
+            if isinstance(attrs, str):
+                attrs = [attrs]
+            for attr in default_attrs:
```
atedstone, line 85: IIRC, this forces all default_attrs to be added to the Raster even when a list of them is provided by the user. Not sure that this is desirable behaviour?

atedstone (commented):
This looks great overall - just a comment on default attrs to consider.

iamdonovan:
> You may want to allow tuples as well in method crop, rather than just lists.
> Would be good to also explain what is the behavior when the crop bounds are not a multiple of res. It seems like it is extended (ceiling).

I will add tuples (and other iterables?) later tonight. Will also keep looking into the crop behavior - I may end up re-writing it, as I'm not completely sure I'm happy right now.

GeoUtils/raster_tools.py
iamdonovan: That is correct. The reason being, we are assuming that a number of default attributes are in place for other functions, such as info() - if they're not set, info() won't work - and crop() may not either as currently implemented.

atedstone: OK, I see. In which case it would be worth noting this in the docstring for the class and/or __init__ - something along the lines of, 'all the defaults are always loaded, if you want to add any other attributes then list them here'

adehecq (approved):

atedstone (approved):
I've reviewed, but haven't specifically tested the functionality of this commit.

PR #13, opened 2020-11-12
Add a from_array() class method.

atedstone:
This method allows to create a Raster from a data array with accompanying transform and CRS information.

To complete this neatly I had to modify __init__ to detect the case when this classmethod is called.

Also, load_data now ==True by default, else this method would not have the desired behaviour. Anyway, I reckon that load_data=True by default is probably more user friendly for most applications ;)


adehecq (approved):

adehecq (approved):

iamdonovan (changes requested):
please add a newline at the end of the file so pycharm stops yelling at me. :)

atedstone:
@iamdonovan, your wish has been granted

iamdonovan (approved):

PR #20, opened 2020-11-13
Fix circular dependencies bug

atedstone:
Both raster_tools and vector_tools were attempting to load each other into their module namespaces, resulting in circular dependencies.

Now, the relevant module is only imported within the scope of the method which needs it.


rhugonnet (approved):
neat fix

adehecq (approved):

PR #21, opened 2020-11-13
Add intersection method

adehecq:
**Summary**
- add a proj_tool library with functions to handle different projections etc. Need to be common to raster and vector classes
- in proj_tools, add a function to create a Polygon from a bounds
- in Raster, add an intersection method to return the extent of intersection between two rasters
- add Emacs backup files to gitignore

**Left to do**
- handle different CRS
- add a tool to check if two CRS are the same

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -132,4 +132,7 @@ dmypy.json
 # Pyre type checker
 .pyre/
 
+# Emacs backup files
+*~
+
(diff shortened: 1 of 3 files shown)


atedstone (approved):

iamdonovan (approved):
looks fine, though I haven't had a chance to test yet...

PR #22, opened 2020-11-13
Minor change + some mess up

adehecq:
- fix a mess up due to conflicts with previous commits/branches. Some changes had been reverted, hence the many commits.
- fix an issue with a missing import

### DIFF
--- a/GeoUtils/proj_tools.py
+++ b/GeoUtils/proj_tools.py
@@ -4,6 +4,8 @@ GeoUtils.proj_tools provides a toolset for dealing with different coordinate ref
 import rasterio as rio
 from rasterio.crs import CRS
 from shapely.geometry.polygon import Polygon
+from shapely.ops import transform
+import pyproj
 
 
 def bounds2poly(boundsGeom, in_crs=None, out_crs=None):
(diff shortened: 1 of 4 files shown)


GeoUtils/proj_tools.py
```diff
@@ -4,6 +4,8 @@
 import rasterio as rio
 from rasterio.crs import CRS
 from shapely.geometry.polygon import Polygon
+from shapely.ops import transform
```
atedstone, line 7: This is a confusing import given that `transform` elsewhere in the repo commonly refers to an Affine transform. Please consider referencing as shapely.ops.transform instead.

atedstone (changes requested):
See line-specific comment.

GeoUtils/proj_tools.py
adehecq: Done!

adehecq:
Also accidentally added to the PR a newer commit that adds a method get_bounds_projected to Raster.

atedstone (approved):

iamdonovan (approved):

PR #36, opened 2020-11-17
Renaming

atedstone:
In all code, GeoUtils is now geoutils, and module names have had the hyphens removed. This makes the package more inline with the likes of geopandas, xarray etc - fewer keystrokes needed.


iamdonovan (approved):

adehecq (commented):
While we're talking naming conventions, I quite liked the georaster, geovector compared to rastertools/vectortools. It could be a time to change it, but I have no strong preference for it.

adehecq (approved):

atedstone:
speak now or forever hold your peace...........I've no strong preference either, but this is probably the best and/or near-last time to make such a large change, before we starting using the package properly.

iamdonovan:
I still don't have strong preferences either way. I use things like "X_tools" because I lack imagination, not because I have strong feelings about it. :)

adehecq:
Then I vote for georaster/geovector because 1) it's more explicit that we're dealing with georeferenced data and 2) I almost don't need to update my scripts ;-)

atedstone:
your wish is my command...

atedstone:
I think that I've made all necessary changes. Merging and closing this PR but do keep an eye out for problems, this is where we need the fabled testing framework!

PR #37, opened 2020-11-17
Plotting

atedstone:
Add a .show() method.

### DIFF
--- a/geoutils/rastertools.py
+++ b/geoutils/rastertools.py
@@ -13,6 +13,7 @@ import rasterio.transform
 from rasterio.io import MemoryFile
 from rasterio.crs import CRS
 from rasterio.warp import Resampling
+from rasterio.plot import show as rshow
 
 from affine import Affine
 from shapely.geometry.polygon import Polygon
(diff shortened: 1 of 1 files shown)


atedstone:
n.b. there's only one commit in this PR to look at, 30484cf. I branched from #36 .

adehecq (commented):
I'm curious how rshow works underneath. Since the rasterio ds does not have any data associated with it, does it mean that rshow loads the data again in memory?

atedstone:
Interesting point. I'll look into this. Sometimes the rasterio ds does actually have data associated with it, most notably if using the classmethod `from_array`.

rhugonnet (approved):
Looks good. Seems some package names were already changed by a previous PR, will see if there is conflict when merging.

adehecq:
@rhugonnet have you looked at my earlier comment? I'm concerned that using rasterio's plotting function might load the data a second time, which is not ideal for large data set...

rhugonnet:
> @rhugonnet have you looked at my earlier comment? I'm concerned that using rasterio's plotting function might load the data a second time, which is not ideal for large data set...

So should we plot using something else instead? Cartopy?

rhugonnet:
I think the idea for now is to have some plotting capability in order to rapidly check the function outputs and move forward. I can  open an issue on this and works towards a better solution?

atedstone:
I'm inclined to agree with @rhugonnet , we can change the code underneath the plot method later if/when we find there to be a problem.

rhugonnet:
Thanks Amaury, I have linked an issue from your comment to check the data loading of rio plotting.

adehecq:
Sounds good!

PR #41, opened 2020-11-18
reproject(): Handle dst_bounds correctly when dst_res provided.

### LINKED ISSUE #40: Raster.reproject returns None if dst_bounds is set
This is because of an improper use of the Raster.crop method, which does not return a data set but update in place.
https://github.com/GlacioHack/GeoUtils/blob/68542e008d685fb9a01ae57d4e4f2befdf2120ae/geoutils/georaster.py#L425

atedstone:
Fixes #40 .

Previously, this was erroneously being handed off to crop().

Now, if dst_bounds and dst_res provided, then the new raster size is calculated in-method, ensuring that pixel size remains fixed to dst_res. This means that dst_bounds will be adjusted (increased slightly) if it was not already divisible into the requested pixel size.

### DIFF
--- a/geoutils/georaster.py
+++ b/geoutils/georaster.py
@@ -393,14 +393,38 @@ class Raster(object):
         # Determine target raster size/resolution
         dst_transform = None
         if dst_res is not None:
-            # Let rasterio determine the maximum bounds of the new raster.
-            reproj_kwargs.update({'dst_resolution': dst_res})
+            if dst_bounds is None:
+                # Let rasterio determine the maximum bounds of the new raster.
+                reproj_kwargs.update({'dst_resolution': dst_res})
+            else:
+                
+                # Bounds specified. First check if xres and yres are different.
+                if isinstance(dst_res, tuple):
+                    xres = dst_res[0]
+                    yres = dst_res[1]
+                else:
+                    xres = dst_res
+                    yres = dst_res
 
-        elif dst_size is not None:
-            # Fix raster size at nx, ny; don't change extent.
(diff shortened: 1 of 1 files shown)


atedstone:
Simple test case:
```

from geoutils import georaster
from importlib import reload
reload(georaster)
import rasterio as rio
im = georaster.Raster('LE71400412000304SGS00_B4_crop.TIF')
bounds = im.get_bounds_projected(rio.crs.CRS.from_epsg(3413))
bounds = rio.coords.BoundingBox(*bounds)
print(bounds)
print(im.reproject(3413, dst_res=100, dst_bounds=bounds))

## -- End pasted text --
BoundingBox(left=5510702.761523411, bottom=4941880.140352617, right=5551144.631103698, top=4982754.361011264)

Driver:               GTiff
File on disk:         None
RIO MemoryFile:       /vsimem/9004b163-4c89-4967-abee-644f5e5ae87e/9004b163-4c89-4967-abee-644f5e5ae87e.tif
Matches file on disk? True
Size:                 405, 409
Coordinate System:    EPSG:3413
NoData Value:         None
Pixel Size:           100.0, 100.0
Upper Left Corner:    5510702.761523411, 4941854.361011264
Lower Right Corner:   5551202.761523411, 4982754.361011264
```

adehecq (approved):
I think we could use rasterio's functionnalities to calculate the width/height. See example below
`window = rio.windows.from_bounds(*img.bounds, img.transform)`
`window.width`
`window.height`

PR #42, opened 2020-11-18
Fix filematch

atedstone:
Fixes a small bug with the Raster.matches_disk flag.


atedstone:
n.b. this was accidentally branched off the fix_reproject branch, so the PR related to that branch should be resolved first.

adehecq (approved):

Issue #31, opened 2020-11-13
Return value at given coordinates


atedstone:
Convenience wrapper, any coordinates, with optional associated crs if coordinates not in Raster crs, to look up value using rasterio functionality for this. like georaster.value_at_coords().

PR #44, opened 2020-11-18
Add method value_at_coords

adehecq:
Almost identical to georaster's value_at_coords, but using rasterio instead of GDAL.
Also added two options:
- masked: set to True to return a masked array with no data masked
- boundless: set to True to enable windows larger than raster extent

Left to do:
- implement the latlon argument
- possibly enable any input CRS

### DIFF
--- a/geoutils/georaster.py
+++ b/geoutils/georaster.py
@@ -586,6 +586,143 @@ class Raster(object):
         else:
             return extent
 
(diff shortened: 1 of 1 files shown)


adehecq:
Solve issue #31.

atedstone (approved):
This looks fine. I haven't tested - need to set up a unit test before we can rely on this.

Issue #28, opened 2020-11-13
Add option to reproject() to get destination info from another Raster


atedstone:
Add a kwarg to reproject method which allows another Raster object to be provided from which the destination CRS,transform,bounds are taken.

adehecq:
Seems to be solved, so closing.

PR #45, opened 2020-11-18
Implement feature #28 - get reprojection info for ref raster

adehecq:


adehecq:
Test case:
`import geoutils.georaster as geor`
`img = geor.Raster('test_data/LE71400412000304SGS00_B4_crop.TIF')`
`dem = geor.Raster('test_data/TDX_90m_everest.tif')`
`img_proj = img.reproject(dst_ref=dem)`
`print(dem)`
`print(img_proj)`

Output:
Driver:               GTiff 
File on disk:         /home/adehecq/development/GlacioHack/test_data/TDX_90m_everest.tif 
RIO MemoryFile:       /vsimem/1ab3b8ed-be9c-4636-9671-794f7104c5a9/1ab3b8ed-be9c-4636-9671-794f7104c5a9.tif 
Matches file on disk? True 
Size:                 6616, 7401
Coordinate System:    EPSG:32645
NoData Value:         None
Pixel Size:           30.0, 30.0
Upper Left Corner:    400740.1202974954, 2986414.2404038543
Lower Right Corner:   599220.1202974954, 3208444.2404038543

Driver:               GTiff 
File on disk:         None 
RIO MemoryFile:       /vsimem/fa990dc4-3602-48bb-83f8-66147f9efafe/fa990dc4-3602-48bb-83f8-66147f9efafe.tif 
Matches file on disk? True 
Size:                 6616, 7401
Coordinate System:    EPSG:32645
NoData Value:         None
Pixel Size:           30.0, 30.0
Upper Left Corner:    400740.1202974954, 2986414.2404038543
Lower Right Corner:   599220.1202974954, 3208444.2404038543

geoutils/georaster.py
```diff
@@ -357,23 +366,54 @@ def reproject(self, dst_crs, dst_size=None, dst_bounds=None, dst_res=None,
...
+        else:
+            if dst_crs is None:
+                raise ValueError("One of `dst_ref` or `dst_crs` must be set.")
+            
+        # Case a raster is provided as reference
+        if dst_ref is not None:
+
+            # Check that dst_ref type is either str, Raster or rasterio data set
+            if isinstance(dst_ref, Raster):
+                ds_ref = dst_ref.ds
```
atedstone, line 388: What is the rationale behind getting the crs,width, height, bounds from the underlying rasterio dataset? All these attributes are copied into the Raster object.

I've found an edge case where this is "important". When doing multiprocessing such as with dask, Rasters with opened rasterio.DatasetReaders cannot be pickled for transfer to a compute node. I found this when trying to pass a Raster containing georeferencing information as an argument for clipping/reprojecting several years of data, where each year is a multiprocessing task. 

If this line were changed to `ds_ref = dst_ref`, then all extra that would be needed would be for the user to run `myim.ds.close()` before dispatching a Raster to a compute node.

adehecq: True, I didn't realize that ds_ref = dst_ref would work too with the correct attributes transferred to the Raster object (although the user can choose not to load them with the attr argument in __init__...
I'm not sure to understand the end message in the end. Should we change this line or it's better not to? If not changing, should I add a comment to the code to explain it?

atedstone: Sorry, I should have been clearer! :)
1) Actually, all the attrs listed in `default_attrs` always get loaded. See the `_read_attrs` method: if the `attrs` keyword is provided by the user, any items in `default_attrs`which are missing from the user-provided list get added automatically.
2) Because of (1), I think that it's safe and useful to change this line to `ds_ref = dst_ref`.

atedstone: Regarding my comment above,

> Rasters with opened rasterio.DatasetReaders cannot be pickled for transfer to a compute node.

I've opened an associated issue regarding closing these readers (#50 ).

adehecq: I would like to add a comment in the code to remember this. Is that good so: "# Preferably use Raster instance to avoid rasterio data set to remain open. See PR #45"?

adehecq: Just pushed the changes. I also modified the case where dst_ref is str.

atedstone (changes requested):
Please git rm version.py from this PR. This file is auto-generated by setup.py.

atedstone:
version.py would have accidentally been committed due to recent renaming changes  which meant that path in .gitignore was incorrect. Fixed by #53 .

adehecq:
Done removing version.py @atedstone

rhugonnet (approved):
All good.

geoutils/georaster.py
rhugonnet: > Rasters with opened rasterio.DatasetReaders cannot be pickled for transfer to a compute node.

Yes, this is a pain indeed. I kind of manually circumvent this issue generally. We'll see how to best implement through the class...

atedstone (approved):

PR #49, opened 2020-11-19
fix import for shapely.ops.transform

### LINKED ISSUE #47: Cannot import proj_tools
from geoutils import projtools -> "ModuleNotFoundError: No module named 'shapely.ops.transform'; 'shapely.ops' is not a package"

atedstone:
Fixes #47


adehecq (approved):
I fixed it locally but realize I never pushed it

PR #56, opened 2020-12-08
Adding interp_points, xy2ij, ij2xy, xxyy, set_ndv, set_dtype, some corrections, and related test functions

### LINKED ISSUE #29: Method to return coordinates of all pixels
e.g. georaster.coordinates(), potentially geoimg.xy()

### LINKED ISSUE #31: Return value at given coordinates

### LINKED ISSUE #32: Method to change nodata value

### LINKED ISSUE #33: Method to change data type

### LINKED ISSUE #46: Add a method to interpolate raster at a list of coordinates
Similar to georaster's interp method: https://github.com/GeoUtils/georaster/blob/0df19bfd91fd10e92beb54308428e212e1d19c1a/georaster/georaster.py#L859
This is different from value_at_coords which extract only exact pixel values. This should be used to interpolate at any position.

### LINKED ISSUE #57: Add copy() method to duplicate Raster in memory
This is currently missing in some instances. For example: if I use crop(), I will lose my original Raster as it operates on the original one. Might want to copy before this operation.

rhugonnet:
interp_points is based on [pybob](https://github.com/iamdonovan/pybob/blob/master/pybob/GeoImg.py#L912)
resolves #46
and with Amaury's implementation: resolves #31 
xy2ij and ij2xy are just wrappers for rio functions with default parameters (which we need 95% of the time)
xxyy to generate either vector or full grid of coordinates for x and y extent of raster
resolves #29 
set_ndv to modify the ndv of a raster, with option of updating the array with the ndv (this is not always desired, e.g. if the ndv was not correctly referenced in the first place)
resolves #32 
set_dtype to modify the dtype of a raster: after testing, seems rio only support one data type for all bands, and requires the numpy array to be updated separately, which is why I did
resolves #33 

trying to add copy() issue based on later comment here for automatic linking:
resolves #57


rhugonnet:
The test functions are based mostly on visual inspection with independent calculations for now(print/figures), but I will try to modify those towards pytest syntax for consistency/robustness
Also need to provide a multi-band test Raster to check we handle those correctly (although honestly I don't want to spend too much time on this)

geoutils/georaster.py
```diff
@@ -822,10 +913,179 @@ def format_value(value):
...
+
+                #TODO: read only that window?
+                z = self.data[band-1, i - nsize:i + nsize + 1, j - nsize:j + nsize + 1]
+                if mode in ['linear', 'cubic', 'quintic', 'nearest']:
+                    X, Y = np.meshgrid(x, y)
+                    try:
+                        zint = griddata((X.flatten(), Y.flatten()), z.flatten(), list(pt), method=mode)[0]
+                    except:
+                        #TODO: currently fails when dealing with the edges
+                        print('Interpolation failed for:')
```
rhugonnet, line 1072: Right now the method fails when dealing with edges, returns NaN. An idea for these exception cases?

geoutils/georaster.py
```diff
@@ -822,10 +913,179 @@ def format_value(value):
...
+        for pt in pts:
+            i,j = self.xy2ij(pt[0],pt[1])
+            if self.outside_image(i,j, index=True):
+                rpts.append(np.nan)
+                continue
+            else:
+                x = xx[j - nsize:j + nsize + 1]
+                y = yy[i - nsize:i + nsize + 1]
+
+                #TODO: read only that window?
```
rhugonnet, line 1064: Here, we could provide the entire list to griddata to get all points derived simultaneously. Or we could write the for loop the same way Amaury did and interpolate only in a window read by rio to optimize RAM (and add a multiprocessing option to the for loop?)

rhugonnet: Maybe we provide both options? And should we merge value_at_coords and interp_points then?

tests/test_geoutils.py
```diff
@@ -0,0 +1,175 @@
...
+
+    arr_2 = np.copy(r.data)
+
+    r.set_dtypes([np.int8],update_array=True)
+
+    print(r.dtypes)
+
+    arr_3 = r.data
+
+    print(np.count_nonzero(~arr_1 == arr_2))
```
rhugonnet, line 164: Can change those and many other ones to "assert" for pytest

geoutils/georaster.py
```diff
@@ -526,6 +527,96 @@ def shift(self, xoff, yoff):
...
+        """
+        Set new nodata values for bands (and possibly update arrays)
+
+        :param ndv: nodata values
+        :type ndv: list or int or float
+        :param update_array: change the existing nodata in array
+        :type update_array: bool
+        """
+
+        if not (isinstance(ndv, list) or isinstance(ndv, int) or isinstance(ndv, float)):
```
adehecq, line 540: A tuple would be fine too. I think now most people test this by comparing if the instance is a collections.Iterable: https://docs.python.org/3/library/collections.abc.html#collections.abc.Iterable

geoutils/georaster.py
```diff
@@ -526,6 +527,96 @@ def shift(self, xoff, yoff):
...
+        if update_array:
+            if self.count == 1:
+                imgdata = imgdata.astype(dtypes[0])
+            else:
+                #TODO: double-check, but I don't think we can have different dtypes for bands with rio (1 dtype in meta)
+                imgdata = imgdata.astype(dtypes[0])
+                for i in imgdata.shape[0]:
+                    imgdata[i,:] = imgdata[i,:].astype(dtypes[0])
+        else:
+            imgdata = None
```
adehecq, line 615: If the data was loaded, your method will remove it. Should imgdata just not be changed here?

geoutils/georaster.py
```diff
@@ -822,10 +913,179 @@ def format_value(value):
...
-    
+
+    def xxyy(self, offset='corner', grid=True):
+        """
+        Get x,y coordinates of all pixels in the raster.
+
+        :param offset: coordinate type. If 'corner', returns corner coordinates of pixels.
+            If 'center', returns center coordinates. Default is corner.
+        :param grid: Return gridded coordinates. Default is True.
+        :type offset: str
```
adehecq, line 928: l 927 & 928 are switched...

geoutils/georaster.py
```diff
@@ -822,10 +913,179 @@ def format_value(value):
...
+        :type grid: bool
+        :returns x,y: numpy arrays corresponding to the x,y coordinates of each pixel.
+        """
+        assert offset in ['corner', 'center'], "ctype is not one of 'corner', 'center': {}".format(offset)
+
+        xmin, ymin, xmax, ymax = self.ds.bounds
```
adehecq, line 934: Here and everywhere, you should use the attributes of self, rather than self.ds, for example: self.bounds, self.width, self.height etc in case self.ds is closed.
Also dx and dy are already available through self.res

adehecq: Also better to stay consistent with names, use height and width instead of 'npix_x' and 'npix_y'.

adehecq: There's also an error, should dx not be list(self.ds.transform)[1]?

geoutils/georaster.py
```diff
@@ -822,10 +913,179 @@ def format_value(value):
...
+        npix_x = self.ds.width
+        npix_y = self.ds.height
+
+        # transform = Affine.from_gdal(self.ds.transform)
+        # nx, ny = self.ds.width, self.ds.height
+        # xx, yy = np.meshgrid(np.arange(nx) + 0.5, np.arange(ny) + 0.5) * transform
+
+        # return xx, yy
+
+        if dx < 0:
```
adehecq, line 946: You could do this in one line
`xx = np.linspace(xmin, xmax, npix_x + 1)[::np.sign(dx)]`
and same for yy.

geoutils/georaster.py
```diff
@@ -822,10 +913,179 @@ def format_value(value):
...
+
+        return x, y
+
+    def outside_image(self, xi,yj, index=True):
+        """
+        #TODO: calculate matricially for all points instead of doing for only one?
+        Check whether a given point falls outside of the raster.
+
+        :param xi: Indices (or coordinates) of x direction to check.
+        :param yj: Indices (or coordinates) of y direction to check.
```
adehecq, line 1006: Ok, I see you do it again. Is it not more standard to have the type just after the param, rather than all the types at the end?

geoutils/georaster.py
```diff
@@ -822,10 +913,179 @@ def format_value(value):
...
+        rpts = []
+
+        #TODO: might need to check if coordinates are center or point in the metadata here...
+
+
+        xx, yy = self.xxyy(offset='center', grid=False)
+        #TODO: right now it's a loop... could add multiprocessing parallel loop outside,
+        # but such a method probably exists already within scipy/other interpolation packages?
+        for pt in pts:
+            i,j = self.xy2ij(pt[0],pt[1])
```
adehecq, line 1056: Here your i,j values are integers (by default with rio's ds.index), so I don't see the point of interpolating...
You should modify xy2ij to have an option to return the non-rounded indexes.

geoutils/georaster.py
adehecq: In georaster, value_at_coords and interp have different use. value_at_coords will only extract the raster at certain pixel locations, for example to analyze the value at a pixel or a window around that pixel (I don't think I've ever used, that was more for Andrew I believe). On the other hand, interp() is meant to interpolate the value at any (including non integer) pixel locations. 
We could potentially merge both, but I see some differences that also support keeping them separate:
- value_at_coords read the data directly at requested location. For large arrays that's a useful feature
- on the other hand, interp loads the entire data set for the interpolation (that answers your first question).
So for now, I would keep them separate as it is, but you need to fix the issue that for now only integer pixels are read. For the interpolation, you could use ndimage.map_coordinates like we did in georaster's interp: https://github.com/GeoUtils/georaster/blob/0df19bfd91fd10e92beb54308428e212e1d19c1a/georaster/georaster.py#L936

geoutils/georaster.py
adehecq: If you were to load the entire data set and use ndimage.map_coordiates, you wouldn't have to worry about your nsize argument and edges.
A better option than loading the entire data set if not desired would be to extract the bounds of the requested pixels, read the array within these bounds and use map_coordinates.

tests/test_geoutils.py
```diff
@@ -0,0 +1,175 @@
...
+Test functions for geoutils (not using unittest)
+"""
+import os
+import geoutils.georaster as gr
+import geoutils.geovector as gv
+import matplotlib.pyplot as plt
+import numpy as np
+import pytest
+
+fn_img = os.path.join('/home/atom/code/devel/libs/GeoUtils/tests','data','LE71400412000304SGS00_B4_crop.TIF')
```
adehecq, line 11: You should define the path relative to the source code!

adehecq: The best thing would be to define a path to the test data folder somewhere, like done in georaster: https://github.com/GeoUtils/georaster/blob/0df19bfd91fd10e92beb54308428e212e1d19c1a/georaster/georaster.py#L147
You could even define a dictionary with paths to several typical data set (the Landsat crops, the RGI outlines etc).
Then you can use this to run your tests, see https://github.com/GeoUtils/georaster/blob/master/test/test_georaster.py
You may also want to copy some of those tests already.

tests/test_geoutils.py
adehecq: Yes, the best would be to copy/paste the output of some of these commands and test if it's the same.

adehecq:
Also, it would have been better to have all these commits into different PRs.

geoutils/georaster.py
```diff
@@ -822,10 +913,179 @@ def format_value(value):
...
                 'Value provided for band was not int or None.')
 
-        if return_window == True:
+        if return_window:
             return (value, win)
         else:
             return value
-    
+
+    def xxyy(self, offset='corner', grid=True):
```
adehecq, line 921: I find the name not very explicit... maybe use self.coordinates()?

geoutils/georaster.py
```diff
@@ -526,6 +527,96 @@ def shift(self, xoff, yoff):
...
+            #nodata values are specific to each band
+
+            #tried to do data cube at once, doesn't work :(
+            # ind = (imgdata == np.array([pre_ndv])[:, None, None] * np.ones((self.ds.height, self.ds.width))[None, :, :])
+            # output = np.array([ndv])[:, None, None] * np.ones((self.ds.height, self.ds.width))[None, :, :]
+            # imgdata[ind] = output
+
+            #let's do a loop then
+            if self.count == 1:
+                ind = imgdata[:] == pre_ndv
```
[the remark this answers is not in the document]
rhugonnet: Bad habit maybe :/ I've always done this. Is there a benefit at using np.where?

geoutils/georaster.py
rhugonnet: I'm not sure it will, as I just pass a None argument to the __update() method which relinks to the data loaded:
>         if imgdata is None:
>             imgdata = self.data
And updating imgdata doesn't update self.data?

geoutils/georaster.py
rhugonnet: Agreed, this can be shortened.
No error I think, the transform provided through rio looks like this: [30.0, 0.0, 478000.0, 0.0, -30.0, 3108140.0, 0.0, 0.0, 1.0]

geoutils/georaster.py
rhugonnet: It is indeed more standard to do: param type param type ([cf doc](https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html))
This was copied from other code and I simply forgot to change (so it's basically Bob's fault @iamdonovan )

geoutils/georaster.py
rhugonnet: Interpolation at the exact coordinates is done further down:
`                        zint = griddata((X.flatten(), Y.flatten()), z.flatten(), list(pt), method=mode)[0]
`
Here we just find the center of the window centered on the coordinates, to create smaller arrays passed to griddata (which is why  the process resembles quite a bit the "value_at_coords" method).

geoutils/georaster.py
rhugonnet: sure

tests/test_geoutils.py
rhugonnet: The path relative to source code is defined just below, it was commented as it works only on command line, and I prefer executing from the console when I test :/. Maybe you know a solution that works both cases? (it's a pain to switch from live console in editor to command line just for testing...)
I think it's better to have the path directly defined in test_geoutils rather than "hidden" all the way to georaster.py (as in https://github.com/GeoUtils/georaster/blob/0df19bfd91fd10e92beb54308428e212e1d19c1a/georaster/georaster.py#L147)

From talking with Fabien (raster_tools slack), we concluded it was probably better to use pytest than unittest. So we'd have to rework the tests a bit before copying them (which is why I did not do it).

geoutils/georaster.py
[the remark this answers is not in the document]
adehecq: Not sure... 
At least, just for clarity, I would write `ind = (imgdata[:] == pre_ndv)`. I read it quickly and thought it was `ind = imgdata[:] = pre_ndv` which would be very different.

geoutils/georaster.py
rhugonnet: Great tip. So basically collections.Iterable includes list, tuples, and arrays? (learning how to do things properly sure takes time :D)

geoutils/georaster.py
adehecq: Alright, I didn't realize that _update's default was None and that by default it would set it to self.data. In that case, it's all good!

geoutils/georaster.py
adehecq: Ok, so rio's transform is different from GDAL's, good to know!

geoutils/georaster.py
adehecq: Of course, it's Bob's fault!

geoutils/georaster.py
rhugonnet: Yeah I fell into the same trap and wrote [1] and [5] at first. I'm so used to GDAL, switching is a challenge... é_è

geoutils/georaster.py
adehecq: Ok I understand. In this case, why not use value_at_coords to extract the pixels needed for interpolation? Did you notice the window option?

tests/test_geoutils.py
adehecq: Regarding finding the path to a code, I think this solves your issue: https://stackoverflow.com/questions/2632199/how-do-i-get-the-path-of-the-current-executed-file-in-python/5748583#5748583
This way, you can access the path to a local function.

geoutils/georaster.py
adehecq: Yes, I just found about it recently. Not sure what else is included in these Iterable, but worth testing with tuple, list and arrays.

tests/test_geoutils.py
rhugonnet: Implemented, works great.

geoutils/georaster.py
rhugonnet: used "coords" to keep things short

tests/test_geoutils.py
rhugonnet: I'll keep on working to improve those tests

rhugonnet:
I have now added a copy() function to in-memory copy a raster (based on the class function from_array)
resolves #57

rhugonnet:
I will work on merging find_values/interp consistently tomorrow so that we can merge this PR

atedstone (approved):
I propose that we accept and merge this monster PR. The 2-3 remaining points of conversation which need to be resolved could be opened as issues to be dealt with as separate smaller fixes, as it looks to me that they are refinements/nuances rather than anything too critical?

atedstone:
@adehecq if you agree then go ahead and merge the PR

adehecq (approved):

Issue #55, opened 2020-12-03
Check if rio plotting duplicates data loading


rhugonnet:
@rhugonnet have you looked at my earlier comment? I'm concerned that using rasterio's plotting function might load the data a second time, which is not ideal for large data set...

_Originally posted by @adehecq in https://github.com/GlacioHack/GeoUtils/issues/37#issuecomment-738100402_

adehecq:
According to rasterio's code here https://github.com/mapbox/rasterio/blob/a1cb28a77ebca80c8a804f24d09350e02ed9bf63/rasterio/plot.py#L91
The dataset is loaded into memory, therefore possibly duplicating the data. Better to provide the loaded data (np array) directly to show, rather than dataset.

PR #59, opened 2021-02-18
Make sure that plotting doesn't reload data. Solves #55

adehecq:


geoutils/georaster.py
```diff
@@ -775,31 +775,46 @@ def intersection(self, rst):
...
-        :returns: None
-        :rtype: None
-
         """
         # If data is not loaded, need to load it
         if not self.isLoaded:
             self.load()
 
+        # Check if specific band selected, or take all
+        # rshow takes care of image dimensions
```
rhugonnet, line 804: that's pretty neat

rhugonnet (approved):

geoutils/georaster.py
rhugonnet: did you test it?! it would be good to put an example in pytest to ensure it does not fail; there's no need for an assertion statement, the fact that it would run without failing is enough information for this, I think

adehecq: I'll look into it

adehecq: I just did! You can try it in the new tests (better if you set DO_PLOT=True).
I saw people adding "assert True" at teh end of their test, but not sure if that's really useful...

rhugonnet:
looks good! (and given this can be directly checked in pytest, I guess runs good too)

rhugonnet (approved):

PR #62, opened 2021-02-18
Set up ReadTheDocs

atedstone:
ReadTheDocs is now working for this branch (https://geoutils.readthedocs.io/en/docs/) - these commits contain the eventual working configuration.

Also note that requirements.txt now added


atedstone:
Fixes #62 .

adehecq:
Very basic, but looks like a good start! Will you be able to add the API?

atedstone:
Pfft, you have no idea the amount of grief it's taken just to get to this point....thanks @jlandmann for help too!

But yes I will try to get an API page working. 

We should also start user guides/tutorials for georaster and geovector.

setup.py
```diff
@@ -37,3 +25,18 @@ def write_version_py(filename=None):
...
+
+
+setup(name='geoutils',
+      version=FULLVERSION,
+      description='',
+      url='',
+      author='The GlacioHack Team',
+      license='BSD-3',
+      packages=['geoutils'],
+      install_requires=['rasterio', 'geopandas', 'pyproj'],
```
adehecq, line 37: This should be consistent with the requirement.txt file, no?

atedstone: Not necessarily: https://packaging.python.org/discussions/install-requires-vs-requirements/

.readthedocs.yml
```diff
@@ -0,0 +1,22 @@
...
+# Optionally build your docs in additional formats such as PDF and ePub
+formats: []
+
+python:
+  version: 3.7
+  install:
+    - requirements: requirements.txt
+    - method: pip
+      path: .
+  system_packages: false
```
jlandmann, line 22: Not even sure if that's "allowed" here, but if it wants a new line, we can give it a try ;-)

doc/conf.py
```diff
@@ -93,160 +58,8 @@
...
-        pass
-
-    text = text_version if '+' not in GeoUtils.__version__ else text_dev
-
-    with open(filename, 'w') as f:
-        f.write(text)
-
-
-write_index()
+html_static_path = ['_static']
```
jlandmann, line 65: same here ;-)

geoutils/__init__.py
```diff
@@ -4,6 +4,10 @@
 
 from . import georaster
 from . import geovector
+from os import path
+fn = path.join(path.dirname(__file__), 'version.py')
+print('Checking', fn)
+print('Does version.py exist?', path.exists(fn))
```
jlandmann, line 10: Maybe this is not necessary anymore, but we can still keep it.

setup.py
```diff
@@ -37,3 +25,18 @@ def write_version_py(filename=None):
 
 if write_version:
     write_version_py()
+
+
+setup(name='geoutils',
+      version=FULLVERSION,
+      description='',
+      url='',
```
jlandmann, line 33: Add URL?

setup.py
```diff
@@ -37,3 +25,18 @@ def write_version_py(filename=None):
...
+      author='The GlacioHack Team',
+      license='BSD-3',
+      packages=['geoutils'],
+      install_requires=['rasterio', 'geopandas', 'pyproj'],
+      extras_require={'rioxarray': ['rioxarray']},
+      scripts=[],
+      zip_safe=False)
+
+print('Does version.py exist?', path.exists(path.join(path.dirname(__file__), 'geoutils','version.py')))
```
jlandmann, line 42: \n

jlandmann (approved):
Thanks for the great work! Looks goodto me! I just some really mini things...we should give it a try!

Issue #65, opened 2021-02-18
Set up testing integration


atedstone:

atedstone:
Solved by #66

PR #66, opened 2021-02-18
Set up CI/testing

atedstone:
For issue #65

### DIFF
--- a/tests/test_geoutils.py
+++ b/tests/test_geoutils.py
@@ -13,9 +13,7 @@ DO_PLOT = False
 
 @pytest.fixture()
 def path_data():
-    path_module = os.path.dirname(os.path.dirname(os.path.abspath(inspect.getsourcefile(gr))))
-    print(path_module)
-    data_folder = os.path.join(path_module, 'tests', 'data')
+    data_folder = os.path.join('tests', 'data')
 
     path2data = {}
     path2data['fn_img'] = os.path.join(data_folder, 'LE71400412000304SGS00_B4_crop.TIF')
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -25,6 +25,7 @@ jobs:
         python -m pip install --upgrade pip
         pip install flake8 pytest
         if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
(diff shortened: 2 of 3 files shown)


atedstone:
The new yml file runs the full test suite on all PRs to the `dev` branch.

tests/test_geoutils.py
```diff
@@ -13,9 +13,7 @@
 
 @pytest.fixture()
 def path_data():
-    path_module = os.path.dirname(os.path.dirname(os.path.abspath(inspect.getsourcefile(gr))))
-    print(path_module)
-    data_folder = os.path.join(path_module, 'tests', 'data')
+    data_folder = os.path.join('tests', 'data')
```
adehecq, line 16: Are you sure this will work from anywhere? Don't you need to specify the module path?

adehecq: Ok, I'll accept the PR and CI should tell us if that works anyway ;-)

adehecq (approved):

tests/test_geoutils.py
atedstone: The previous approach, using abspath, failed on CI - 'file not found' errors. This approach works - see the tick next to #69cd427.

atedstone: (and locally, simply run `pytest` from the dev folder and it will automatically sniff out the tests

adehecq: But what if we want to use this data from somewhere else, let's say a notebook folder with some examples. Will that still work?

atedstone: I haven't thought about that side of it - to do so, one would need to import test_geoutils then run the function, which seems a bit odd. perhaps we need to look at how other big packages manage their demo data instead, so that we are not relying on the test suite to provide this.

atedstone: opened #70

rhugonnet: yes, this is why I wrote it this way in the first place! it was a pain to run part of tests from a console, like through PyCharm, with a relative path while the previous approach worked from both command line and Python project console (based on a previous comment from Amaury, I think); shame it didn't work for CI

PR #71, opened 2021-02-19
Raster data getter/setter methods

### LINKED ISSUE #24: set and get methods for raster data array
At the moment, the raster data (if loaded) are stored in Raster.data. This is a publicly available np.array.

The problem with this is that data can be arbitrarily overwritten without any sanity check of whether the dimensions match the Raster. It could also cause problems with nodata values.

I propose that the raw np.array is moved to Raster._data. We keep the Raster.data accessor, but overload it with set/get methods.

elischwat:
Closes #24 

`get_data()` and `set_data()` methods added to the Raster class.
I put a low level check in `set_data` that makes sure new data array is the same shape as the existing one. I'm not sure that addresses the check that @atedstone discussed in his issue - Ted will you review and let me know?

I'm also a bit confused as to how this `set_data()` method conflicts with the purpose of _update(), which seems to update the original file with new data. I'm pretty sure that is distinct from the goal of `set_data()`, but again, @atedstone, let me know.

I changed the tests accordingly and those are passing on my machine.

Note that this will probably break people's existing code that lives outside of GeoUtils.

### DIFF
--- a/geoutils/georaster.py
+++ b/geoutils/georaster.py
@@ -93,13 +93,13 @@ class Raster(object):
         self._read_attrs(attrs)
 
         if load_data:
-            self.data = self.ds.read(bands)
-            self.nbands = self.data.shape[0]
+            self._data = self.ds.read(bands)
+            self.nbands = self._data.shape[0]
             self.isLoaded = True
             if isinstance(filename, str):
                 self.matches_disk = True
         else:
-            self.data = None
+            self._data = None
             self.nbands = None
             self.isLoaded = False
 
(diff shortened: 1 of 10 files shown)


geoutils/georaster.py
```diff
@@ -194,6 +194,27 @@ def _read_attrs(self, attrs=None):
         for attr in attrs:
             setattr(self, attr, getattr(self.ds, attr))
 
+    def set_data(self, new_data):
```
adehecq, line 197: Something to be tested (I just found about these functions), but I think we should use setter and getter decorators like in the last example here: https://www.geeksforgeeks.org/getter-and-setter-in-python/

atedstone: yes, exactly my thinking too.

adehecq: Some quick example I've tested:
```
class Raster:
    def __init__(self, data=None):
        self.shape = (3,3)
        if data is None:
            self._data = np.arange(9).reshape((3,3))
        else:
            self.data = data

    @property
    def data(self):
        return self._data
        
    @data.setter
    def data(self, d):
        if d.shape != self.shape:
            raise ValueError("Data has wrong shape")
        else:
            self._data = d
```
Then
```
img = Raster()
img.data
```
array([[0, 1, 2],
       [3, 4, 5],
       [6, 7, 8]])
```
img.data = np.arange(9)
```
ValueError                                Traceback (most recent call last)
<ipython-input-27-e54a5000bc4c> in <module>
----> 1 img.data = np.arange(9)

<ipython-input-19-cb9ce1fef3a3> in data(self, d)
     14     def data(self, d):
     15         if d.shape != self.shape:
---> 16             raise ValueError("Data has wrong shape")
     17         else:
     18             self._data = d

ValueError: Data has wrong shape

adehecq:
Note that it is preferable to use _data rather than __data. In the latter case, the Python interpreter would rename it _Raster__data, making it unavailable for child classes. See example below.
image

adehecq (approved):

elischwat:
Thanks for bringing this across the finish line!

PR #77, opened 2021-02-19
Replace _data by data where possible

adehecq:


atedstone (approved):
This looks fine to me. I think that the remaining cases where _data is set directly are justified.

PR #78, opened 2021-02-19
Add number of bands and datatypes to info(). Fix #63

### LINKED ISSUE #63: Add number of bands to info()

adehecq:


rhugonnet:
rio doesn't have those lines in it's "info", so I'm surprised that the test_info() check is still passing:
```
#check summary matches that of RIO
        assert print(r) == print(r.info())
```
do you know why?

adehecq:
in your example, r.info() is not rasterio's info, it's our info!
I don't think rasterio has something similar, does it?

rhugonnet:
Oh, so the "print" function on an object automatically calls the .info() function within its class?! that's pretty neat, I didn't know

adehecq:
It's because of this line: https://github.com/GlacioHack/GeoUtils/blob/951f78cb8e3f7a6be284bdbe0415c139688b3a81/geoutils/georaster.py#L177

adehecq:
Can you accept the PR?

rhugonnet (approved):

PR #79, opened 2021-02-19
Create a separate dataset folder and function to access it

adehecq:
@rhugonnet @atedstone 
What do you think of this to access the demo data? It's similar to geopandas. The data path can be found with:
```
from geoutils import datasets
datasets.get_path("landsat_B4")
```
and the list of available datasets
`datasets.datasets`


adehecq:
If that works, I'll update the tests accordingly.

rhugonnet:
looks good indeed! hope it works without the circular import issue

atedstone (approved):
this looks good to me

adehecq:
Try using relative paths in the tests so that it doesn't fail in CI tests.

adehecq:
For the record: the issue with failed tests on the datasets is that datasets is a package (folder with init file) rather than module. Packages must be defined in the seup.py file.
Additionally, the data files must be provided in the setup.py file to be copied when using `pip install .`.

PR #80, opened 2021-02-19
Remove redundant memfile

atedstone:
This commit removes the attribute self.memfile, which is redundant as far as I can tell.


adehecq (approved):
Looks fine to me

PR #85, opened 2021-02-25
Minor edits when merging with Amaury's fork

adehecq:


rhugonnet (approved):

rhugonnet:
out of curiosity, why is it better to write it "from geoutils import datasets"?

adehecq:
There's no better one, just the new one is more concise. It only showed up because Andrew made an attempt to solve the test issue with a different import statement, but it didn't work. On my fork I was still use the older import.

PR #86, opened 2021-03-01
Fix coords() function

atedstone:
`Raster.coords()` was returning incorrect y-coordinate grids. This was probably caused by an inconsistency in syntax between the old Georaster class from which this function was taken, vs new rasterio functionality.

* Method updated
* Test case added
* No test case for images with positive y axis resolution increment - I don't have any test images like this to hand, ideas welcome!


geoutils/georaster.py
```diff
@@ -1020,18 +1020,17 @@ def coords(self, offset='corner', grid=True):
...
         assert offset in ['corner', 'center'], "ctype is not one of 'corner', 'center': {}".format(offset)
 
-        xmin, ymin, xmax, ymax = self.bounds
-        dx = list(self.transform)[0]
-        dy = list(self.transform)[4]
+        dx = self.res[0]
+        dy = self.res[1]
 
-        xx = np.linspace(xmin, xmax, self.width + 1)[::int(np.sign(dx))]
-        yy = np.linspace(xmin, xmax, self.height + 1)[::int(np.sign(dy))]
```
rhugonnet, line 1034: looks like it was xmin/xmax instead of ymin/ymax here

rhugonnet: my mistake, probably... weird because I tested it at some point

rhugonnet (approved):

rhugonnet:
looks great!

rhugonnet:
I'm thinking: this modification will affect the current "interp_points" function where coords is used. There is no assertion yet in the test_interp() so the tests passed, but I'll have a look at the printed results of the test to see if it's still consistent or if we need to fix interp_points as well

atedstone:
thanks @rhugonnet , i'll wait to hear back before we merge in this PR

adehecq (approved):
Sounds all good to me!
For the negative y resolution, we could create a synthetic test case.

adehecq:
Want to merge this @atedstone ?

atedstone:
@rhugonnet , do you want to look at `interp_points` before we close this PR?

rhugonnet:
> @rhugonnet , do you want to look at `interp_points` before we close this PR?

Just opened #100 to check this in details, we can merge this PR in the meantime :+1:

Issue #90, opened 2021-03-03
Raster.show() does not take nodata into account


adehecq:
A raster with no data is plotted with no data values masked.

adehecq:
Fixed by #91

Issue #89, opened 2021-03-03
set_ndv does not accept np.float32


adehecq:
This line should be modified to also accept numpy numbers/
https://github.com/GlacioHack/GeoUtils/blob/19e1fd521a0899529a8eef23b5ea0ae5258bc1c7/geoutils/georaster.py#L588

adehecq:
Solved by #91

PR #91, opened 2021-03-03
Use masked_array to handle no data values

adehecq:
- now the default is to load the data as a np.masked_array using rasterio's option
- update the set_ndv and test_set_ndv accordingly

For now, both normal numpy arrays and masked arrays are allowed, but the default should probably be only masked_arrays which means:
- updating data.setter to check type
- check that all comparison to no data value are moved to using self.data.mask


adehecq:
Solves #90

adehecq:
Solves #89

geoutils/georaster.py
```diff
@@ -585,13 +590,18 @@ def set_ndv(self, ndv, update_array=False):
...
-        if not (isinstance(ndv, collections.abc.Iterable) or isinstance(ndv, int) or isinstance(ndv, float)):
+        if not isinstance(ndv,
+                          (collections.abc.Iterable, int, float,
+                           np.integer, np.floating)):
             raise ValueError(
                 "Type of ndv not understood, must be list or float or int")
-        elif (isinstance(ndv,int) or isinstance(ndv,float)) and self.count>1:
+
+        elif (isinstance(ndv,
+                         (int, float, np.integer, np.floating))) and self.count > 1:
```
rhugonnet, line 600: didn't know the int types didn't include np.integer (and vice-versa), this can be quite troublesome! I guess you also did something like this to double-check: 
```python
isinstance(np.int16(1),int)
Out[36]: False
isinstance(np.int16(1),np.integer)
Out[37]: True
isinstance(int(1),np.integer)
Out[38]: False
```

rhugonnet: and checked on stack overflow... no better way! quite surprised

adehecq: Yes, I didn't know either and found out while testing set_ndv. But for example if you get your value with np.min or np.max, then the output is a np integer or floating.

adehecq: At least, the fact that you can put a list in `isinstance` helps to reduce the length a bit.

tests/test_georaster.py
```diff
@@ -147,18 +147,36 @@ def test_interp(self):
         # print(rpts)
 
     def test_set_ndv(self):
-
+        """
+        Read Landsat dataset and set 255 to no data. Save mask.
+        Then, set 254 as new no data (after setting 254 to 0). Save mask.
```
rhugonnet, line 152: look at this, a well commented code *flies away*

tests/test_georaster.py
```diff
@@ -147,18 +147,36 @@ def test_interp(self):
...
+            plt.imshow(ndv_index_2[0], interpolation='nearest')
+            plt.title('Mask 2 (should be identical)')
+            plt.show()
+
+        # Check both masks are identical
+        assert np.all(ndv_index_2 == ndv_index)
 
-        assert np.count_nonzero(~ndv_index_2 == ndv_index) == 0
+        # Check that the number of no data value is correct
+        assert np.count_nonzero(ndv_index.data) == 112088
```
rhugonnet, line 179: you kept my np.count_nonzero, I'm touched

rhugonnet (approved):

rhugonnet:
looks good, and solves 2 issues! :+1:

tests/test_georaster.py
adehecq: ;-)

adehecq:
> looks good, and solves 2 issues! +1

Yes, and in the meantime I added at least 3 issues!

PR #95, opened 2021-03-04
Add more plotting functionalities

adehecq:
For now, simply adding functionalities to Raster.show. Planning to add a new geoviewer file with command line tools etc for plotting.
I think georaster should only rely on matplotlib for plotting at this stage. Anything more complex like cartopy will be in geoviewer I think.


adehecq:
Added an initial geoviewer script, mostly based on https://github.com/GeoUtils/geoutils/blob/master/geoutils/plot_geo_image.py

geoutils/georaster.py
```diff
@@ -851,17 +854,29 @@ def intersection(self, rst):
         else:
             return extent
 
-    def show(self, band=None, **kwargs):
+    def show(self, band=None, cmap=None, vmin=None, vmax=None, cb_title=None,
+             no_cb=False, ax=None, **kwargs):
```
jlandmann, line 858: Does it have a specific reason that it's ckaaed `*no*_cb`? I would call it `add_colorbar` or so...that avoids confusion in my head ;-)

geoutils/georaster.py
```diff
@@ -889,8 +904,64 @@ def show(self, band=None, **kwargs):
...
+        # Use rcParam default
+        if cmap is None:
+            cmap = plt.get_cmap(plt.rcParams['image.cmap'])
+        elif isinstance(cmap, str):
+            cmap = plt.get_cmap(cmap)
+        elif isinstance(cmap, matplotlib.colors.Colormap):
+            pass
+
+        # Set colorbar min/max values
+        if vmin is None:
```
rhugonnet, line 922: this you could probably leave out as it's just passing arguments, the exceptions should be handled by matplotlib.colors/matplotlib.pyplot

geoutils/geoviewer.py
```diff
@@ -0,0 +1,164 @@
+#!/usr/bin/env python
+#coding=utf-8
+"""
+geoutils.geoviewer provides a toolset for plotting raster and vector data
```
rhugonnet, line 4: command line executables should probably live somewhere else than packages for clarity (/bin?)

setup.py
```diff
@@ -47,5 +47,5 @@ def write_version_py(filename=None):
       package_data={"geoutils": data_files},
       install_requires=['rasterio', 'geopandas', 'pyproj'],
       extras_require={'rioxarray': ['rioxarray']},
-      scripts=[],
+      scripts=['geoutils/geoviewer.py'],
```
rhugonnet, line 50: to update in case this is moved to /bin or such

jlandmann (commented):
Great work!
I just found the `no_cbar`argument a bit complicated.

tests/test_georaster.py
```diff
@@ -215,9 +222,20 @@ def test_plot(self):
...
+        ax = plt.subplot(111)
+        img_RGB.show(band=0, cmap='gray', ax=ax, no_cb=True,
+                     title="Plotting one band B/W")
+        if DO_PLOT:
+            plt.show()
+        else:
+            plt.close()
+        assert True
+
+        # Test vmin, vmax and cb_title
```
rhugonnet, line 235: looks good! there's even practical comments describing the tests :D

rhugonnet (approved):

geoutils/georaster.py
adehecq: Mmmh, good question. I think it's mostly because I;m used to run it from command line with argparse. It's easier to make cbar the default, and add a -nocb option to remove it if not wanted.
What do you think? You can look at geoviewer.py how that would translate.

geoutils/georaster.py
adehecq: Which lines exactly do you mean are not needed? Just the `isinstance(cmap, matplotlib.colors.Colormap)` or the whole condition block?

geoutils/geoviewer.py
adehecq: Good point... It all depends what we put in this geoviewer.py. We could have a library where we put a bunch of useful functions for displaying and plots options, that could be accessed as a submodule. Then we could have a simple CLI tool with some of the options. But how to call them both in that case??

geoutils/georaster.py
rhugonnet: I'm thinking that all the arguments later passed without modification to a matplotlib function should not be checked by us. This is already done by matplotlib

geoutils/georaster.py
jlandmann: Well, if I get it correctly it's about changing the argument name and default in this line, right? (and then of course the subsequent uses) : https://github.com/adehecq/GeoUtils/blob/084b0a263283a9830917e65cd11bed4be8ab5fb6/geoutils/geoviewer.py#L34

adehecq: Just replaced it in commit 36ab162 !

geoutils/georaster.py
adehecq: The issue is that vmin/vmax are not only used in imshow, but also in ScalarMappable. There, setting them to None results in a colorbar between 0 and 1...

rhugonnet: Yes that part is perfect, it's the `isinstance(vmin, (int, float, np.integer, np.floating))` that does not seem necessary because will be raised by matplotlib.colors.Normalize, but this is really minor you can leave it

adehecq: I just improved/shortened a bit the condition in commit b45dc66.
Without the try/except, mpl would raise an error that can be difficult to trace.

geoutils/georaster.py
jlandmann: Cool thanks!!

ShashankBice (approved):
Looks good to me, some really nice and detailed feedback by Romain and Johannes. Nice use of try and except to make sure the vmin/vmax are actually numbers. I also like this review process to get feedback over changes from several places, likely improving things before final merge :)

PR #99, opened 2021-03-04
Add a downsampling functionality and improve test suite

adehecq:


tests/test_georaster.py
```diff
@@ -31,6 +32,64 @@ def test_info(self):
         # Check summary matches that of RIO
         assert print(r) == print(r.info())
 
+    def test_loading(self):
```
adehecq, line 35: What do you think about that @rhugonnet ?

geoutils/georaster.py
```diff
@@ -40,24 +40,25 @@ class Raster(object):
...
         :type filename: str
-        :param attrs: Additional attributes from rasterio's DataReader class to add to the Raster object.
-            Default list is ['bounds', 'count', 'crs', 'dataset_mask', 'driver', 'dtypes', 'height', 'indexes',
-            'name', 'nodata', 'res', 'shape', 'transform', 'width'] - if no attrs are specified, these will be added.
-        :type attrs: list of strings
-        :param load_data: Load the raster data into the object. Default is True.
-        :type load_data: bool
         :param bands: The band(s) to load into the object. Default is to load all bands.
         :type bands: int, or list of ints
+        :param load_data: Load the raster data into the object. Default is True.
```
rhugonnet, line 52: took me 20s to realize you switched this to be consistent with order of  parameters!

geoutils/georaster.py
```diff
@@ -95,8 +96,28 @@ def __init__(self, filename, attrs=None, load_data=True, bands=None,
...
+        if downsampl == 1:
+            out_shape = (nbands, self.height, self.width)
+        else:
+            down_width = int(np.ceil(self.width/downsampl))
+            down_height = int(np.ceil(self.height/downsampl))
+            out_shape = (nbands, down_height, down_width)
+
         if load_data:
-            self.load()
+            self.load(bands=bands, out_shape=out_shape)
```
rhugonnet, line 120: I see, with out_shape passed down directly to rio.ds.read(); should we leave the default nearest interpolation, or change to bilinear? in some case (e.g. for DEMs), I guess it's still better to downsample using the exact decimal coordinates. Not sure if truly necessary... and would need to add a `downsamp_method` argument to __init__()

tests/test_georaster.py
```diff
@@ -31,6 +32,78 @@ def test_info(self):
...
+        Check that self.data is correct when using downsampling
+        """
+        # Test single band
+        r = gr.Raster(datasets.get_path("landsat_B4"), downsampl=4)
+        assert r.data.shape == (1, 164, 200)
+        assert r.height == 655  # this should not have changed
+        assert r.width == 800
+
+        # Test multiple band
+        r = gr.Raster(datasets.get_path("landsat_RGB"), downsampl=2)
```
rhugonnet, line 104: Looking at the code, this should even work with a decimal downsampling factor, no?

tests/test_georaster.py
rhugonnet: I think it looks good!

rhugonnet (approved):

geoutils/georaster.py
adehecq: For me, the main use case is simply for quickly visualizing large rasters. So nearest is probably the fastest. I would not use it for running calculations on it. In that case, we have self.interp etc.
There is still the option to use a different resampling algo using self.load directly.

tests/test_georaster.py
adehecq: Yes, I guess that's true!

adehecq: But it could be a bit weird with the nearest resampling. For example with 1.5, basically every 3 column would be skipped.

adehecq: I enabled it in the last commit.

Issue #102, opened 2021-03-11
Suggestion: Raster.reproject() should only reproject if needed


erikmannerfelt:
If the user wants one DEM to fit another, they would presumably use the `.reproject()` method directly. If the DEMs already fit, however, the reprojection is still done. This adds an unnecessary processing time (or even failure in case of gigantic DEMs), and could be avoided with a sanity check within the `.reproject()` method.

To exemplify what I mean:
```python

import time

import geoutils as gu
import rasterio as rio

# Load a KH-9 DEM sent by Amaury which is quite big
dem = gu.georaster.Raster("DZB1208-500032L004006/DZB1208-500032L004006-DEM.tif")

# I reproject the DEM to its own exact bounds using cubic spline
# This is a proxy for if a DEM with the exact same bounds as a reference DEM should (not) be reprojected.
time_before = time.time()
dem2 = dem.reproject(dem, resampling=rio.warp.Resampling.cubic_spline)
time_after = time.time()
print(f"{time_after - time_before:.2f} s")

```
Which returned: 
```
4.98 s
```
I would argue that this is a shortcoming or even bug in either `georaster` or `rasterio`. Why would resampling be done if the source and destination bounds/crs/resolution is the same?

Something along this could be added:
```python

if dst_ref.bounds == self.bounds and dst_ref.res == self.res and dst_ref.crs == self.crs and dst_ref.nodata == self.nodata:
     return dst_ref
```
or the same with shorter lines:
```python
if all([
    dst_ref.bounds == self.bounds,
    dst_ref.res == self.res,
    dst_ref.crs == self.crs,
    dst_ref.nodata == self.nodata
]):
    return dst_ref
```

This is not a huge issue, but I think the check is a valuable addition!

Issue #106, opened 2021-03-11
In reproject, dtype not taken into account


adehecq:

adehecq:
Solevd by #107

PR #107, opened 2021-03-11
Ensure that reproject does not run if not needed

adehecq:
It was longer than expected to make sure that all cases where covered, but I think I got it.

Also made another simple change. Before, either dst_ref or dst_crs had to be set, but actually, if dst_size is different, both can be None. Of course, it is not really "reprojection" if the CRS is the same, but the method covers all kind of CRS/transform change.


geoutils/georaster.py
```diff
@@ -579,6 +579,20 @@ def reproject(self, dst_ref=None, dst_crs=None, dst_size=None, dst_bounds=None,
...
 
+        # Check that reprojection is actually needed
+        if all([
+                (dst_transform == self.transform) or (dst_transform is None),
+                (dst_crs == self.crs) or (dst_crs is None),
+                (dst_size == self.shape) or (dst_size is None),
+                (dst_res == self.res) or (dst_res == self.res[0] == self.res[1]) or (dst_res is None)
+        ]):
+            if (nodata == self.nodata) or (nodata is None):
+                warnings.warn("Output projection, bounds and size are identical -> return self")
```
adehecq, line 590: I put a warning for now. Is that the most sensible?

rhugonnet: Mmmh.. Don't know. We could even skip the warning?

geoutils/georaster.py
```diff
@@ -484,7 +484,7 @@ def reproject(self, dst_ref=None, dst_crs=None, dst_size=None, dst_bounds=None,
                 raise ValueError("Either of `dst_ref` or `dst_crs` must be set. Not both.")
         else:
             if dst_crs is None:
-                raise ValueError("One of `dst_ref` or `dst_crs` must be set.")
+                dst_crs = self.crs
```
rhugonnet, line 487: I would raise a warning here: "destination crs not specified, not reprojecting"

geoutils/georaster.py
```diff
@@ -579,6 +579,20 @@ def reproject(self, dst_ref=None, dst_crs=None, dst_size=None, dst_bounds=None,
...
+                (dst_crs == self.crs) or (dst_crs is None),
+                (dst_size == self.shape) or (dst_size is None),
+                (dst_res == self.res) or (dst_res == self.res[0] == self.res[1]) or (dst_res is None)
+        ]):
+            if (nodata == self.nodata) or (nodata is None):
+                warnings.warn("Output projection, bounds and size are identical -> return self")
+                return self
+
+            else:
+                raise ValueError("Only nodata is different, use self.set_ndv")
```
rhugonnet, line 594: or we could default to a warning + apply set_ndv()? not sure this is going to happen very often

rhugonnet (approved):
Looks good!

geoutils/georaster.py
adehecq: As said in the main discussion, there is still a case where dst_size (or res) can be set and it will require reshaping the array without changing the CRS. This is why I allow this.

geoutils/georaster.py
adehecq: Actually, one thing is that set_ndv modifies self in place. This needs to return a copy, so that will be the only difference.
Should there be an option in set_ndv to make a copy?? Maybe not, users can always copy then set_ndv if needed.

geoutils/georaster.py
adehecq: After thinking, I think there should be a warning. In all other cases, reproject returns a copy, here we return self, so any changes made to the output after reproject will also be applied to the original dataset...

geoutils/georaster.py
```diff
@@ -579,6 +580,23 @@ def reproject(self, dst_ref=None, dst_crs=None, dst_size=None, dst_bounds=None,
...
+                (dst_crs == self.crs) or (dst_crs is None),
+                (dst_size == self.shape) or (dst_size is None),
+                (dst_res == self.res) or (dst_res == self.res[0] == self.res[1]) or (dst_res is None)
+        ]):
+            if (nodata == self.nodata) or (nodata is None):
+                warnings.warn("Output projection, bounds and size are identical -> return self")
+                return self
+
+            else:
+                warnings.warn("Only nodata is different, running self.set_ndv instead")
```
erikmannerfelt, line 595: Wouldn't it be nice to have a keyword argument to disable this warning?

An example where I would get annoyed by the warning:
```python
# I want to make sure that the to-be-aligned DEM has the same shape and extent as the reference
# and I don't care what is done to make sure of this! Maybe they already have the same shape.
# Kind of like np.asarray() just to make sure that it's an array
ref_dem = Raster("path/to/ref")
tba_dem = Raster("path/to/tba").reproject(ref_dem)


```

adehecq: We could possibly have a silent option... Or you can disable warnings...
It's just that in your case, tba_dem and ref_dem are the same objects, so it's a bit dangerous...

geoutils/georaster.py
rhugonnet: Ah yes, fair enough. Maybe write this in the warning (not returning a copy)

geoutils/georaster.py
adehecq: Your wish was just granted @erikmannerfelt !

geoutils/georaster.py
adehecq: Done

erikmannerfelt (approved):
Wohoo!

geoutils/georaster.py
erikmannerfelt: Nice! 

"path/to/ref" and "path/to/tba" is different so they would not be the same object!

adehecq: Looks like you're right... But then this is not doing what we want, because it means it is still reproject and not returning self...

adehecq: For example:
```
img2 = gr.Raster(datasets.get_path('landsat_B4'))
img2_proj = img2.reproject(img)  # This should return img2
print(img2_proj.data[0,0,0])
img2_proj.data[0,0,0] = 0
print(img2_proj.data[0,0,0])
print(img2.data[0,0,0])
```
returns
255
0
255
so it means img2_proj and img2 are different objects...

adehecq: @erikmannerfelt, with commit 08783b1, this is different now. If your dst_ref has the same projection/dimension as your raster, it will just return self, so NOT a copy. Something to be aware of, hence the warning... but at least you can silent it now!

rhugonnet:
You can merge I think.
Resolves #102
Maybe #106 ?

PR #108, opened 2021-03-11
Satimg fix

rhugonnet:


adehecq:
Du coup, c'est pas un problème la nouvelle syntaxe?

rhugonnet:
> Du coup, c'est pas un problème la nouvelle syntaxe?

No impact on tuples as they don't support assignment. This is a good way of creating a N-size tuple of value X.

erikmannerfelt (approved):

adehecq:
I don't see what changed apart from a deleted line!

PR #109, opened 2021-03-11
Add copy method for SatelliteImage class (overrides Raster copy)

rhugonnet:
Add subclass copy method + associated tests
Resolves issue in xdem


geoutils/satimg.py
```diff
@@ -314,4 +315,20 @@ def __parse_metadata_from_fn(self,silent=False):
     def __parse_metadata_from_file(self,fn_meta):
         pass
 
+    def __copy__(self):
```
erikmannerfelt, line 318: This will have the exact same problem as I mentioned in xdem:

> A test should be added to assert the consistency of the `.data` attribute.
>
> A quick version would be:
> ```python
> dem1 = DEM("path/to/file")
> dem1.data += 5
>
> dem2 = dem1.copy()
>
>assert np.mean(np.abs(dem1.data - dem2.data)) < 0.001
> ```
> And I suspect the above test will fail right now.

_Originally posted by @erikmannerfelt in https://github.com/GlacioHack/xdem/pull/46#discussion_r592968427_

The problem is that loading from the `filename` attribute will not retain intermediate changes to the `data` attribute.

erikmannerfelt: I rather suggest something along:

```python
def copy(self):
    img_copy = self.super().copy()
    # copy all the SatelliteImage attributes somehow, maybe with a __dict__comparison
    for key in self.__dict__:
           if key in img_copy.__dict__:
               continue
           img_copy.__dict__[key] = self.__dict__[key]
```
I don't know if the above will work though, because it may be a Raster instance with SatelliteImage methods and attributes!

tests/test_satimg.py
```diff
@@ -23,6 +22,34 @@ def test_load_subclass(self):
...
+
+        img2 = copy.copy(img)
+        img3 = img.copy()
+
+        # check the copied object is indeed a SatelliteImage
+        assert isinstance(img2,geoutils.satimg.SatelliteImage)
+        assert isinstance(img3,geoutils.satimg.SatelliteImage)
+
+        # check all immutable attributes are equal
+        georaster_attrs = ['bounds', 'count', 'crs', 'dtypes', 'height', 'indexes', 'nodata',
```
adehecq, line 39: There's a list of these attributes at the top of georaster, so better to use that one (in case it's updated)

tests/test_satimg.py
```diff
@@ -23,6 +22,34 @@ def test_load_subclass(self):
...
+                         'res', 'shape', 'transform', 'width']
+        satimg_attrs =  ['satellite', 'sensor', 'product', 'version', 'tile_name', 'datetime']
+
+        all_attrs = georaster_attrs + satimg_attrs
+
+        for attr in all_attrs:
+            assert img2.__getattribute__(attr) == img.__getattribute__(attr)
+
+        # check data array
+        assert np.count_nonzero(~img.data == img2.data) == 0
```
adehecq, line 49: again, you can use np.all(img.data == img2.data) which is easier to understand

erikmannerfelt: May I suggest `np.array_equal(img.data, img2.data, equal_nan=True)` instead?
This is why:
```python
In [1]: arr1 = np.zeros(3)

In [2]: arr1[0] = np.nan

In [3]: arr2 = arr1.copy()

In [4]: arr1 == arr2
Out[4]: array([False,  True,  True])

In [5]: np.all(arr1 == arr2)
Out[5]: False
```
The `__eq__` method in Numpy arrays doesn't like NaNs!

erikmannerfelt:
With @adehecq 's PR, I think the copy method works without your PR, @rhugonnet.

This is with the latest geoutils and xdem/main
```python:
In [1]: import xdem

In [2]: import numpy as np

In [3]: dem = xdem.dem.DEM("examples/Longyearbyen/data/DEM_2009_ref.tif")
No metadata could be read from filename.

In [4]: dem2 = dem.copy()

In [5]: dem.data += 1

In [6]: assert not np.array_equal(dem.data, dem2.data, equal_nan=True)
```
This passes, meaning the copy method should work correctly.

tests/test_satimg.py
rhugonnet: You mean calling it from georaster?

rhugonnet: Done. Calling all attributes from the class from Raster, SatelliteImage and DEM. Some have to be removed in the test for Raster.

tests/test_satimg.py
rhugonnet: Done everywhere.

rhugonnet:
**Copy of the description in the same PR in xdem** (for whichever you read first)
I have:

squash-merged Amaury's PR into geoutils
merged with mine,
modified the __copy__ Subclass methods to simply return the object out of super().copy() (corrected by Amaury),
added a loop for Subclass attributes to be copied (all immutable so it's a simple = for now). Because of this immutable/mutable characteristic that cannot be planned, we might have to add some specific copy() methods to newer objects in the future. Thus I didn't not use the "dict" attribute solution of @erikmannerfelt and kept the hardcoded list of immutable attributes. This list of attributes is written only once just before the start of the class, and is called to be used by copy(), or by test methods (as suggested by Amaury).
kept the copy() wrapper based on the package copy.
implemented your suggestions in test_georaster, test_satimg and test_dem (not in this PR) : np.array_equal + r1 is not r2. The list of attributes to check on is imported directly from the Classes. For georaster, 3 attributes cannot be checked by a simple eq statement and have to be remove from list for obvious reasons (e.g., driver).

All tests passing!

tests/test_satimg.py
```diff
@@ -24,6 +24,47 @@ def test_load_subclass(self):
...
+
+        # Check dataset_mask array
+        assert np.all(r.data.mask == r2.data.mask)
+
+        # Check that if r.data is modified, it does not affect r2.data
+        r.data += 5
+        assert not np.array_equal(r.data, r2.data, equal_nan=True)
+
+        # Check that both have same output type
+        assert type(r) == type(r2)
```
adehecq, line 66: Also redundant with l 43.

adehecq:
Looks good!

adehecq (approved):

Issue #113, opened 2021-03-12
Add __eq__ method to Raster class


erikmannerfelt:
Right now, the "==" operator works in a quite unexpected way:
```python
In [1]: raster = gu.georaster.Raster("examples/Longyearbyen/data/DEM_2009_ref.tif")

In [2]: raster2 = gu.georaster.Raster("examples/Longyearbyen/data/DEM_2009_ref.tif")

In [2]: raster == raster2
Out[2]: False
```
I would argue that the above comparison should be true. It could be by adding an `__eq__(self, other)` method to the Raster class. I think this would work (EDIT: It passes my test in the PR #114 ):

```python
def __eq__(self, other: Raster) -> bool:
    return all([
        np.array_equal(self.data, other.data, equal_nan=True),
        self.transform == other.transform,
        self.crs == other.crs,
        self.nodata == other.nodata
    ])
```
Other comparisons could be added, such as the `filename` and `matches_disk` attributes. I would argue that only the attributes in my example should be tested, because the others are unique to Rasters loaded from disk (would not work on a Raster made from `Raster.from_array()`).

Such an addition would also be great for testing purposes (see related discussions in xdem: [46](https://github.com/GlacioHack/xdem/pull/46) and [43](https://github.com/GlacioHack/xdem/issues/43), and #112 on geotuils):

```python
raster3 = raster2.copy()

assert raster3 == raster2  # This should pass (but won't right now)

raster3.data += 5

assert raster3 != raster2  # This should also pass (will right now, but not for the right reason!)
```

rhugonnet:
Closed by #114

Issue #73, opened 2021-02-19
Add overloading methods to Raster


atedstone:
We would like to be able to support operations like this:

```python
im1 = Raster(...)
im2 = Raster(...)
combined = im1 + im2
```

At the moment, this requires a fairly unsemantic approach:

```python
combined = im1.copy()
combined.data = combined.data + im2.data
```

Proposal: add operator overloading methods to the `Raster` class, e.g. `__add__`, `__multiply__`. See e.g. https://stackabuse.com/overloading-functions-and-operators-in-python/ for more info.

adehecq:
Now done for add/sub and neg with PR #173.
Still waiting to add mult/div to close this issue.

PR #114, opened 2021-03-12
Add __eq__ method and tests

erikmannerfelt:
Please see #113 for the motivation.

### DIFF
--- a/geoutils/georaster.py
+++ b/geoutils/georaster.py
@@ -203,6 +203,20 @@ class Raster(object):
         """ Provide string of information about Raster. """
         return self.info()
 
+    def __eq__(self, other) -> bool:
+        """Check if a Raster's data and georeferencing is equal to another."""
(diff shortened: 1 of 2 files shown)


geoutils/georaster.py
```diff
@@ -203,6 +203,20 @@ def __str__(self):
         """ Provide string of information about Raster. """
         return self.info()
 
+    def __eq__(self, other) -> bool:
```
rhugonnet, line 206: so this would mean the data + CRS are equal, but not necessarily on disk (could be a copy) ?!

rhugonnet: Nvm I see this in your description now

rhugonnet (approved):
Looks good to me :)

adehecq:
Nice! Looks good to me too.

rhugonnet:
All good, merging (as I'm not sure @erikmannerfelt has the rights yet)

rhugonnet:
Resolves #113

atedstone:
Starts to resolve #73

PR #116, opened 2021-03-12
Fix issue with from_array and add tests

adehecq:
Still need to add the tests.


erikmannerfelt (approved):
I love the simplicity of the fix!

tests/test_georaster.py
```diff
@@ -121,9 +131,17 @@ def test_copy(self):
             assert r.__getattribute__(attr) == r2.__getattribute__(attr)
 
         # Check data array
-        assert np.count_nonzero(~r.data == r2.data) == 0
+        assert np.all(r.data == r2.data)
```
erikmannerfelt, line 134: This  won't work with NaNs, as previously mentioned. The assertion below is safer
```python
assert np.array_equal(r.data, r2.data, nan_equal=True)
```

tests/test_georaster.py
```diff
@@ -121,9 +131,17 @@ def test_copy(self):
...
+
         # Check dataset_mask array
-        assert np.count_nonzero(~r.dataset_mask() == r2.dataset_mask()) == 0
+        assert np.all(r.data.mask == r2.data.mask)
+
+        # Check that if r.data is modified, it does not affect r2.data
+        r.data += 5
+        assert not np.all(r.data == r2.data)
```
erikmannerfelt, line 141: Same as above

tests/test_georaster.py
```diff
@@ -105,14 +105,24 @@ def test_downsampling(self):
...
+        Test that the copy method works as expected. In particular
+        when copying r to r2:
+        - creates a new memory file
+        - if r.data is modified and r copied, the updated data is copied
+        - if r is copied, r.data changed, r2.data should be unchanged
+        """
+        # Open dataset, update data and make a copy
         r = gr.Raster(datasets.get_path("landsat_B4"))
+        r.data += 5
         r2 = r.copy()
```
erikmannerfelt, line 119: In addition, one could check:
```python
assert r is not r2
```
This will raise an error if it's the same object:
```python
a = [1, 2]
b = a

assert a is b  # This passes

c = a.copy()

assert a is not c   # This also passes
```

tests/test_satimg.py
```diff
@@ -78,3 +79,28 @@ def test_sw_tile_naming_parsing(self):
...
+        - if r.data is modified and r copied, the updated data is copied
+        - if r is copied, r.data changed, r2.data should be unchanged
+        """
+        # Open dataset, update data and make a copy
+        r = si.SatelliteImage(datasets.get_path("landsat_B4"))
+        r.data += 5
+        r2 = r.copy()
+
+        # Check data array
+        assert np.all(r.data == r2.data)
```
erikmannerfelt, line 96: As mentioned above, this works poorly with NaNs.

erikmannerfelt (commented):
Looking good! I just have some small comments.

I ran all the tests from this PR on my machine and they pass.

tests/test_georaster.py
rhugonnet: good to know

rhugonnet: Amaury I let you change those, I'll do the same in `test_satimg` and `test_dem`

tests/test_georaster.py
rhugonnet: Will also add this in `test_satimg` and `test_dem`

rhugonnet (approved):
I'll modify my PR to only copy the attributes of the subclass and, with those changes, all should work.

tests/test_georaster.py
rhugonnet: @adehecq actually let me change those everywhere, I'll squash and merge

tests/test_georaster.py
adehecq: I added a new test in test_satimg, hope you noticed

Issue #111, opened 2021-03-11
Allow SatImg creation from a Raster object, possibly without a filename


rhugonnet:
Mirrors [https://github.com/GlacioHack/xdem/issues/43](https://github.com/GlacioHack/xdem/issues/43) for SatImg. Should be able to do something like:
```python
SatelliteImage(Raster)
```
Currently only this works if `filename` is defined:
```python
SatelliteImage(Raster.filename)
```

rhugonnet:
Closed by #118

PR #118, opened 2021-03-12
Allow Raster/SatelliteImage class init through same or parent Class + tests

rhugonnet:
We can now input Raster and SatelliteImage (and DEM in xdem PR) with their own class or that of parent. Also added rio.io.DatasetReader as possible input for Raster.
Added tests checking all loading work, all attributes are equal and all arrays are equals (passing the same Class: returns the same object).
Example, we can do this (more details covered in the tests):
```python
# get SatelliteImage from Raster
r = gr.Raster(fn_img)
img = si.SatelliteImage(r)
# get SatelliteImage from SatelliteImage
img2 = si.SatelliteImage(img)
```
And for passing back the object in `__init__` of each Class, I used @erikmannerfelt 's code (I knew this would be useful!):
```python
# If Raster is passed, simply point back to Raster
if isinstance(filename, Raster):
    for key in filename.__dict__:
        setattr(self,key,filename.__dict__[key])
    return
```
Voilà! And all tests passing ;)

Will PR the same in xdem and point to this description.


rhugonnet:
Resolves #111

erikmannerfelt (approved):
Looks great!

Only slightly related to this PR, but I think the `filename` argument name is a bit weird, especially with these additions. It's only sometimes that it's a filename! It could have been called `fp_or_data` or something along those lines (but less ugly)... Anyway, that's something for later I guess.

PR #119, opened 2021-03-14
Fix match disk attribute

### LINKED ISSUE #112: 'Raster.matches_disk' is not updated when modifying the 'data' attribute
A minimal example:
```python
In [1]: raster = gu.georaster.Raster("examples/Longyearbyen/data/DEM_2009_ref.tif")

In [2]: raster.data += 5

In [3]: assert not raster.matches_disk
AssertionError                            Traceback (most recent call last)
<ipython-input-26-067f9d70f291> in <module>
----> 1 assert not raster.matches_disk

AssertionError:
```

rhugonnet:
Added `matches_disk` as a property, fixed its implementation, and it now gets updated with the `data.setter`.
Resolves #112 

However, the current behaviour doesn't really match the name: if any operation is performed the tag will be updated.
After some testing, I see no way around this "simplified" behavior, because:
1. If I add a check in the `data.setter` such as:
```python
#check file already matched disk
if self._matches_disk:
    # check if arrays are equal
    if not np.array_equal(self._data,new_data,equal_nan=True):
      self._matches_disk = False
```
This will work with classic operation such as:
```python
r.data = r.data + 5
r.matches_disk
Out[8]: False
```
but won't with in-place operations:
```python
r.data += 5
r.matches_disk
Out[8]: True
```
Here the instance gets rewritten so there is nothing to compare to.

2. If I add a matches_disk property that reads the data on the disk again and checks if everything is equal, we need to load the entire array, and go again through all properties (load, etc...). Currently this is not even possible because the "band" parameter of `__init__` is not passed and so the function `load` cannot be reproduced (I haven't tested but using the Raster class from within a property might trigger a circular issue).

I advocate for using the current attribute, and renaming to something more fair such as "isUpdated" or "isOriginal" or any better name that comes to your mind (basically: has any operation been performed on it, no matter if it still matches disk or not).

We could also add an independent "matches_disk" property that checks the disk only if called by the user (additional RAM/computing), but this means changing several things upstream.


erikmannerfelt:
Hmm, what a mess!

I feel like a `matches_disk` attribute needs to be absolutely rock-solid if it should exist. In the suggested implementation, does the `r.data += 5` update the flag or not? You say that it doesn't, but it looks like it does in your test. Or does the test fail right now?

Also, what happens if the transform, crs or nodata value is modified? Does `matches_disk` get updated accordingly?

I have a suggestion which may solve these issues: using hashes. How does git keep track of if a file has changed? It uses sha-1 hashes to compare. A geoutils example is:
```python
In [1]: import geoutils as gu

In [4]: r = gu.georaster.Raster(gu.datasets.get_path("landsat_B4"))

In [5]: import hashlib

In [6]: hash1 = hash((hashlib.sha1(r.data), r.transform, r.crs, r.nodata))

In [7]: r.data += 1

In [8]: hash2 = hash((hashlib.sha1(r.data), r.transform, r.crs, r.nodata))

In [9]: assert hash1 != hash2
```
which passes, because `r` has been changed and doesn't have the same "signature" any more.

The commands in `In [6]` and `In [8]` took me a while to figure out, because I found that a combination of `hashlib` and the builtin `hash()` function had to be used... `hashlib.sha1()` requires that the object supports the "buffer API" (whatever that means...), which `np.ndarray`s and `np.ma.masked_array`s do, but the transform, CRS and nodata objects don't! On the other hand, `hash()` doesn't work on arrays... A compromise is to take the `hash()` of `hashlib.sha1(r.data)` together with all the other attributes in a tuple (don't worry about the double-hashing; it's done frequently in cryptography).

The way this could work in Raster is to hash each Raster attribute in the `__init__()` and then define the `matches_disk` to generate and compare a new hash:

```python
class Raster(...):
    def __init__(self, ...):
         # all of the stuff

         # This should be near the very end
         self._disk_hash =  hash((hashlib.sha1(self.data), self.transform, self.crs, self.nodata))

    @property
    def matches_disk(self) -> bool:
        new_hash = hash((hashlib.sha1(self.data), self.transform, self.crs, self.nodata))
        return self._disk_hash == new_hash
```
I don't know what the performance overhead is, but I imagine it to be minimal/nonexistent. My computer has computed and compared hundreds or thousands of hashes by writing this message in `https`!

rhugonnet:
This looks quite interesting!
We could implemented it as is, but again it won't match the `matches_disk` term because you can simply re-instance or point to a same object and the hash code will change:
```python
import hashlib
hash1 = hash((hashlib.sha1(r.data), r.transform, r.crs, r.nodata))
r.data += 0
hash2 = hash((hashlib.sha1(r.data), r.transform, r.crs, r.nodata))
print(hash1 != hash2)
True
```
So I think we should change `matches_disk` into a method that goes and check on the disk (only way), and rename this functionality (which is still valuable) into something else like `is_updated`.
What do you think?

adehecq:
> This looks quite interesting!
> We could implemented it as is, but again it won't match the `matches_disk` term because you can simply re-instance or point to a different instance with the same data and the hash code will change:
> 
> ```python
> import hashlib
> hash1 = hash((hashlib.sha1(r.data), r.transform, r.crs, r.nodata))
> r.data += 0
> hash2 = hash((hashlib.sha1(r.data), r.transform, r.crs, r.nodata))
> print(hash1 != hash2)
> True
> ```
> 
> So I think we should change `matches_disk` into a method that goes and check on the disk (only way), and rename this functionality (which is still valuable) into something else like `is_updated`.
> What do you think?

I think the hash approach is interesting too. 
Maybe @atedstone has a better idea of what the use of matches_disk is. Personally, I just see it as a way to have a warning flag that the data stored in self might differ from what was originally on disk, because it has been modified, or reprojected etc. 
What's your use case for it @rhugonnet and @erikmannerfelt ?

adehecq:
Also, self.isOriginal woudl make the most sense maybe? If any data/metadata has been altered (even if it hasn't really changed the value like adding 0), it should be False, just to say "be careful, this is not the original data".

erikmannerfelt:
> > This looks quite interesting!
> > We could implemented it as is, but again it won't match the `matches_disk` term because you can simply re-instance or point to a different instance with the same data and the hash code will change:
> > ```python
> > import hashlib
> > hash1 = hash((hashlib.sha1(r.data), r.transform, r.crs, r.nodata))
> > r.data += 0
> > hash2 = hash((hashlib.sha1(r.data), r.transform, r.crs, r.nodata))
> > print(hash1 != hash2)
> > True
> > ```
> > 
> > 
> > So I think we should change `matches_disk` into a method that goes and check on the disk (only way), and rename this functionality (which is still valuable) into something else like `is_updated`.
> > What do you think?
> 
> I think the hash approach is interesting too.
> Maybe @atedstone has a better idea of what the use of matches_disk is. Personally, I just see it as a way to have a warning flag that the data stored in self might differ from what was originally on disk, because it has been modified, or reprojected etc.
> What's your use case for it @rhugonnet and @erikmannerfelt ?

In your case, @adehecq, I think hashing is the best bet so far. As @rhugonnet showed, however, it is not waterproof, but it still gives an indication of if any modification (significant or not) has been made.

I think this ends up mattering based on what it will be used for. I think a hash approach would be similar to the pseudocode: `Raster is 'the file on disk'`, while the `__eq__()` (`Raster == Raster2`) indicates practical equality. A user could always do `assert original_raster == modified_raster`, which will always be the most robust approach.

rhugonnet:
> Maybe @atedstone has a better idea of what the use of matches_disk is. Personally, I just see it as a way to have a warning flag that the data stored in self might differ from what was originally on disk, because it has been modified, or reprojected etc.
> What's your use case for it @rhugonnet and @erikmannerfelt ?

Same. I don't see much use in a fully robust `matches_disk` functionality. Such a warning flag is enough.
Should I rename in "is_original"? Better naming idea @erikmannerfelt ? Maybe this needs another bathroom break

erikmannerfelt:
> > Maybe @atedstone has a better idea of what the use of matches_disk is. Personally, I just see it as a way to have a warning flag that the data stored in self might differ from what was originally on disk, because it has been modified, or reprojected etc.
> > What's your use case for it @rhugonnet and @erikmannerfelt ?
> 
> Same. I don't see much use in a fully robust `matches_disk` functionality. Such a warning flag is enough.
> Should I rename in "is_original"? Better naming idea @erikmannerfelt ? Maybe this needs another bathroom break

Or maybe `is_modifed`? Go ask the toilet!

`is_original` would work too.

rhugonnet:
Toilet is telling me the most exact would be `is_untouched`. You guys agree?

atedstone:
`is_modified` probably my preference over `un_touched`. 

The original `matches_disk` was implemented on day 1 of GlacioHack to provide some indication of whether the in-memory version is the same as on disk - but was never meant to be waterproof so I agree that the name of this attribute should be changed.

just not to `isOriginal` etc - no camel case allowed around here ;)

adehecq:
> Toilet is telling me the most exact would be `is_untouched`. You guys agree?

Your toilets are telling you weird stuff!! :'-D
I prefer is_modified too.

rhugonnet:
OK, let's go with `is_modified`! I will use this opportunity to change the `isLoaded` that was written in camel case on day 1 !

erikmannerfelt:
> OK, let's go with `is_modified`! I will use this opportunity to change the `isLoaded` that was written in camel case on day 1 !

Lovely!!

rhugonnet:
@erikmannerfelt I'm curiously getting hash changes without any update in the data, just printing a few seconds apart.
```python
print(hash((hashlib.sha1(r.data), r.transform, r.crs, r.nodata)))
   ...: 
1592015780544301241
print(hash((hashlib.sha1(r.data), r.transform, r.crs, r.nodata)))
   ...: 
-8879540271875856243
```
Do you have this? Any idea where it could come from? My implementation:
```python
    @property
    def is_modified(self):
        """
        Getter method for the _method class member

        Returns: boolean: the _method member of this instance of Raster
        """
        if not self._is_modified:
            new_hash = hash((hashlib.sha1(self._data), self.transform, self.crs, self.nodata))
            self._is_modified = not (self._disk_hash == new_hash)

        return self._is_modified
```
and the self._disk_hash is the last thing that is defined in `__init__`

rhugonnet:
The test pass in command-line. It's only in PyCharm. Maybe it's the way PyCharm manages objects?!

erikmannerfelt:
> @erikmannerfelt I'm curiously getting hash changes without any update in the data, just printing a few seconds apart.
> 
> ```python
> print(hash((hashlib.sha1(r.data), r.transform, r.crs, r.nodata)))
>    ...: 
> 1592015780544301241
> print(hash((hashlib.sha1(r.data), r.transform, r.crs, r.nodata)))
>    ...: 
> -8879540271875856243
> ```
> 
> Do you have this? Any idea where it could come from? My implementation:
> 
> ```python
>     @property
>     def is_modified(self):
>         """
>         Getter method for the _method class member
> 
>         Returns: boolean: the _method member of this instance of Raster
>         """
>         if not self._is_modified:
>             new_hash = hash((hashlib.sha1(self._data), self.transform, self.crs, self.nodata))
>             self._is_modified = not (self._disk_hash == new_hash)
> 
>         return self._is_modified
> ```
> 
> and the self._disk_hash is the last thing that is defined in `__init__`

I found a simpler approach without hashlib! 

```python
hash((self._data.tobytes(), self.transform, self.crs, self.nodata))
```
There could be a small overhead here because of the `ndarray -> str` conversion, but it's without delay on my machine at least.

erikmannerfelt:
> > @erikmannerfelt I'm curiously getting hash changes without any update in the data, just printing a few seconds apart.
> > ```python
> > print(hash((hashlib.sha1(r.data), r.transform, r.crs, r.nodata)))
> >    ...: 
> > 1592015780544301241
> > print(hash((hashlib.sha1(r.data), r.transform, r.crs, r.nodata)))
> >    ...: 
> > -8879540271875856243
> > ```
> > 
> > 
> > Do you have this? Any idea where it could come from? My implementation:
> > ```python
> >     @property
> >     def is_modified(self):
> >         """
> >         Getter method for the _method class member
> > 
> >         Returns: boolean: the _method member of this instance of Raster
> >         """
> >         if not self._is_modified:
> >             new_hash = hash((hashlib.sha1(self._data), self.transform, self.crs, self.nodata))
> >             self._is_modified = not (self._disk_hash == new_hash)
> > 
> >         return self._is_modified
> > ```
> > 
> > 
> > and the self._disk_hash is the last thing that is defined in `__init__`
> 
> I found a simpler approach without hashlib!
> 
> ```python
> hash((self._data.tobytes(), self.transform, self.crs, self.nodata))
> ```
> 
> There could be a small overhead here because of the `ndarray -> str` conversion, but it's without delay on my machine at least.

Oh and I should mention that it returns the same hash every time.

rhugonnet:
@erikmannerfelt this works like a charm. I've implemented it and it works also as intended for `r.data += 0` or`r.data = r.data + 0` where the `is_modified` stays with a value of `False`.
I have changed all nomenclature, also for `isLoaded`. And modified all tests, everything passing.
Approval, @adehecq ?

geoutils/georaster.py
```diff
@@ -124,7 +124,7 @@ def __init__(self, filename, bands=None, load_data=True, downsampl=1,
             self.is_loaded = True
             if isinstance(filename, str):
                 self._is_modified = False
-                self._disk_hash = hash((hashlib.sha1(self._data), self.transform, self.crs, self.nodata))
+                self._disk_hash = hash((self._data.tobytes(), self.transform, self.crs, self.nodata))
```
erikmannerfelt, line 127: Is this triggered when creating from another Raster instance as well?

erikmannerfelt (approved):
Nice! I just have one question

geoutils/georaster.py
rhugonnet: No, this only works with input filename. For the rest, the `is_modified` will stay as `None`.

erikmannerfelt: It's my afternoon brain thinking. Of course it shouldn't be triggered without a file on the disk!

adehecq:
> @erikmannerfelt this works like a charm. I've implemented it and it works also as intended for `r.data += 0` or`r.data = r.data + 0` where the `is_modified` stays with a value of `False`.
> I have changed all nomenclature, also for `isLoaded`. And modified all tests, everything passing.
> Approval, @adehecq ?

I approve a posteriori!

PR #120, opened 2021-03-15
Fix interp_points + add assertion in test

### LINKED ISSUE #69: test_interp does not contain any assert statements
under the CI testing approach, all tests need asserts so that they can be checked automatically - we are no longer relying on checking the outputs from print statements manually.

### LINKED ISSUE #100: Check interp_points with news coords() and related tests

rhugonnet:
Changed `interp_points()` to `scipy.ndimage.map_coordinates` instead of griddata. Commented the rest of the code that might allow to write a mutiprocessing loop if needed in the future.
Added assertions to `test_interp()`.
All tests passing.
Resolves #100
Resolves #69

One current limitation: we need to know if the pixel coordinates refers to center or corner (is_area or is_point). Right now I shift x/y by 0.5 pixel assuming it's the center, which works for our test Raster.
For robustness, we'd need to add the is_area or is_point info from GDAL into the Raster class. This is necessary to further improve the inter_points method.


rhugonnet:
Resolves #100 and #69

geoutils/georaster.py
```diff
@@ -15,6 +15,8 @@
 from rasterio.warp import Resampling
 from rasterio.plot import show as rshow
 import matplotlib
+import pyproj
+from scipy.ndimage import map_coordinates
```
erikmannerfelt, line 19: `scipy` is not in the `setup.py` requirements.

geoutils/georaster.py
```diff
@@ -1250,41 +1254,62 @@ def interp_points(self,pts,nsize=1,mode='linear',band=1):
...
+
+        i, j = self.ds.index(x, y, op=np.float32)
+        i = np.array(i) - 0.5
+        j = np.array(j) - 0.5
+
+        if i.ndim == 0:
+            i = [i]
+            j = [j]
+
+        ind_invalid = [self.outside_image(j[k],i[k],index=True) for k in range(len(i))]
```
erikmannerfelt, line 1278: Have you tried using `np.vectorize` here?

Something like:
```python

ind_invalid = np.vectorize(lambda x, y: self.outside_image(x, y, index=True))(j, i)
```
Using `np.vectorize()` is often orders of magnitude faster than python loops.

geoutils/georaster.py
```diff
@@ -1250,41 +1254,62 @@ def interp_points(self,pts,nsize=1,mode='linear',band=1):
...
-                        print(x)
-                        print(y)
-                        print(z)
-                        zint = np.nan
-                else:
-                    zint = np.nanmean(z.flatten())
-                rpts.append(zint)
-        rpts = np.array(rpts)
+        # get coordinates
+        x, y = list(zip(*pts))
```
erikmannerfelt, line 1262: Does this have an overhead?
An alternative may be:
```python
x, y = np.asarray(pts, dtype=float) # maybe .T has to be added. Not sure
```

erikmannerfelt (commented):
I only have some performance suggestions.

geoutils/georaster.py
rhugonnet: Done!

geoutils/georaster.py
rhugonnet: Done

geoutils/georaster.py
```diff
@@ -1218,23 +1220,25 @@ def outside_image(self, xi,yj, index=True):
         :returns is_outside: True if ij is outside of the image.
         """
         if not index:
-            xi,xj = self.xy2ij(xi,yj)
+            xi,yj = self.xy2ij(xi,yj)
```
adehecq, line 1223: It's strange to have the double name x and i or y and j. It's a bit redundant...

geoutils/georaster.py
```diff
@@ -1218,23 +1220,25 @@ def outside_image(self, xi,yj, index=True):
...
 
-    def interp_points(self,pts,nsize=1,mode='linear',band=1):
+    def interp_points(self,pts,input_latlon=False,nsize=1,mode='linear',band=1, **kwargs):
 
         """
         Interpolate raster values at a given point, or sets of points.
 
        :param pts: Point(s) at which to interpolate raster value. If points fall outside of image,
        value returned is nan.'
        :type pts: array-like
```
adehecq, line 1239: What exactly is the expected format? From l 1258 it seems like it is ( (x1, y1), (x2, y2)...)?
If so, mention that shape should be (N, 2).

geoutils/georaster.py
```diff
@@ -1250,41 +1254,67 @@ def interp_points(self,pts,nsize=1,mode='linear',band=1):
...
                         'nearest'], "mode must be mean, linear, cubic, quintic or nearest."
 
-        rpts = []
+        # get coordinates
+        x, y = list(zip(*pts))
 
-        #TODO: might need to check if coordinates are center or point in the metadata here...
+        # if those are in latlon, convert to Raster crs
+        if input_latlon:
+            init_crs = pyproj.CRS(4326)
```
adehecq, line 1262: If we don't, we should definitely have a small function for this type of operation in projtools !

geoutils/georaster.py
```diff
@@ -1250,41 +1254,67 @@ def interp_points(self,pts,nsize=1,mode='linear',band=1):
...
+            dest_crs = pyproj.CRS(self.crs)
+            transformer = pyproj.Transformer.from_crs(init_crs,dest_crs)
+            x, y = transformer.transform(x,y)
 
+        i, j = self.ds.index(x, y, op=np.float32)
+        # # necessary because rio.Dataset.xy does not return Iterable for a single point
+        if not isinstance(i, Iterable):
+            i, j = (np.asarray([i,]), np.asarray([j,]))
+        else:
+            i, j = (np.asarray(i), np.asarray(j))
```
adehecq, line 1272: what's the difference between np.array and np.asarray?

geoutils/georaster.py
```diff
@@ -1250,41 +1254,67 @@ def interp_points(self,pts,nsize=1,mode='linear',band=1):
...
-                else:
-                    zint = np.nanmean(z.flatten())
-                rpts.append(zint)
-        rpts = np.array(rpts)
+        # AREA_OR_POINT GDAL attribute, i.e. does the value refer to the lower left corner (AREA) or the center of pixel (POINT)
+        # This has no influence on georeferencing, it's only about the interpretation of the raster values, and thus only
+        # affects sub-pixel interpolation
+
+        # if point, shift index by half a pixel
+        if self.ds.tags()['AREA_OR_POINT'] == 'Point':
```
adehecq, line 1279: I think it would be better to implement this in xy2ij and use it.
You can find the info about center/corner in the dataset tags: `self.ds.tags()`

geoutils/georaster.py
```diff
@@ -1250,41 +1254,67 @@ def interp_points(self,pts,nsize=1,mode='linear',band=1):
...
+            i -= 0.5
+            j -= 0.5
+
+        ind_invalid = np.vectorize(lambda k1, k2: self.outside_image(k1,k2,index=True))(j,i)
+
+        rpts = map_coordinates(self.data[band-1,:,:], [i,j], **kwargs)
```
adehecq, line 1285: This makes the assumption that the full data has been loaded, in particular all bands...

tests/test_georaster.py
```diff
@@ -213,7 +213,7 @@ def test_interp(self):
 
         xmin, ymin, xmax, ymax = r.ds.bounds
 
-        # testing interp, find_value, and read when it falls right on the coordinates
+        # Test for several points: interp, value_at_coords, and read should be the same right on the coordinates
```
adehecq, line 216: It would be good to check that it works fine after the data has been reprojected, cropped or downsampled...

geoutils/georaster.py
rhugonnet: This is because input can be coordinate or index

geoutils/georaster.py
rhugonnet: I think np.asarray is the official one for convert all types of Iterables: https://numpy.org/doc/stable/reference/generated/numpy.asarray.html
Don't really know the difference, maybe @erikmannerfelt can help here

erikmannerfelt: From the documentation linked by @rhugonnet:

> Array interpretation of a. No copy is performed if the input is already an ndarray with matching dtype and order. If a is a subclass of ndarray, a base class ndarray is returned.

It's a zero-copy operation, assuming the array is already an array. Otherwise it will convert to an array. Calling `np.array(arr)` makes for potential copying, which is unnecessary overhead.

geoutils/georaster.py
erikmannerfelt: Could 
```python
if not self.isLoaded:
    self.load(1)

```
be added above to fix this?

Speaking of camelCase, what's up with that variable??

geoutils/georaster.py
adehecq: Good to know!

geoutils/georaster.py
adehecq: That's not sufficient. If isLoaded is True, but only one band was loaded, it won't work...
That's one more argument to save the info of loaded band in an attribute...

rhugonnet: I guess I'll add to all the other "is_loaded" issues :D

geoutils/georaster.py
rhugonnet: Done. `self.ds.tags()` was used already

geoutils/georaster.py
```diff
@@ -1167,20 +1170,44 @@ def coords(self, offset='corner', grid=True):
...
         Return row, column indices for a given x,y coordinate pair.
 
         :param x: x coordinates
         :type x: array-like
         :param y: y coordinates
         :type y: array-like
+        :param op: operator to calculate index
+        :type op: operator
+        :param shift_area_or_point: shift according to AREA_OR_POINT GDAL attribute:
+         interpretation to what the raster value corresponds to (AREA = lower left or POINT = center)
```
adehecq, line 1184: area is actually upper left

geoutils/georaster.py
```diff
@@ -1167,20 +1170,44 @@ def coords(self, offset='corner', grid=True):
...
+        if not isinstance(i, Iterable):
+            i, j = (np.asarray([i, ]), np.asarray([j, ]))
+        else:
+            i, j = (np.asarray(i), np.asarray(j))
+
+        # AREA_OR_POINT GDAL attribute, i.e. does the value refer to the lower left corner (AREA) or the center of pixel (POINT)
+        # This has no influence on georeferencing, it's only about the interpretation of the raster values, and thus only
+        # affects sub-pixel interpolation
+
+        if shift_area_or_point:
```
adehecq, line 1202: It's a bit strange to have the area_or_point info but only use it in rare cases. Why not have an option 'area_or_point', which could be set to area, point or None. In the latter case, it would be set to the value in the metadata.
If you want to keep the same as current, you can just set default to 'area'.

tests/test_georaster.py
rhugonnet: Done for crop. This also changes the GDAL interpretation to "Area" instead of "Point" for our test sample. So I tested that this worked correctly as well.

geoutils/georaster.py
rhugonnet: You mean upper left in y/x dimension? Because y is inverted? So in indexes: Area = (0,0) and Point = (0.5,0.5), right? The new tests work that way.

adehecq: Yes, y axis is downward, so (0,0) is the upper left corner.

tests/test_georaster.py
```diff
@@ -209,46 +209,83 @@ def test_inters_img(self):
...
+        # at the center of pixel
         r = gr.Raster(datasets.get_path("landsat_B4"))
+        assert r.ds.tags()['AREA_OR_POINT'] == 'Point'
 
         xmin, ymin, xmax, ymax = r.ds.bounds
 
-        # testing interp, find_value, and read when it falls right on the coordinates
+        # We generate random points within the boundaries of the image
         xrand = (np.random.randint(low=0, high=r.ds.width, size=(10,))
                  * list(r.ds.transform)[0] + xmin + list(r.ds.transform)[0]/2)
```
adehecq, line 221: Why not have `r.ds.transform[0]*1.5` instead of `list(r.ds.transform)[0] + list(r.ds.transform)[0]/2`?

tests/test_georaster.py
```diff
@@ -209,46 +209,83 @@ def test_inters_img(self):
...
+        # the values interpolated should be equal
+        assert np.array_equal(np.array(list_z_ind,dtype=np.float32),rpts,equal_nan=True)
+
+        # Test there is no failure with random coordinates (edge effects, etc)
+        xrand = np.random.uniform(low=xmin, high=xmax, size=(1000,))
+        yrand = np.random.uniform(low=ymin, high=ymax, size=(1000,))
+        pts = list(zip(xrand, yrand))
+        rpts = r.interp_points(pts)
+
+        # SECOND, test after a crop: the Raster now has an Area interpretation, those should fall right on the integer
```
adehecq, line 249: It's weird that crop changes the area_or_point attribute...

rhugonnet: You think it comes from `_update()` ?

adehecq: You can test...

adehecq: If you can't find out why, maybe add as an issue.

rhugonnet:
No wonder I couldn't get my head around those tests. 
I've discovered a new interesting behaviour of the `rio.ds.index()` function:
```python
r.ds.index(x,y)
Out[33]: (75, 301)
r.ds.index(x,y,op=np.float32)
Out[34]: (75.0, 302.0)
```
Just by changing operator, you shift y by 1. Right now `test_interp()` works the way it is intended. But, when comparing to `value_at_coords()`, it only works if the operator is math.floor (lower index).

How should we go about this?

tests/test_georaster.py
rhugonnet: Removed

tests/test_georaster.py
rhugonnet: Added #127

erikmannerfelt:
> No wonder I couldn't get my head around those tests.
> I've discovered a new interesting behaviour of the `rio.ds.index()` function:
> 
> ```python
> r.ds.index(x,y)
> Out[33]: (75, 301)
> r.ds.index(x,y,op=np.float32)
> Out[34]: (75.0, 302.0)
> ```
> 
> Just by changing operator, you shift y by 1. Right now `test_interp()` works the way it is intended with `np.float32`. But, when comparing to `value_at_coords()`, it only works if the operator is math.floor (lower y index).
> 
> How should we go about this?

Hahaha what!? That should probably be added as an issue on rasterio, no? A quick search in their issues shows no result corresponding to this phenomenon.

Could this be circumvented for now, and then hopefully be fixed by rasterio in time?

rhugonnet:
Wait for the next level:
```python
r.ds.index(x,y,op=np.float32,precision=10)
Out[20]: (138.0, 504.0)
r.ds.index(x,y,op=np.float32)
Out[21]: (138.0, 504.0)
r.ds.index(x,y)
Out[22]: (138, 503)
r.ds.index(x,y,precision=0)
Out[23]: (137, 504)
r.ds.index(x,y,precision=1)
Out[24]: (137, 504)
r.ds.index(x,y,precision=10)
Out[25]: (138, 504)
```
The issue is that the "math.floor" argument is by default in the underlying function:
```python
def rowcol(transform, xs, ys, op=math.floor, precision=None):
```
which is wrapper by `index`

rhugonnet:
Need to test this in more details, but it looks like the "right" solution is the `np.float32` and the default setting with `math.floor` can be falsely rounding (with default `precision=None` or other) when you provide the exact coordinates:
```python
r.bounds
Out[27]: BoundingBox(left=478000.0, bottom=3088490.0, right=502000.0, top=3108140.0)
xmin = 478000
yax=3108140
ymax=3108140
r.ds.index(xmin,ymax)
Out[31]: (0, 0)
r.ds.index(xmin,ymax,op=np.float32)
Out[32]: (0.0, 0.0)
r.ds.index(xmin,ymax,op=np.float32,precision=10)
Out[33]: (0.0, 3.637979e-12)
r.ds.index(xmin,ymax,precision=10)
Out[34]: (0, 0)
r.ds.index(xmin,ymax,precision=1)
Out[35]: (-1, 0)
r.ds.index(xmin,ymax,precision=0)
Out[36]: (-1, 0)
r.ds.index(xmin,ymax)
Out[37]: (0, 0)
```

erikmannerfelt:
> Wait for the next level:
> 
> ```python
> r.ds.index(x,y,op=np.float32,precision=10)
> Out[20]: (138.0, 504.0)
> r.ds.index(x,y,op=np.float32)
> Out[21]: (138.0, 504.0)
> r.ds.index(x,y)
> Out[22]: (138, 503)
> r.ds.index(x,y,precision=0)
> Out[23]: (137, 504)
> r.ds.index(x,y,precision=1)
> Out[24]: (137, 504)
> r.ds.index(x,y,precision=10)
> Out[25]: (138, 504)
> ```
> 
> What's problematic is that the "math.floor" argument is by default in the underlying function, and is the one showing all those changes:
> 
> ```python
> def rowcol(transform, xs, ys, op=math.floor, precision=None):
> ```
> 
> which is wrapped by `index`

Call an exorcist!

This should definitely be an issue in rasterio, no? 

Also, why do they need `math.floor()` in the first place? Maybe it's new (and rio still supports py=2.7), but `round(1.1)` returns `1` (`type(round(1.1)) == int`).

adehecq:
I thought we had discussed this before, but I cannot find any mention to it. But I agree, the use of floor by default is a bit strange. Maybe float32 should be the default. We could add an optional argument opt to xy2ij too.

adehecq:
@rhugonnet what is the status of this??

adehecq:
@rhugonnet  Tomorrow, your first challenge will be to merge this PR! :-P

geoutils/georaster.py
rhugonnet: fixed in 37da8c2

geoutils/georaster.py
rhugonnet: Done in 
903c260

rhugonnet:
@erikmannerfelt @adehecq 
Defaulted to np.float32 as operator. Raises warning when other operator is used.
Modified area_or_point as optional input (None, 'Area', 'Point') that defaults to GDAL metadata AREA_OR_POINT fetched through `rasterio` when the value is None.
Updated all tests, passing.

erikmannerfelt:
There seems to be a merge conflict. Otherwise, it looks very good!

erikmannerfelt:
There should be an `Iterable` import here:
https://github.com/GlacioHack/GeoUtils/blob/5e2b975a1a281fb38e1695e58916b4eadddeaeb1/geoutils/georaster.py#L9

rhugonnet:
I've fixed some additional bugs after merging.
Now only need approval to merge the PR!

erikmannerfelt (approved):
Great job!!

Issue #76, opened 2021-02-19
Raster's shift method fails if data is not loaded


adehecq:
For example:
```
from geoutils import georaster as geor
img = geor.Raster('tests/data/LE71400412000304SGS00_B4_crop.TIF', load_data=False)
img.shift(20,10)
```
returns an AttributeError, 

> AttributeError                            Traceback (most recent call last)
> <ipython-input-15-38771ac4c269> in <module>
> ----> 1 img.shift(20,10)
> 
> ~/development/GlacioHack/GeoUtils/geoutils/georaster.py in shift(self, xoff, yoff)
>     575         meta.update({'transform': rio.transform.Affine(dx, b, xmin + xoff,
>     576                                                        d, dy, ymax + yoff)})
> --> 577         self._update(metadata=meta)
>     578 
>     579     def set_ndv(self, ndv, update_array=False):
> 
> ~/development/GlacioHack/GeoUtils/geoutils/georaster.py in _update(self, imgdata, metadata, vrt_to_driver)
>     251 
>     252         with memfile.open(**metadata) as ds:
> --> 253             ds.write(imgdata)
>     254 
>     255         self.memfile = memfile
> 
> rasterio/_io.pyx in rasterio._io.DatasetWriterBase.write()
> 
> AttributeError: 'NoneType' object has no attribute 'shape'
>

PR #126, opened 2021-03-15
Fix several bugs when data not loaded

adehecq:


geoutils/georaster.py
```diff
@@ -280,9 +280,14 @@ def data(self, new_data):
...
-        if new_data.shape != self._data.shape:
+        if self.is_loaded:
+            orig_shape = self._data.shape
+        else:
+            orig_shape = (self.count, self.height, self.width)
+
+        if new_data.shape != orig_shape:
             raise ValueError("New data must be of the same shape as\
- existing data: {}.".format(self.shape))
+ existing data: {}.".format(orig_shape))
```
erikmannerfelt, line 290: There might be something with GitHub's visualisation, but it seems to me like this line is unindented. Is that true?

rhugonnet (approved):
Looks good :D

geoutils/georaster.py
adehecq: It's correct, but if I indent, the text that is printed has a large space. How do I avoid this?

adehecq: I think I found it. Should I just add the quotes at the end and beginning of each newline?

adehecq: See commit 63e5717

adehecq:
Resolves #76

Issue #115, opened 2021-03-12
Re-think the from_array method


adehecq:
Currently, child classes of Raster also inherit the from_array function, which always returns a Raster instance. This is problematic for example with reproject, which returns a Raster instance:
Here's a minimum working example:
```
from geoutils import georaster as gr
from geoutils import datasets
from geoutils import satimg
img = satimg.SatelliteImage(datasets.get_path('landsat_B4'))
img2 = satimg.SatelliteImage(datasets.get_path('landsat_B4_crop'))
test = img.reproject(dst_ref=img2)
type(test)
```

adehecq:
Actually, from_array works as expected,a s it returns an object of the right type, see below. The problem of reproject must be elsewhere...
```
img_out = img.from_array(img.data, transform=img.transform, crs=img.crs)
type(img_out) == type(img)
```

adehecq:
We also need to add tests for this case, when solving the issue.

erikmannerfelt:
Maybe link this to #116 ?

adehecq:
Fixed by #116

PR #128, opened 2021-03-15
Modify input name + parse name for dataset + fix SatelliteImage copy signature

### LINKED ISSUE #110: Save original filename attribute when copying Raster in memory
Maybe a `.orig_filename` attribute? If I understood correctly, `_filename` would have to be related to .filename (to avoid users from setting it directly), so the naming wouldn't be a good fit.
The filename can be very useful for initialization of SatImg/DEM at a later stage.

### LINKED ISSUE #123: Change input parameter name of class and subclasses (filename)
`Raster`, `SatelliteImage` and `DEM` class can take many things other than file name, as raised by @erikmannerfelt.
Change to something more generic, like `input` ?

rhugonnet:
As pointed out by @erikmannerfelt, the naming `filename` for input parameter was not ideal.
I've switched to `filename_or_dataset` but we can adapt, fyi pandas uses `filename_or_buffer`. (its Ctrl + R if we want to change)

Now the `.filename` attribute is clearly differentiated. I removed redundant setting of filename to "None" (the parameter is defined at the start of the class). Now filename is parsed when user provides a rio.io.DatasetReader.

Resolves #123 
Resolves #110 (mostly by #115)


geoutils/georaster.py
```diff
@@ -43,13 +43,13 @@ class Raster(object):
     _is_modified = None
     _disk_hash = None
 
-    def __init__(self, filename, bands=None, load_data=True, downsampl=1,
```
erikmannerfelt, line 46: Just since wer're on revising attribute names, is `downsampl` an abbreviation of `downsample`? Feels like an abbreviation just for abbreviates' sake!

geoutils/satimg.py
```diff
@@ -326,20 +321,15 @@ def __parse_metadata_from_fn(self,silent=False):
     def __parse_metadata_from_file(self,fn_meta):
         pass
 
-    def __copy__(self):
```
erikmannerfelt, line 329: I trust that removing the magic method doesn't have a weird side-effect anywhere? I think it is used for the syntax:
```python
r2 = copy(r)
```
which I don't think anybody uses ever, so it won't matter too much!

tests/test_georaster.py
```diff
@@ -33,7 +33,6 @@ def test_init(self):
         # third, rio.Dataset
         ds = rio.open(datasets.get_path("landsat_B4"))
         r3 = gr.Raster(ds)
-        assert r3.filename is None
```
erikmannerfelt, line 36: Does this mean `r3` will have a filename, or why is this assertion removed?

erikmannerfelt (commented):
I approve of the name change! I just have some small comments.

geoutils/georaster.py
rhugonnet: Agree, changed to `downsample`.

geoutils/satimg.py
rhugonnet: All tests are passing, so shouldn't be any issue. 
They were not passing with the `__copy__` method in DEM and SatelliteImage and without one in Raster, it had weird side-effects (infinite recursions).

tests/test_georaster.py
rhugonnet: I know read the filename from a `rio.io.DatasetReader`:
```python
self.filename = filename_or_dataset.files[0]
```

geoutils/satimg.py
erikmannerfelt: Okay. We can just fix that later if it turns out to be a problem.

tests/test_georaster.py
erikmannerfelt: So should the line instead be changed to:
```python
assert r3.filename is not None
```
maybe? 

Sorry this might not be very important. I just didn't understand why the assertion was removed.

adehecq: I agree with Erik here, it's not clear

adehecq:
Looks good!

tests/test_georaster.py
rhugonnet: Added assertion back, it's a detail

rhugonnet:
@adehecq @erikmannerfelt need one approving review!

adehecq (approved):

erikmannerfelt (approved):

Issue #125, opened 2021-03-15
Add a function for points reprojection in projtools


adehecq:
If we don't, we should definitely have a small function for this type of operation in projtools !

_Originally posted by @adehecq in https://github.com/GlacioHack/GeoUtils/pull/120#r594352768_

adehecq:
See e.g. https://github.com/rhugonnet/GeoUtils/blob/a65d7c3ba2546d80e20c9535634d9f6e13d1e774/geoutils/georaster.py#L1287
And missing implementation here:
https://github.com/rhugonnet/GeoUtils/blob/a65d7c3ba2546d80e20c9535634d9f6e13d1e774/geoutils/georaster.py#L1083

PR #129, opened 2021-03-16
Add a function to convert coordinates between two CRS, or from/to latlon

adehecq:
To be used for example by Raster.value_at_coords or self.interp.

I need to add tests, but there is an issue with the accuracy of the results, e.g.:
```
from geoutils import georaster as gr
from geoutils import datasets
from geoutils import projtools as pt
lat, lon = pt.reproject_to_latlon([img.bounds.left, img.bounds.top], img.crs)
x, y = pt.reproject_from_latlon([lat, lon], img.crs)
x - img.bounds.left
```
=> -5.238689482212067e-10
This yields uncertain outputs if for example included in Raster.value_at_coords. If requesting the value of the top-left pixel, it will actually look for the pixel with column -1 instead of 0.
Any idea @rhugonnet ?

### DIFF
--- a/geoutils/georaster.py
+++ b/geoutils/georaster.py
@@ -1054,7 +1054,7 @@ to be cleared due to the setting of GCPs.")
     def value_at_coords(self, x, y, latlon=False, band=None, masked=False,
                         window=None, return_window=False, boundless=True,
                         reducer_function=np.ma.mean):
-        """ Extract the pixel value(s) at the specified coordinates.
+        """ Extract the pixel value(s) at the nearest pixel(s) from the specified coordinates.
 
         Extract pixel value of each band in dataset at the specified
         coordinates. Alternatively, if band is specified, return only that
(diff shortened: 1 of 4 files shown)


atedstone:
A floating point math issue, so there's not going to be a nice fix I don't think.

We could add a rounding option to `reproject_from_latlon`, perhaps with a default value of 2 d.p. (equivalent to cms accuracy). 

```
In [15]: x,y = pt.reproject_from_latlon([lat,lon],img.crs)

In [16]: y
Out[16]: 3089999.9999999995

In [17]: np.round(y,2)
Out[17]: 3090000.0
```

so `pt.reproject_from_latlon` would have kwarg `round=2`.

adehecq:
What do you think about commit a3f3278 @atedstone ?

adehecq:
Solves #125

atedstone (approved):

Issue #34, opened 2020-11-13
Improve correspondence to 'recommended [GitHub] community standards'


atedstone:
E.g. contributors page, issues template...

erikmannerfelt:
I see that this has been mentioned in four PRs now! Is this solved or are there still things to fix?

adehecq:
> I see that this has been mentioned in four PRs now! Is this solved or are there still things to fix?

According to https://github.com/GlacioHack/GeoUtils/community, there are still a few things that do not follow the community standard...

PR #132, opened 2021-03-18
Add howto release

atedstone:
These instructions are based on OGGM (https://github.com/OGGM/oggm/blob/master/HOWTO_RELEASE).

Broadly, part of #34


HOWTO_RELEASE
```diff
@@ -0,0 +1,45 @@
+How to issue a GeoUtils release.
+
+ 1. Go to your local main repository (not the fork) and ensure your master branch is synced:
+       git checkout master
+       git pull
+ 2. Look over whats-new.rst and the docs. Make sure "What's New" is complete
+    (check the date!) and add a brief summary note describing the release at the
+    top.
+ 3. If you have any doubts, run the full test suite one final time!
+      pytest --run-slow --mpl .
```
adehecq, line 10: Can't get this to run on my machine. Does it require a specific pytest version?

HOWTO_RELEASE
```diff
@@ -0,0 +1,45 @@
+How to issue a GeoUtils release.
+
+ 1. Go to your local main repository (not the fork) and ensure your master branch is synced:
```
adehecq, line 3: > not the fork
Does this mean one need to clone the Glaciohack repo?

adehecq:
Made some minor comments, apart from that it seems good!

adehecq (approved):

HOWTO_RELEASE
atedstone: yes.

HOWTO_RELEASE
atedstone: me neither. Can't find out anything useful by googling. @fmaussion , any tips?

fmaussion: `--mpl` is for matplotlib image tests, you probably don't need that.

HOWTO_RELEASE
fmaussion: I recommend to have only one or two persons in charge of doing releases - this needs a bit coordination.

PR #133, opened 2021-03-18
Update raster attributes if downsample is not 1

### LINKED ISSUE #131: dowsample does not update raster attributes
```
from geoutils import georaster as gr
from geoutils import datasets
print(img.width, img.height, img.res)
```
800 655 (30.0, 30.0)

```
img = gr.Raster(datasets.get_path('landsat_B4'), downsample=2)
print(img.width, img.height, img.res)
```
800 655 (30.0, 30.0)

adehecq:
Resolves #131

### DIFF
--- a/geoutils/georaster.py
+++ b/geoutils/georaster.py
@@ -141,22 +141,23 @@ class Raster(object):
         # update attributes when downsample is not 1
         if downsample != 1:
 
-            # withd, height and shape must be same as data
-            self.width = down_width
-            self.height = down_height
-            self.shape = (down_height, down_width)
+            # Original attributes
+            meta = self.ds.meta
+
+            # width and height must be same as data
+            meta.update({'width': down_width,
+                         'height': down_height})
 
             # Resolution is set, transform must be updated accordingly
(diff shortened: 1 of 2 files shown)


atedstone (approved):

atedstone:
Looks good to me. Feel free to merge :)

adehecq:
Actually, some methods relying on rasterio's functionalities like xy2ij will still work incorrectly. I need to check if that can be updated.

rhugonnet:
> Actually, some methods relying on rasterio's functionalities like xy2ij will still work incorrectly. I need to check if that can be updated.

As Erik said, might be worth opening an issue in rasterio on the default behaviour of `.index()`. I'll do that next week.

adehecq:
> > Actually, some methods relying on rasterio's functionalities like xy2ij will still work incorrectly. I need to check if that can be updated.
> 
> As Erik said, might be worth opening an issue in rasterio on the default behaviour of `.index()`. I'll do that next week.

No, it was another issue. Until now I was updating self's attributes, but not self.ds attributes so rio's functionalities where not working correctly. I just fixed that hopefully.

PR #134, opened 2021-03-18
add bands attribute to Raster

### LINKED ISSUE #121: Add info about loaded band in attributes
This is needed whenever extracting some data or plotting.

atedstone:
Resolves #121 

Note that there is also the attribute `self.indexes`. This is always set by the contents of `ds.indexes`. Here, `self.bands` refers uniquely to the band order of `self.data`, only if data has been loaded.

### DIFF
--- a/geoutils/georaster.py
+++ b/geoutils/georaster.py
@@ -271,6 +271,8 @@ class Raster(object):
         """
         Setter method for the _data class member.
 
+        new_data must have the same shape as existing data! (bands dimension included)
+
         :param new_data: New data to assign to this instance of Raster
         :type new_data: np.ndarray
         """
@@ -411,8 +413,11 @@ class Raster(object):
         """
         if bands is None:
             self._data = self.ds.read(masked=self._masked, **kwargs)
+            bands = self.ds.indexes
         else:
             self._data = self.ds.read(bands, masked=self._masked, **kwargs)
+            if type(bands) is int:
+                bands = (bands)
 
         # If ndim is 2, expand to 3
         if self._data.ndim == 2:
@@ -420,6 +425,7 @@ class Raster(object):
 
         self.nbands = self._data.shape[0]
         self.is_loaded = True
+        self.bands = bands
 
     def crop(self, cropGeom, mode='match_pixel'):
         """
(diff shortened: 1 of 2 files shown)


adehecq (approved):
Seems alright!

adehecq:
> Resolves #121
> 
> Note that there is also the attribute `self.indexes`. This is always set by the contents of `ds.indexes`. Here, `self.bands` refers uniquely to the band order of `self.data`, only if data has been loaded.

Or do you think one could update the indexes attribute with _update?

adehecq:
Is it worth updating the count attributes?
```
img = gr.Raster(datasets.get_path('landsat_RGB'), bands=(2,3))
meta = img.ds.meta
meta.update({'count': 2})
img._update(metadata=meta)
print(img)
```
Driver:               GTiff 
Opened from file:     /home/adehecq/development/GlacioHack/GeoUtils/geoutils/datasets/LE71400412000304SGS00_RGB.TIF 
Filename:             /vsimem/2ffec74c-472f-4c0c-b1f0-13a580d962c7/2ffec74c-472f-4c0c-b1f0-13a580d962c7.tif 
Raster modified since disk load?  True 
Size:                 800, 655
Number of bands:      2
Data types:           ('uint8', 'uint8')
Coordinate System:    EPSG:32645
NoData Value:         None
Pixel Size:           30.0, 30.0
Upper Left Corner:    478000.0, 3088490.0
Lower Right Corner:   502000.0, 3108140.0

adehecq:
> Is it worth updating the count attributes?
> 
Upon thinking, probably not. It's good to keep the info about the original file. But maybe we can add a line to info() to print the loaded bands?

atedstone:
not sure there's a right and/or single answer to either of these questions. But...

I think that `bands` should be set in addition to `indexes`. `indexes` should always refer to the ds (i.e. rio DatasetReader) as this is a property of the underlying dataset, whereas `bands` refers specifically to what we've loaded in to memory.

by the same logic, we shouldn't touch `count` either (we've got `nbands` for that).

adehecq:
> not sure there's a right and/or single answer to either of these questions. But...
> 
> I think that `bands` should be set in addition to `indexes`. `indexes` should always refer to the ds (i.e. rio DatasetReader) as this is a property of the underlying dataset, whereas `bands` refers specifically to what we've loaded in to memory.
> 
> by the same logic, we shouldn't touch `count` either (we've got `nbands` for that).

If you use self._update, the attributes will match the self.ds attributes, in particular count and indexes. But not the file on disk, so either way...

PR #144, opened 2021-03-22
Add `copy()` and `query()` methods to Vector and add `test_geovector.py` file

erikmannerfelt:
First off, no tests existed for `geovector.py`!
I created some very simple tests, but these should be improved in the future.

I have added two methods to `Vector`:

* `copy()` just returns a copy of itself. Since Vector is only a wrapper for a gpd.GeoDataFrame right now, this was dead simple.
* `query()` allows for text-based queries to the dataframe. In xdem, this is used to query individual glacier outlines from a larger outlines vector. It would be nice to have this method directly instead of a hacky copy approach done right now.

### DIFF
--- a/geoutils/geovector.py
+++ b/geoutils/geovector.py
@@ -59,6 +59,11 @@ class Vector(object):
 
         return "".join(as_str)
 
+    def copy(self):
+        """Return a copy of the Vector."""
+        # Utilise the copy method of GeoPandas
+        return Vector(self.ds.copy())
+
     def crop2raster(self, rst):
         """
         Update self so that features outside the extent of a raster file are cropped. Reprojection is done on the fly if both data set have different projections.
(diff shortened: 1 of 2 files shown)


erikmannerfelt:
To make the improvement apparent, this is what had to be done previously in my case:
```python
glacier_specific_outline = gu.Vector(all_outlines.ds.copy())
glacier_specific_outline.ds = glacier_specific_outline.ds.query(query_string)
```
but with this addition:
```python
glacier_specific_outline = all_outlines.query(query_string)
```

adehecq (approved):
Looks good to me. Yes, geovector is quite far behind now... Would need to add more functionalities to make it more useful.

rhugonnet (approved):
Looks good! Geopandas is so great, maybe we won't need to add much in Vector :D

Issue #149, opened 2021-04-08
from_array does not take masked arrays into account


adehecq:
**Describe the bug**
if `my_array` is a np.ma.masked_array, calling
`out_raster = gu.Raster.from_array(my_array, transform, crs)`
will not preserve my_array's mask.

**To Reproduce**
```
import geoutils as gu
img = gu.Raster(gu.datasets.get_path('landsat_B4'))
img.set_ndv(255)
print(img.data[0, 0, 0])
```
-> print `masked`

```
out_raster = gu.Raster.from_array(img.data, img.transform, img.crs)
print(out_raster.data[0, 0, 0])
```
-> print 255

**Expected behavior**
The mask of img.data should be preserved and second print statement returns a masked value.

**System (please complete the following information):**
irrelevant

adehecq:
The above example is incomplete, as it would be solved by calling:
`out_raster = gu.Raster.from_array(img.data, img.transform, img.crs, nodata=img.nodata)`

The issue arise when several nodata values exist, for example when differencing two rasters with different no data:
```
import geoutils as gu
img = gu.Raster(gu.datasets.get_path('landsat_B4'))
img.set_ndv(255)
img2 = gu.Raster(gu.datasets.get_path('landsat_B4'))
img2.set_ndv(254)
print(img.data[0, 0, 0])
print(img2.data[0, 0, 592])
```
-> print two masked values
```
diff = img.data - img2.data
out_raster = gu.Raster.from_array(diff, img.transform, img.crs, nodata=img.nodata)
print(out_raster.data[0, 0, 0])
print(out_raster.data[0, 0, 592])
```
print -- and 254.

**Expected behavior**
Second print should return two masked values.

adehecq:
Solved by #150

PR #150, opened 2021-04-08
Make sure from_array takes into account input data mask

adehecq:
See issue #149

### DIFF
--- a/geoutils/georaster.py
+++ b/geoutils/georaster.py
@@ -248,6 +248,10 @@ class Raster(object):
         if len(data.shape) < 3:
             data = np.expand_dims(data, 0)
 
+        # Preserves input mask
+        if isinstance(data, np.ma.masked_array):
+            data.data[data.mask] = nodata
+
         # Open handle to new memory file
         mfh = MemoryFile()
 
--- a/geoutils/georaster.py
+++ b/geoutils/georaster.py
@@ -249,7 +249,7 @@ class Raster(object):
             data = np.expand_dims(data, 0)
 
         # Preserves input mask
-        if isinstance(data, np.ma.masked_array):
+        if isinstance(data, np.ma.masked_array) & (nodata is not None):
             data.data[data.mask] = nodata
 
         # Open handle to new memory file
--- a/geoutils/georaster.py
+++ b/geoutils/georaster.py
@@ -249,8 +249,12 @@ class Raster(object):
             data = np.expand_dims(data, 0)
 
         # Preserves input mask
-        if isinstance(data, np.ma.masked_array) & (nodata is not None):
-            data.data[data.mask] = nodata
+        if isinstance(data, np.ma.masked_array):
+            if nodata is None:
+                if np.sum(data.mask) > 0:
+                    raise ValueError("For masked arrays, a nodata value must be set")
+            else:
+                data.data[data.mask] = nodata
 
         # Open handle to new memory file
         mfh = MemoryFile()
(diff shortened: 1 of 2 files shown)


erikmannerfelt:
Could you add one or multiple tests for this, like you do in the issue? The addition looks good, but it's hard to assess without testing.

geoutils/georaster.py
```diff
@@ -248,6 +248,14 @@ def from_array(cls, data, transform, crs, nodata=None):
         if len(data.shape) < 3:
             data = np.expand_dims(data, 0)
 
+        # Preserves input mask
+        if isinstance(data, np.ma.masked_array):
+            if nodata is None:
+                if np.sum(data.mask) > 0:
```
erikmannerfelt, line 254: I didn't know `np.sum()` works on boolean arrays! Makes me wonder why `np.count_nonzero()` exists, since it basically offers the same functionality..

tests/test_georaster.py
```diff
@@ -504,3 +504,26 @@ def test_value_at_coords(self):
         assert img.value_at_coords(x, y) == \
             img.value_at_coords(lon, lat, latlon=True) == \
             img.data[0, -1, -2]
+
+    def test_from_array(self):
+
+        # Test that from_array works if nothing is changed
+        # -> most tests already performed in test_copy, no need for more
+        img = gr.Raster(datasets.get_path('landsat_B4'))
+        out_img = gr.Raster.from_array(img.data, img.transform, img.crs, nodata=img.nodata)
```
erikmannerfelt, line 513: If I understand `Raster.from_array()` correctly, `data` is not copied, so the below assertion will always be true, because it is the same object. I might be wrong though...

tests/test_georaster.py
```diff
@@ -504,3 +504,26 @@ def test_value_at_coords(self):
...
+        assert np.array_equal(out_img.data, img.data)
```
erikmannerfelt, line 514: Could do `assert out_img == img` ;) But this works too!

erikmannerfelt (approved):
Looks great! I have some comments on the test, but they are quite small, so this PR could be approved as is in my book.

geoutils/georaster.py
adehecq: Yes, count_nonzero and sum will return the same value for boolean, but not for arrays containing other values than 0 and 1!

erikmannerfelt: Ah, thanks for the clarification!

tests/test_georaster.py
adehecq: What do you mean? img and out_img are pointing to different objects. For example, if I do:
```
out_img.data += 5
np.array_equal(img.data, out_img.data)
```
returns False.
Or did I misunderstand your statement?

tests/test_georaster.py
adehecq: Good call!

tests/test_georaster.py
erikmannerfelt: Oh, it must be copying implicitly somewhere, which I think is the preferred way.
Maybe here?:

https://github.com/GlacioHack/GeoUtils/blob/f1641a8182713389aaec8e1c2fdd87ad4c47bccb/geoutils/georaster.py#L265

Then everything is fine!

tests/test_georaster.py
```diff
@@ -511,7 +511,7 @@ def test_from_array(self):
         # -> most tests already performed in test_copy, no need for more
         img = gr.Raster(datasets.get_path('landsat_B4'))
         out_img = gr.Raster.from_array(img.data, img.transform, img.crs, nodata=img.nodata)
-        assert np.array_equal(out_img.data, img.data)
+        assert out_img == img
```
erikmannerfelt, line 514: "The author of this commit finds the most beautifully prepared syntax and uses it in textbook fashion. 5/5 raspberries"  - Mannerfelt Times 2021

PR #152, opened 2021-04-08
Raster.reproject was not taking into account self.nodata if none was …

adehecq:
…given. Fix this


erikmannerfelt (approved):
I think I've had this issue actually! Great that it can be fixed (although I just trust you because you haven't provided a test ;)  )

adehecq:
> I think I've had this issue actually! Great that it can be fixed (although I just trust you because you haven't provided a test ;) )

Just to make you happy, I decided to add a simple test. Then I realized that tests for reproject were really lacking. And by adding some tests, I found a bug!! ^^

erikmannerfelt (commented):
Great!!

That bug's been bugging me for a while, so I am happy to see it fixed!

PR #154, opened 2021-04-12
Improve Vector's rasterize options

adehecq:


adehecq:
Will need some testing, hence issue #155

geoutils/geovector.py
```diff
@@ -158,8 +156,111 @@ def create_mask(self, rst=None, crs=None, xres=None, yres=None, bounds=None, in_
...
+                    "in_value must have same length as self.ds.geometry, currently {} != {}".format(
+                        len(in_value), len(vect.geometry))
+                )
+
+            out_geom = ((geom, value) for geom, value in zip(vect.geometry, in_value))
+
+            mask = features.rasterize(shapes=out_geom, fill=out_value, out_shape=out_shape,
+                                      transform=transform)
+
+        elif isinstance(in_value, (int, float, np.integer)):
```
erikmannerfelt, line 257: What happens here if it is a `np.float64` or `np.float32`?

I propose that this is changed to something alike:

```python
else:  # If it wasn't an iterable
    try:  # Try to use the in_value
        mask = features.rasterize(shapes=vect.geometry, fill=out_value, out_shape=out_shape,
                                  transform=transform, default_value=in_value)
    # If in_value could not be used.
    except ValueError:  # I just presume that this will raise a ValueError. It will have to be checked.
        raise ValueError(
                "in_value must be a single number or an iterable with same length as self.ds.geometry"
        )

**Review by erikmannerfelt — COMMENTED** [review]
This will be a great addition! I see some small room for improvement.

I lack tests for the new methods, however. I am especially interested in the output shape of `create_mask()` and `rasterize()`. Will this work?:
```python
vector = gu.Vector("path/to/vector.shp")
raster = gu.Raster("path/to/raster.tif")

mask = vector.create_mask(raster)

masked_values = raster.data[mask]
``` 
I presume that it won't, since `mask` has the shape `(height, width)` and `raster.data` has the shape `(1, height, width)`, or am I wrong? Would be nice with simple tests to make sure that things work as they are supposed to!

geoutils/geovector.py
adehecq: Regarding the np.float issues, I thought that np.float32 and float64 would be considered of type float too. That's because I kept having the deprecation warning that np.float is an alias for float.
What I was looking for is actually np.floating, that covers all cases (even float128!).

Regarding your second comment, no, weirdly rasterio does not raise an error, it will just return an array filled with the 'fill" value...

erikmannerfelt: Isinstance doesn't work perfectly here:
```python
In [1]: import numpy as np

In [2]: val = np.float32(4.0)

In [3]: isinstance(val, float)
Out[3]: False
```
`np.isscalar()`works though!
```python
In [4]: np.isscalar(val)
Out[4]: True

In [5]: np.isscalar(4)
Out[5]: True

In [6]: np.isscalar(4.0)
Out[6]: True

```

adehecq:
> I lack tests for the new methods, however. I am especially interested in the output shape of `create_mask()` and `rasterize()`. 

Yes, that's why I added issue #155. I don't have time for tests now, as basically the geovector test suite was very limited so far and I needed to include new datasets. You included a URL to a test dataset, but as it does not cover the same area as our test rasters, it is useless for certain tests...


> 
> ```python
> vector = gu.Vector("path/to/vector.shp")
> raster = gu.Raster("path/to/raster.tif")
> 
> mask = vector.create_mask(raster)
> 
> masked_values = raster.data[mask]
> ```
> 
> I presume that it won't, since `mask` has the shape `(height, width)` and `raster.data` has the shape `(1, height, width)`, or am I wrong? Would be nice with simple tests to make sure that things work as they are supposed to!

Very good point! You're right, it won't work. So far, I've always considered the mask to be of dimension 2, and I often had to catch that issue. Although I hate all these arrays of dimension 3 with an empty dimension, it's better to be self consistent here...

erikmannerfelt:
> 
> Very good point! You're right, it won't work. So far, I've always considered the mask to be of dimension 2, and I often had to catch that issue. Although I hate all these arrays of dimension 3 with an empty dimension, it's better to be self consistent here...

I also think the 3 dimensions are really annoying!! Makes sense for RGB+ images, but not DEMs!

erikmannerfelt:
If I was dictator, I would have the shapes (height, width, band) for multiband images, and (height, width) for singleband — just like matplotlib, opencv and Pillow.

Below is an example, showcasing reading an image first as RGB (or BGR in opencv's case...) and then as grayscale:
```python
In [1]: import cv2

In [2]: cv2.imread("picture.jpg").shape
Out[2]: (863, 863, 3)

In [3]: cv2.imread("picture.jpg", cv2.IMREAD_GRAYSCALE).shape
Out[3]: (863, 863)
```

geoutils/geovector.py
adehecq: I checked the documentation of np.isscalar and it is True if val is a string too!
There they recommend using `isinstance(x, numbers.Number)` instead for numeric type.

erikmannerfelt: Ah, good to know!

erikmannerfelt (approved):
Great!

erikmannerfelt (approved):

Issue #156, opened 2021-04-14
`Raster` attributes are not instantiated correctly.


erikmannerfelt:
**Describe the bug**
The attributes, (e.g. `transform`, `crs`) are instantiated in the `__init__()` in a way that pylint and other tools do not understand. In addition, they have no type hints.

**To Reproduce**

If this is saved in `/tmp/script.py`:
```python
"""Sample code that should conform to pylint's standards."""
import geoutils as gu

raster = gu.Raster(gu.datasets.get_path("landsat_B4"))

TRANSFORM = raster.transform
CRS = raster.crs
```
And `pylint` is run:
```
❯ pylint /tmp/script.py
************* Module script
/tmp/script.py:6:12: E1101: Instance of 'Raster' has no 'transform' member (no-member)
/tmp/script.py:7:6: E1101: Instance of 'Raster' has no 'crs' member (no-member)

----------------------------------------------------------------------
Your code has been rated at -15.00/10 (previous run: -17.50/10, +2.50)
```
The `transform` and `crs` attributes seem to be nonexistent, which is completely wrong.

**Expected behavior**
I expect the above script to run without `pylint` errors

I am currently working on a PR to fix this.

PR #157, opened 2021-04-14
Fix type hints and conform to pep-8

erikmannerfelt:
See #156 for why this was needed.

In addition, my `autopep8` went to town and fixed some pep-8-unfrendly statements (mostly function calls without proper spacing, e.g. `func(a=1,b=2)` -> `func(a=1, b=2)`, and better import sorting.)


geoutils/georaster.py
```diff
@@ -85,8 +90,25 @@ class Raster(object):
     # This only gets set if a disk-based file is read in.
     # If the Raster is created with from_array, from_mem etc, this stays as None.
     filename = None
-    _is_modified = None
-    _disk_hash = None
+    _is_modified: Optional[bool] = None
+    _disk_hash: Optional[int] = None
+
+    # Rasterio-inherited names and types are defined here to get proper type hints.
```
adehecq, line 96: Oh, that seems annoying for future maintenance since it is more or less a duplicate of default_attrs...
Isn't there a way to combine both? Could we replace the list by a dictionary containing the type?

adehecq:
Good job on enforcing PEP-8 !

geoutils/georaster.py
erikmannerfelt: It was hard, but I did it! The `default_attrs` is commented out and is instead replaced by type annotations in the class definition, combined with a new `self._get_rio_attrs()`.

`self._get_rio_attrs()` finds all attributes that have been annotated that corresponds to attributes in the DatasetReader (transform, bounds, res, etc.) and sets the attributes accordingly. This means that as long as the annotations are up to date, everything works well.

rhugonnet: I think that this is necessary for pylint. I also had issues with those attributes in PyCharm (not recognized), because those were not initialized but only defined by `setattr(self)`. It should be fine, hopefully we won't add a new attribute every day!

adehecq: Does it mean that now all of rasterio's attributes are copied? Or still the same as before?

erikmannerfelt: It's exactly the same as before. It's just that the common attributes are set in a different place (in the class instead of in the list)

adehecq:
Tests fail because pylint is missing in the dev-requirements.txt

adehecq (approved):

PR #158, opened 2021-04-16
Add `split_bands()` method to Raster

erikmannerfelt:
As discussed with @adehecq, this functionality may have widespread use, and will be especially useful for a function in the making in `xdem`.

## Example:
```python
import geoutils as gu

rgb_img = gu.Raster(datasets.get_path('landsat_RGB'))

red, green, blue = img.split_bands()

# Do all kinds of analyses on each band separately
```
Maybe a `.stack()` method or similar should be added to reverse the functionality of `.split_bands()`?

Unfortunately, I wanted to add a `copy: bool =  False` kwarg to reduce copying, but this is not currently possible. `Raster.from_array()` always returns a copy, which is great in most cases, but maybe something can be done about that in the future? A `copy: bool = True` kwarg for `from_array()` which can be changed in case of memory-sensitive work?

### DIFF
--- a/geoutils/georaster.py
+++ b/geoutils/georaster.py
@@ -1446,10 +1446,10 @@ to be cleared due to the setting of GCPs.")
 
         :returns: A list of Rasters for each band.
         """
-        bands: list[Raster] = []
+        bands: list[self] = []
 
         for band_n in range(self.nbands):
-            bands.append(Raster.from_array(
+            bands.append(self.from_array(
                 self.data[band_n, :, :],
                 transform=self.transform,
                 crs=self.crs,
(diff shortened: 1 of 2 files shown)


adehecq (approved):
Looks good to me! 
Indeed in that case it would make sense to just have a pointer to the previous data, rather than copying...

erikmannerfelt:
> Looks good to me!
> Indeed in that case it would make sense to just have a pointer to the previous data, rather than copying...

I'll add it as an issue!

adehecq:
> > Looks good to me!
> > Indeed in that case it would make sense to just have a pointer to the previous data, rather than copying...
> 
> I'll add it as an issue!

I think by construction it would be really difficult to do, because from_array opens a memory file, with the data saved in it... We would need another way to create a Raster object from an array.

Issue #159, opened 2021-04-16
Allow `Raster.from_array` and `Raster.split_bands()` to work without copying.


erikmannerfelt:
Right now, `Raster.from_array()` always returns a copy of the data that are given. This is most often the wanted behaviour, but there may be cases when this is not sought. The method should optimally have a `copy` attribute which defaults to `True`.

A case where this would be nice is with the `Raster.split_bands()` method (which calls `Raster.from_array()` and is therefore dependent on its copy-ness):
```python
rgb = gu.Raster(datasets.get_path("landsat_RGB"))

red = rgb.split_bands(copy=False)[0]  # copy=False could be the default here.

red.data += 1  # Modify the red channel (it should also modify the rgb image)
assert np.array_equal(red.data, rgb.data[0, :, :])  # This should pass.

```

adehecq:
I believe this will be really difficult by design, because from_array creates a memory file, which need to have the data saved in itself.
This is probably a deeper level issue with rasterio...

PR #160, opened 2021-04-16
Add copy/no-copy functionality to `split_bands()` and `subset` argument

erikmannerfelt:
I should have experimented a little more before my previous PR!

As shown in the test, `split_bands(copy=False)` now works as one would expect it to, and `.split_bands(copy=True, subset=0)` can for example be used to only return the red band (instead of copying every band)!

Solves the example in #159.

### DIFF
--- a/geoutils/georaster.py
+++ b/geoutils/georaster.py
@@ -1444,7 +1444,7 @@ to be cleared due to the setting of GCPs.")
 
         return rpts
 
-    def split_bands(self, copy=False, subset: Optional[Union[list[int], int]] = None):
+    def split_bands(self, copy: bool = False, subset: Optional[Union[list[int], int]] = None):
         """
         Split the bands into separate copied rasters.
 
--- a/geoutils/georaster.py
+++ b/geoutils/georaster.py
@@ -1446,10 +1446,10 @@ to be cleared due to the setting of GCPs.")
 
         :returns: A list of Rasters for each band.
         """
-        bands: list[Raster] = []
+        bands: list[self] = []
 
         for band_n in range(self.nbands):
-            bands.append(Raster.from_array(
+            bands.append(self.from_array(
                 self.data[band_n, :, :],
                 transform=self.transform,
                 crs=self.crs,
--- a/geoutils/georaster.py
+++ b/geoutils/georaster.py
@@ -1,8 +1,12 @@
 """
 geoutils.georaster provides a toolset for working with raster data.
 """
+from __future__ import annotations
+
 import os
 import warnings
+from typing import Optional, Union
+
 import numpy as np
 import collections
 import rasterio as rio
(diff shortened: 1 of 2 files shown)


geoutils/georaster.py
```diff
@@ -1440,20 +1444,43 @@ def interp_points(self,pts,nsize=1,mode='linear',band=1):
 
         return rpts
 
-    def split_bands(self):
+    def split_bands(self, copy=False, subset: Optional[Union[list[int], int]] = None):
```
adehecq, line 1447: What is the idea behing the Optional class? Isn't it obvious by the fact that it has a default value?
Otherwise, why is it not needed for copy, for example?

geoutils/georaster.py
```diff
@@ -1440,20 +1444,43 @@ def interp_points(self,pts,nsize=1,mode='linear',band=1):
...
+                    transform=self.transform,
+                    crs=self.crs,
+                    nodata=self.nodata
+                ))
+        else:
+            for band_n in indices:
+                # Generate a new instance with the same underlying values.
+                raster = Raster(self)
+                # Set the data to a slice of the original array
+                raster._data = self.data[band_n, :, :].reshape(((1,) + self.data.shape[1:]))
```
adehecq, line 1481: What if indices is only a subset of the original bands? 
Let's say e.g. self had 3 bands originally, and you extract only 1 band. Then you will loop only once, but self.data still hasa length of 3, no?

adehecq: Oh no, I think I get it. The output is always a single band raster.
But does this effectively not copy the data??

tests/test_georaster.py
```diff
@@ -560,17 +560,44 @@ def test_split_bands(self):
...
+        # Extract a subset with a list in a weird direction
+        blue2, green2 = img.split_bands(copy=False, subset=[2, 1])
+
+        # Check that the subset functionality works as expected.
+        assert np.array_equal(red.data.astype("float32"), red2.data.astype("float32"))
+        assert np.array_equal(blue.data.astype("float32"), blue2.data.astype("float32"))
+        assert np.array_equal(green.data.astype("float32"), green2.data.astype("float32"))
+
+        # Check that the red channel and the rgb data shares memory
+        assert np.shares_memory(red.data, img.data)
```
adehecq, line 583: I didn't know this function! So I guess that's the answer to my earlier question. If red.data is modified, so is img.data?

tests/test_georaster.py
```diff
@@ -560,17 +560,44 @@ def test_split_bands(self):
...
         # Test that the red band corresponds to the first band of the img
         assert np.array_equal(red.data.squeeze().astype("float32"), img.data[0, :, :].astype("float32"))
 
-        # Modify the red band and make sure it doesn't propagate to the original img (it's a copy)
+        # Modify the red band and make sure it propagates to the original img (it's not a copy)
         red.data += 1
-        assert not np.array_equal(red.data.squeeze().astype("float32"), img.data[0, :, :].astype("float32"))
+        assert np.array_equal(red.data.squeeze().astype("float32"), img.data[0, :, :].astype("float32"))
+
+        # Copy the bands instead of
```
adehecq, line 595: Just a missing part to the comment here

adehecq (approved):
Magic !! So magic that I had to ask some silly questions.
I would have to try it to believe it, but I trust you that it works, and the tests shows it too.

geoutils/georaster.py
erikmannerfelt: There's no copying! Check the tests! It's black magic.

tests/test_georaster.py
erikmannerfelt: Yes

tests/test_georaster.py
erikmannerfelt: Fixed in 7a85dd3

geoutils/georaster.py
adehecq: Ok I better understand now! Type hinting is tough!! Good that we have a pro among us ;-)

Issue #163, opened 2021-04-20
Add a dev-requirement.txt or equivalent


adehecq:
Several dependencies in the environment.yml file are only needed for development/testing (e.g. pytest, pylint...).
Could they be moved to a different requirement file?
Maybe @erikmannerfelt if you feel havin ga look at that?

erikmannerfelt:
Yes!

erikmannerfelt:
Implemented in #164

PR #164, opened 2021-04-20
Separate dev-requirements and improve build workflow

erikmannerfelt:
#163 proposed to only have the non-testing requirements in `environment.yml`. This is implemented with this PR.

Secondly, following the implementation in xdem, I changed the build workflow to use `mamba` instead of `conda` for dependency solving. This is much faster. Below is an example in my fork using mamba (top) and not mamba (bottom):
image

As you can see, builds should now take less than a third of the time it took before. This is because most of the build process was just conda being bad at resolving dependencies!

Finally, I changed the name of the build workflow from `build-geoutils-dev` to `build` in order to make a nice badge, like in xdem:

[build](https://github.com/GlacioHack/xdem/actions/workflows/python-package.yml)

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -1,7 +1,7 @@
 # This workflow will install Python dependencies, run tests and lint with a single version of Python
 # For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
 
-name: build-geoutils-dev
+name: build
 
 on:
   push:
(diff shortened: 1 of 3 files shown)


rhugonnet (approved):
It's the first thing I thought about when I was reading your `xdem` PR on CI and saw you were using mamba when building the environment. "Geoutils would probably go 3x faster, because the downloading time should be smaller than that of xdem so it's conda having trouble"
Bam, solved :D

adehecq:
Fantastic!!

PR #165, opened 2021-04-22
Simplify reproject resampling argument to accept strings

erikmannerfelt:
Something that bugs me about rasterio is that a resampling method cannot be given as a string, but must be given as their enums, which are not even imported by the base rio. This means that cubic spline resampling has to be made with:

```python
import geoutils as gu
import rasterio as rio
import rasterio.warp

img1 = gr.Raster(datasets.get_path("landsat_B4"))
img2 = gr.Raster(datasets.get_path("landsat_B4_crop"))

img1_reproj = img1.reproject(img2, resampling=rio.warp.Resampling.cubic_spline)
```

In this PR, I've added parsing functionality for strings as an alternative to the enums:
```python
import geoutils as gu

img1 = gu.Raster(gu.datasets.get_path("landsat_B4"))
img2 = gu.Raster(gu.datasets.get_path("landsat_B4_crop"))

img1_reproj = img1.reproject(img2, resampling="cubic_spline")
```
The `"cubic_spline"` string is matched with the `rio.warp.Resampling.cubic_spline` enum automatically.

If an incorrect string is given, it returns an error showing every valid string:
```python
>>> img1.reproject(img2, resampling="whatever's on the menu")
ValueError: 'whatever's on the menu' is not a valid rasterio.warp.Resampling method. Valid methods: ['nearest', 'bilinear', 'cubic', 'cubic_spline', 'lanczos', 'average', 'mode', 'gauss', 'max', 'min', 'med', 'q1', 'q3', 'rms']
```

### DIFF
--- a/geoutils/georaster.py
+++ b/geoutils/georaster.py
@@ -36,6 +36,29 @@ else:
 default_attrs = ['bounds', 'count', 'crs', 'dataset_mask', 'driver', 'dtypes', 'height', 'indexes', 'name', 'nodata',
                  'res', 'shape', 'transform', 'width']
 
+
+def _resampling_from_str(resampling: str) -> Resampling:
+    """
+    Match a rio.warp.Resampling enum from a string representation.
+
+    :param resampling: A case-sensitive string matching the resampling enum (e.g. 'cubic_spline')
+    :raises ValueError: If no matching Resampling enum was found.
+    :returns: A rio.warp.Resampling enum that matches the given string.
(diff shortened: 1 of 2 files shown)


adehecq:
Looks good! But maybe this should be a PR to rasterio directly!
Probably the reason they do it this way is that GDAL also works like this, with integers. They must have simply followed the same structure.

adehecq (approved):

PR #179, opened 2021-05-20
Improve documentation

erikmannerfelt:
The documentation had already become quite outdated, so I fixed it, and added testing.

Now, if someone does something that makes code in the documentation fail, that code has to be fixed accordingly. This means that the documentation will always be correct!

Also, I added a chapter on Vectors. More will come on it.

You can see the PR version here:
https://geoutils-erikm.readthedocs.io/en/improve_readthedocs/


adehecq (approved):
The new documentation looks fantastic!

Issue #177, opened 2021-05-20
Improve `Raster.reproject()` performance by changing `rio.warp.reproject` kwargs


erikmannerfelt:
`rio.warp.reproject` has two kwargs that we should expose (and possibly have defaults for): `num_threads` and `warp_mem_limit`. The first is an obvious improvement, and just defaults to 1 right now with no way to change it. The second needs more checking, but since rio is so careful with memory, it might speed up things considerably to allow for more memory to be used.

Also, the `**kwargs` for `Raster.reproject()` is not used and should be removed or be given a use!

https://rasterio.readthedocs.io/en/latest/api/rasterio.warp.html?highlight=reproject#rasterio.warp.reproject

So:
1. Expose `num_threads` as a keyword argument. I think it should default to `os.cpu_count() - 1` or similar. No reason to not use multithreading apart from quite specific use-cases!
2. Experiment with the `warp_mem_limit` kwarg and either just expose it, and/or assign a more aggressive default value.
3. Remove `**kwargs` or give it a use.

erikmannerfelt:
Closed by #181

PR #181, opened 2021-05-21
Enable multithreading for `Raster.reproject()` by default.

erikmannerfelt:
`Raster.reproject()` should now be faster as it uses `rio.warp.reproject()` with multiple cores now. I say "should" because it turns out that the single-threaded operation is really fast as well, and I don't have a big raster to benchmark on!!

Solves #177

### DIFF
--- a/geoutils/georaster.py
+++ b/geoutils/georaster.py
@@ -631,8 +631,8 @@ class Raster(object):
 
     def reproject(self, dst_ref=None, dst_crs=None, dst_size=None,
                   dst_bounds=None, dst_res=None, nodata=None, dtype=None,
-                  resampling=Resampling.nearest, silent=False,
-                  **kwargs):
+                  resampling=Resampling.nearest, silent=False, n_threads=0,
+                  memory_limit=64):
         """ 
         Reproject raster to a specified grid.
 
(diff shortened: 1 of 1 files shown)


geoutils/georaster.py
```diff
@@ -631,8 +631,8 @@ def clip(self):
 
     def reproject(self, dst_ref=None, dst_crs=None, dst_size=None,
                   dst_bounds=None, dst_res=None, nodata=None, dtype=None,
-                  resampling=Resampling.nearest, silent=False,
-                  **kwargs):
+                  resampling=Resampling.nearest, silent=False, n_threads=0,
+                  memory_limit=64):
```
adehecq, line 635: Should we leave it to default (0)?
At first I thought this was in GB, so I was a bit surprised!!

adehecq:
Apart from a minor comment, this is great! I didn't know those rasterio options, but it's good they exist.

adehecq (approved):

geoutils/georaster.py
erikmannerfelt: If it's 0, an appropriate amount will be calculated. If >0, it will use that amount of threads.

adehecq: I meant for the memory_limit, not n_threads. If set to 0, it will use rasterio's default. But not a big deal...

PR #185, opened 2021-05-21
Polygonize a raster

AdrienWehrle:
Polygonize a raster based on a given value (default is 1)


geoutils/georaster.py
```diff
@@ -1661,4 +1664,34 @@ def split_bands(self, copy: bool = False, subset: Optional[Union[list[int], int]
...
+        """
+        Return a GeoDataFrame polygonized from a raster.
+        :param in_value: Value of the raster from which to create geometries (Default is 1)
+        :type in_value: int, float
+        :returns: GeoDataFrame containing the polygonized geometries
+        :rtype: geopandas.geodataframe.GeoDataFrame
+        """
+    
+        # Polygonize raster
+        if isinstance(in_value, Number):
```
erikmannerfelt, line 1679: It would be nice to somehow accept a range. With floating point values, it's often hard to specify the exact value one wants, so 1 should rather be like 0.9-1.1. I that this could be done in multiple ways:

1. Accept a tuple as the in_value, where the tuple is `(min_value, max_value)`
2. Accept a list or array of exact values, where the list is `[value1, value2, value3, ...]`
3. Accept a python and/or numpy slice object: `slice(min_value, max_value)` / `np.s_[min_value, max_value]`.

Perhaps, it would be best to allow all three?

For example, I see use-cases where one wants to polygonize all non-zero/non-nan values with a field representing the pixel value. In this case, number 2 would be great to have, and probably not a too difficult thing to implement.

What do you think, @AdrienWehrle, @adehecq, @atedstone, @rhugonnet?

geoutils/georaster.py
```diff
@@ -1661,4 +1664,34 @@ def split_bands(self, copy: bool = False, subset: Optional[Union[list[int], int]
...
-        return bands if len(bands) > 1 else bands[0]
\ No newline at end of file
+        return bands if len(bands) > 1 else bands[0]
+    
+    def polygonize(self, in_value=1):
+        """
+        Return a GeoDataFrame polygonized from a raster.
+        :param in_value: Value of the raster from which to create geometries (Default is 1)
+        :type in_value: int, float
+        :returns: GeoDataFrame containing the polygonized geometries
```
erikmannerfelt, line 1674: Should it return a GeoDataFrame or a Vector, @AdrienWehrle, @adehecq, @atedstone?

tests/test_georaster.py
```diff
@@ -795,3 +795,9 @@ def test_resampling_str(self):
         img3a = img1.reproject(img2, resampling="q1")
         img3b = img1.reproject(img2, resampling=rio.warp.Resampling.q1)
         assert img3a == img3b
+        
+    def test_polygonize(self):
+        """Test that polygonize doesn't raise errors."""
+        img = gr.Raster(datasets.get_path('landsat_B4'))
+        assert gdf = img.polygonize()
```
erikmannerfelt, line 802: This assertion fails because it's not an assertion; it's an assignment. Also, an instantiated class is not equal to the class definition, so `isinstance` has to be used.
It should rather be:
```python
assert isinstance(img.polygonize(), gdf)
```

erikmannerfelt (changes requested):
This is looking really cool, and will be a great addition to the interplay between raster and vector files!!

I see some more potential input types for the `in_value`, but I think they should be relatively easy to implement. Also, there should be tests for different kinds of `in_value`s.

geoutils/georaster.py
erikmannerfelt: It could be something like:
```python
# If it's a simple numeric value
if isinstance(in_value, Number):
    bool_msk = np.array(self.data == in_value).astype(np.uint8)
# If a tuple, assume that it's a (min_value, max_value) range.
elif isinstance(in_value, tuple):
    bool_msk = ((self.data > in_value[0]) & (self.data < in_value[1])).astype("uint8")
# If any other type of sequence (list, ndarray, etc), assume that they represent values to keep.
elif isinstance(in_value, Sequence):
    bool_msk = np.isin(self.data, in_value).astype("uint8")
```
I experimented a bit with the slicing and I guess it doesn't make sense because slicing is on indices, not on cell values. So maybe my nr 3 should be skipped!

It would also be nice to specify "all" somehow, so that all unique pixel values will be rasterized.

AdrienWehrle: Fully agree with that! And with your suggestion! 
I implemented "all" in an earlier version, but removed it at the end because raster values have to be converted to `uint8` (or equivalent) as required for the polygonization, therefore the `raster_value` column of `gdf` would only shows the integer part of the raster values... But maybe that's ok? We could link back the integers with the raster values, but I'm not sure we want such a thing...!

erikmannerfelt: What do you mean exactly? The rasterio documentation shows support for complexities up to float32 (not float64 it seems)

https://rasterio.readthedocs.io/en/latest/api/rasterio.features.html#rasterio.features.shapes

geoutils/georaster.py
adehecq: I think a Vector would make more sense to be consistent with the geoutils' frame.

geoutils/georaster.py
adehecq: > It would be nice to somehow accept a range. With floating point values, it's often hard to specify the exact value one wants, so 1 should rather be like 0.9-1.1. I think that this could be done in multiple ways:
> 
>     1. Accept a tuple as the in_value, where the tuple is `(min_value, max_value)`
> 
>     2. Accept a list or array of exact values, where the list is `[value1, value2, value3, ...]`
> 
>     3. Accept a python and/or numpy slice object: `slice(min_value, max_value)` / `np.s_[min_value, max_value]`.
> 
> 
> Perhaps, it would be best to allow all three?
> 
> For example, I see use-cases where one wants to polygonize all non-zero/non-nan values with a field representing the pixel value. In this case, number 2 would be great to have, and probably not a too difficult thing to implement.
> 
> What do you think, @AdrienWehrle, @adehecq, @atedstone, @rhugonnet?

A tuple with min/max as an option sounds good to me.

erikmannerfelt:
Hey @AdrienWehrle!

This is a great addition to GeoUtils, and I wonder what your plan is on this PR?

AdrienWehrle:
> Hey @AdrienWehrle!
> 
> This is a great addition to GeoUtils, and I wonder what your plan is on this PR?

Hi @erikmannerfelt ! Sorry for my lack of productivity on that PR lately, I handed and then directly defended my MSc thesis over the past two weeks! I'll add the tuple capability in in_value in the coming days!

erikmannerfelt:
> > Hey @AdrienWehrle!
> > This is a great addition to GeoUtils, and I wonder what your plan is on this PR?
> 
> Hi @erikmannerfelt ! Sorry for my lack of productivity on that PR lately, I handed and then directly defended my MSc thesis over the past two weeks! I'll add the tuple capability in in_value in the coming days!

Hi @AdrienWehrle, no rush at all! I just wanted to make sure there was a plan. Take your time, and congratulations on finishing your thesis!

geoutils/georaster.py
```diff
@@ -1857,3 +1860,68 @@ def to_points(
...
+    
+    def polygonize(self, in_value=1):
+        """
+        Return a GeoDataFrame polygonized from a raster.
+        
+        :param in_value: Value or range of values of the raster from which to create geometries (Default is 1). If 'all',
+          all unique pixel values of the raster are used.
+        :type in_value: int, float, tuple, list, string np.ndarray
+        
+        :returns: GeoDataFrame containing the polygonized geometries.
```
erikmannerfelt, line 1872: Change the `GeoDataFrame` word to `Vector`

erikmannerfelt (commented):
Very nice, @AdrienWehrle!!

The build tests fail because you polygonize value=1, but there's no 1 in the img! I have some recommendations on how to fix the test.

Also, you need to run `pre-commit` to format the code to the standards.

If you only want to run it once, do:
```bash
pre-commit run --all-files
```
If this passes, everything's good. Note that some formatters "fail" if they needed to be run, so run the command twice to see how the code does after formatting. For the future, do:
```bash
pre-commit install
```
to run this command each time a commit is made (to make sure it conforms to the code standards)

I had a look at the pre-commit messages and there are a few small things that need to be added. I made comments for these. If you fix them and then run pre-commit twice, the checks should all pass.

AdrienWehrle:
> Very nice, @AdrienWehrle!!
> 
> The build tests fail because you polygonize value=1, but there's no 1 in the img! I have some recommendations on how to fix the test.
> 
> Also, you need to run `pre-commit` to format the code to the standards.
> 
> If you only want to run it once, do:
> 
> ```shell
> pre-commit run --all-files
> ```
> 
> If this passes, everything's good. Note that some formatters "fail" if they needed to be run, so run the command twice to see how the code does after formatting. For the future, do:
> 
> ```shell
> pre-commit install
> ```
> 
> to run this command each time a commit is made (to make sure it conforms to the code standards)
> 
> I had a look at the pre-commit messages and there are a few small things that need to be added. I made comments for these. If you fix them and then run pre-commit twice, the checks should all pass.

Thank you a lot @erikmannerfelt! Alright, all the hooks passed now!

geoutils/georaster.py
```diff
@@ -1857,3 +1859,60 @@ def to_points(
...
+            vals_for_msk = list(set(self.data.flatten()))
+            bool_msk = np.isin(self.data, vals_for_msk).astype("uint8")
+
+        else:
+
+            raise ValueError("in_value must be a number, a tuple or a sequence")
+
+        results = (
+            {"properties": {"raster_value": v}, "geometry": s}
+            for i, (s, v) in enumerate(shapes(self.data, mask=bool_msk))
```
erikmannerfelt, line 1911: I cloned your branch to try to figure out why the test failed, and it's because this does not include any georeferencing! The outgoing coordinates here are in pixel units, not georeferenced units!

erikmannerfelt: Good thing for tests, huh!? ;)

erikmannerfelt: I assume that it's as simple as adding `transform=self.transform` to the `shapes(...)` call.

Source: https://rasterio.readthedocs.io/en/latest/api/rasterio.features.html#rasterio.features.shapes

geoutils/georaster.py
```diff
@@ -1857,3 +1859,60 @@ def to_points(
...
+        )
+
+        gdf = gpd.GeoDataFrame.from_features(list(results))
+        gdf.insert(0, "New_ID", range(0, 0 + len(gdf)))
+        gdf.set_geometry(col="geometry", inplace=True)
```
erikmannerfelt, line 1916: Could you also set the CRS?

`gdf.crs = self.crs`

tests/test_georaster.py
```diff
@@ -876,6 +877,21 @@ def test_resampling_str(self) -> None:
...
+        img = gr.Raster(datasets.get_path("landsat_B4"))
+
+        value = np.unique(img)[0]
+
+        pixel_area = np.sum(img == value) * img.res[0] * img.res[1]
+
+        polygonized = img.polygonize(value)
+
+        polygon_area = polygonized.ds.area.sum()
+
```
erikmannerfelt, line 891: Add:

```python
assert polygonized.crs == img.crs
```

Coordinate systems are important to keep!

erikmannerfelt (approved):
Great job, and great feature!! Thanks for the addition, and I hope to review more features in the future!

PR #186, opened 2021-05-22
Add CI for windows, macos and linux, for multiple python versions

erikmannerfelt:
Also, some tests are improved to work better on Windows, and `scipy` was missing from the `environment.yml`.

This was a really hard one!!

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -10,28 +10,44 @@ on:
     branches: [ master ]
 
 jobs:
-  build:
+  build-all:
+    name: ${{ matrix.os }}, python ${{ matrix.python-version }}
+    runs-on: ${{ matrix.os }}
+    
+    strategy:
+      matrix:
+        os: ["ubuntu-latest", "macos-latest", "windows-latest"]
+        python-version: ["3.7", "3.8", "3.9"]
 
-    runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: bash -l {0}
 
     steps:
     - uses: actions/checkout@v2
-    - name: Set up Python 3.9
-      uses: actions/setup-python@v2
+    
+    - name: Setup environment
+      uses: conda-incubator/setup-miniconda@v2
       with:
-        python-version: 3.9
-    - name: Install dependencies
+        auto-update-conda: true
+        python-version: ${{ matrix.python-version }}
+        mamba-version: "*"
+        channels: conda-forge
+        channel-priority: strict
+        environment-file: environment.yml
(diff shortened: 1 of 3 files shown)


erikmannerfelt:
As you can see by the time the tests take, this doesn't slow down the checking step because all checks are run concurrently.

erikmannerfelt:
By the way, it looks like one test doesn't pass, but it's because I've replaced the build action, so the GitHub falsely expects a message from the old build as well.

jlandmann:
Hi Erik,
thanks for adding scipy to environment.yml! Since you are working on it, maybe it would make sense to also eliminate the invalid escapes (e.g. [here](https://github.com/GlacioHack/GeoUtils/pull/186/checks?check_run_id=2646660011#step:6:32)). They occur as warnings for all actions. I think the badness level is not too high, but it would remove some lengthy warnings from the logs.

.github/workflows/python-app.yml
```diff
@@ -10,28 +10,44 @@ on:
...
-        $CONDA/bin/pip install -r dev-requirements.txt
-        $CONDA/bin/pip install .
+        mamba install sphinx numpydoc sphinx_rtd_theme sphinx-autodoc-typehints sphinxcontrib-programoutput flake8 pytest matplotlib pylint
+        pip install -e . --no-dependencies
+
     - name: Lint with flake8
       run: |
         # stop the build if there are Python syntax errors or undefined names
-        $CONDA/bin/flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
-        # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
```
adehecq, line 33: The comment why you set to 127 was useful, why remove it?

adehecq:
Well done!!

adehecq (approved):

.github/workflows/python-app.yml
erikmannerfelt: I didn't remove it; it's just called with `flake8` instead of `$CONDA/bin/flake8` to work better with Windows!

erikmannerfelt:
> Hi Erik,
> thanks for adding scipy to environment.yml! Since you are working on it, maybe it would make sense to also eliminate the invalid escapes (e.g. [here](https://github.com/GlacioHack/GeoUtils/pull/186/checks?check_run_id=2646660011#step:6:32)). They occur as warnings for all actions. I think the badness level is not too high, but it would remove some lengthy warnings from the logs.

Good point, @jlandmann! I'll add an issue for it. There are quite many small fixes like that, so they would be good to squash in a PR.

Issue #178, opened 2021-05-20
Add easier creation options for `Raster.save()`


erikmannerfelt:
Many argue that compression should be used in every case today, as read/write algorithms are often as fast or faster than the disk itself. Therefore, `Raster.save()` should make it more user-frendly to specify compression. I would argue that it should also be on by default!

`rio.open` has two kwargs when writing that should be more obvious in geoutils:
1. `compress="deflate"` or any other compression algorithm. Equates to the COMPRESS=DEFLATE co
2. `tiled=False` or `True`. Equates to the TILED=YES / TILED=NO co

I know that these can be given as GDAL-like arguments with `co_opts` (creation option options? ;) ), but for a non-GDAL person, this is very far from obvious.

Finally, I think a compression algorithm should be default, and I propose deflate, as it works okay with every dtype that I know of.

rhugonnet:
Agreed, and `DEFLATE` is really the best performing (compared to `LWZ`, for instance) when it comes to rasters with a lot of no data values (most satellite data).
I haven't tested `PACKBITS` though, that might perform well with nodata values too.

PR #187, opened 2021-05-24
Add `compress` and `tiled` arguments for `Raster.save()` and enable compression by default.

erikmannerfelt:
As mentioned in #178, this will especially help newcomers with little to no GDAL experience. Also, compression should be on by default.

@adehecq  and I discussed what the default compression algorithm should be, and we ended up somewhere between LZW and DEFLATE. @rhugonnet recommended DEFLATE over LZW, so I chose DEFLATE. 

We could easily test different compression algorithms in a future project to find the absolute best default for every use-case, but DEFLATE will definitely do the job for now!

### DIFF
--- a/geoutils/georaster.py
+++ b/geoutils/georaster.py
@@ -1019,7 +1019,8 @@ class Raster(object):
 
         self._update(imgdata=imgdata, metadata=meta)
 
-    def save(self, filename, driver='GTiff', dtype=None, blank_value=None, co_opts={}, metadata={}, gcps=[], gcps_crs=None):
+    def save(self, filename, driver='GTiff', dtype=None, compress="deflate", tiled=False, blank_value=None,
+             co_opts={}, metadata={}, gcps=[], gcps_crs=None):
         """ Write the Raster to a geo-referenced file.
 
         Given a filename to save the Raster to, create a geo-referenced file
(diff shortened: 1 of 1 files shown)


geoutils/georaster.py
```diff
@@ -1073,7 +1079,9 @@ def save(self, filename, driver='GTiff', dtype=None, blank_value=None, co_opts={
                       dtype=save_data.dtype,
                       crs=self.ds.crs,
                       transform=self.ds.transform,
-                      nodata=self.ds.nodata, **co_opts) as dst:
+                      nodata=self.ds.nodata,
+                      compress=compress,
+                      tiled=tiled, **co_opts) as dst:
```
adehecq, line 1084: Would it not be safer to check that compress and tiled are not set within co_opts?
Apparently, setting twice the same argument does not raise any error and simply the last value is used, e.g.:
`img.save('test.tif', co_opts={"compress":"LZW", "compress":"DEFLATE"})`
saves with compression DEFLATE.
Or this could be the normal behavior, if we assume that people would simply try to overwrite `compress` using `co_opts`...

erikmannerfelt: That sounds like perfectly normal behavior to me! What we want is "if a user sets something, any default gets overwritten", which is currently the case, right?

adehecq: Yes, that's mostly fine. It's just weird not to have a warning for duplicated options (it should be fixed in rasterio).
The only case I can see is if the user set both compress and co_opts with conflicting values. Then co_opt will override compress. But that should not happen if the user uses save correctly.
Let's accept it as it is for now!

adehecq (approved):

geoutils/georaster.py
erikmannerfelt: Then submit a PR to python ;) It's in python that the duplication is allowed!:
```python
In [1]: {"compress": "lzw", "compress": "deflate"}
Out[1]: {'compress': 'deflate'}

```

adehecq: Haha, I wasn't aware of that! But the issue is that the same argument can be passed multiple times to rio.open here. Anyway, it is ok as it is.

Issue #174, opened 2021-05-20
`Raster.info(stats=True)` warns about ignoring masked_array mask


erikmannerfelt:
**Describe the bug**
`Raster.info(stats=True)` does not handle masked arrays, and ignores the mask with a warning. This should be fixed!

**To Reproduce**
```python
In [1]: import geoutils as gu

In [2]: img = gu.Raster(gu.datasets.get_path("landsat_B4"))

In [3]: img.info(stats=True)
/home/erik/.local/share/conda/geoutils/lib/python3.9/site-packages/numpy/core/fromnumeric.py:753: UserWarning: Warning: 'partition' will ignore the 'mask' of the MaskedArray.
  a.partition(kth, axis=axis, kind=kind, order=order)
Out[3]: "Driver:               GTiff \nOpened from file:     /home/erik/Projects/ETH/GeoUtils/geoutils/datasets/LE71400412000304SGS00_B4_crop.TIF \nFilename:             /home/erik/Projects/ETH/GeoUtils/geoutils/datasets/LE71400412000304SGS00_B4_crop.TIF \nRaster modified since disk load?  False \nSize:                 800, 655\nNumber of bands:      1\nData types:           ('uint8',)\nCoordinate System:    EPSG:32645\nNoData Value:         None\nPixel Size:           30.0, 30.0\nUpper Left Corner:    478000.0, 3088490.0\nLower Right Corner:   502000.0, 3108140.0\n[MAXIMUM]:          255.00\n[MINIMUM]:          13.00\n[MEDIAN]:           124.00\n[MEAN]:             144.04\n[STD DEV]:          79.44\n"
```

**Expected behavior**
There should not be a UserWarning

**System (please complete the following information):**
 - OS: irrelevant
 - Environment: conda

PR #189, opened 2021-05-24
Fix `Raster.info(stats=True)` warning

erikmannerfelt:
Solves #174.

A test is added to validate that no warnings occur in the future.

### DIFF
--- a/tests/test_georaster.py
+++ b/tests/test_georaster.py
@@ -72,12 +72,24 @@ class TestRaster:
             assert r.__getattribute__(attr) == r.ds.__getattribute__(attr)
 
         # Check summary matches that of RIO
-        assert print(r) == print(r.info())
+        assert str(r) == r.info()
 
         # Check that the stats=True flag doesn't trigger a warning
         with warnings.catch_warnings():
             warnings.simplefilter("error")
-            r.info(stats=True)
+            stats = r.info(stats=True)
+
+        r.data.ravel()[:1000] = 0
+        r.set_ndv(0)
+
+        new_stats = r.info(stats=True)
+
+        for i, line in enumerate(stats.splitlines()):
+            if "MAXIMUM" not in line:
+                continue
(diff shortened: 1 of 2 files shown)


adehecq:
Looks good to me!
While at it, should we add a test to make sure that the returned min/max values are the actual ones, with/without mask?
Something like this:
```
np.ma.min(img.data)
```
-> compare to info's output (should be 13 for that case).

```
img.data[0, 0, 0] = 0
img.set_ndv(0)
np.ma.min(img.data)
```
-> should be unchanged.

erikmannerfelt:
> Looks good to me!
> While at it, should we add a test to make sure that the returned min/max values are the actual ones, with/without mask?
> Something like this:
> 
> ```
> np.ma.min(img.data)
> ```
> 
> -> compare to info's output (should be 13 for that case).
> 
> ```
> img.data[0, 0, 0] = 0
> img.set_ndv(0)
> np.ma.min(img.data)
> ```
> 
> -> should be unchanged.

Added in f1ff021

tests/test_georaster.py
```diff
@@ -72,12 +72,24 @@ def test_info(self):
...
-        assert print(r) == print(r.info())
+        assert str(r) == r.info()
 
         # Check that the stats=True flag doesn't trigger a warning
         with warnings.catch_warnings():
             warnings.simplefilter("error")
-            r.info(stats=True)
+            stats = r.info(stats=True)
+
+        r.data.ravel()[:1000] = 0
```
adehecq, line 82: Thanks for adding it! I would just add a quick comment here to explain what you do here.
Also, I believe the minimum should be checked, rather than maximum (the latter will not change if you set values to 0...).

adehecq (approved):

tests/test_georaster.py
erikmannerfelt: Yes, my bad! Fixed in 10146b2

PR #192, opened 2021-05-26
Add PyPI publishing workflow

atedstone:
Made the .yml action for automatically adding releases to PyPI.

I did some searching, there's no ability to have organisational accounts on PyPI, so once we've actioned the first release to PyPI, creating the project there, I'll add admins as package maintainers.


erikmannerfelt:
Great! We still don't have the `geoutils` name though (but I'm working on it).. Perhaps we could upload it to a temporary name in the meantime? Something like `GeoUtils` or `geoutils-dev`?

Also, could you add a manual dispatch functionality for the workflow? In case the workflow fails, one would otherwise have to make an entirely new release just to get it to run again!
```yml
on:
  release:
         types: [created]
  workflow_dispatch:
         inputs:
            reason:
              description: 'Reason'     
              required: true
              default: 'testing' 
```
You could change the description and default value if you like.

erikmannerfelt:
I haven't tried it, but after looking at the twine documentation, it seems like `-r XYZ` or `--repository XYZ` will change the target path. It can also be set with the `TWINE_REPOSITORY` environment variable.

https://twine.readthedocs.io/en/latest/#twine-upload

atedstone:
thanks. actually, what would the purpose behind uploading with a temporary name? if the idea is for people to start using the package, it would probably be confusing if/when we get our proper name....I'd be in favour of waiting to see what the PyPI team come back with first.

erikmannerfelt:
We now own the `geoutils` pypi name!!! 

This means that this PR is highly relevant now. The README should also be updated. Also, the PyPi project description should be the README somehow. That's the case in xdem, so we can just copy whatever's in that setup.py

atedstone:
@erikmannerfelt , hopefully this is ready to go ...

erikmannerfelt (approved):

PR #196, opened 2021-05-27
Empty PR to solve conflict with my fork

adehecq:
I'm not sure why exactly, but my fork is ahead of GlacioHack's and I need to merge this into the main repository if I want to get rid of the difference.
I believe it's because I made changes to the documentation while it was being re-organized.
If there is a better solution, I'll take it!


erikmannerfelt:
Huh!??

Have you tried:
```bash
git pull upstream main
git push -f 
```
?

adehecq:
Yes, that worked!

PR #197, opened 2021-05-27
Fix several issues with reproject when dst_bounds is set

adehecq:
Now, Raster.reproject should be closer (if not equivalent) to what gdalwarp would do, when dst_bounds is set i.e.:
- if dst_bounds is set but nothing else, the dst_res and shape are modified to fit dst_bounds
- if dst_bounds and dst_res are set, will enforce dst_res and will adjust dst_bounds accordinly (by up to one pixel)

I think it would be tough to get a complete test suite for all the possible cases, but I did my best! 
I compared the outputs with gdalwarp on a few tests cases and it's the same. Ideally we would compare with gdal for a million cases!

P.S: I don't know what I did wrong again, but my previous PR is also included here...

### DIFF
--- a/geoutils/georaster.py
+++ b/geoutils/georaster.py
@@ -788,17 +788,19 @@ class Raster(object):
             # Determine target CRS
             dst_crs = CRS.from_user_input(dst_crs)
 
-        # If dst_ref is None, check other input arguments
-        if dst_size is not None and dst_res is not None:
-            raise ValueError(
-                'dst_size and dst_res both specified. Specify only one.')
-
+        # Set output dtype
         if dtype is None:
-            # CHECK CORRECT IMPLEMENTATION! (rasterio dtypes seems to be on a per-band basis)
+            # Warning: this will not work for multiple bands with different dtypes
             dtype = self.dtypes[0]
 
         if nodata is None:
             nodata = self.nodata
+            # If no data was set, need to set one by default, in case reprojection is done outside original bounds
+            # Otherwise, output nodata will be 99999 by default which will not work as expected for uint8 data.
+            if nodata is None:
+                if not silent:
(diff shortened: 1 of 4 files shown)


erikmannerfelt:
If you want to fix the "other PR" problem, you can do:
```bash
git fetch upstream
git reset upstream/main
git commit -am "Whatever message you had last"
```
It will go back to the original main and then apply your commit on top (I think this is rebasing, but I'm not sure!)

erikmannerfelt:
Otherwise the squashing will take care of the weird history!

erikmannerfelt (approved):
Very nice!!

adehecq:
> If you want to fix the "other PR" problem, you can do:
> 
> ```shell
> git fetch upstream
> git reset upstream/main
> git commit -am "Whatever message you had last"
> ```
> 
> It will go back to the original main and then apply your commit on top (I think this is rebasing, but I'm not sure!)

I tried this and I think I just made it worse... Bah!

erikmannerfelt:
> > If you want to fix the "other PR" problem, you can do:
> > ```shell
> > git fetch upstream
> > git reset upstream/main
> > git commit -am "Whatever message you had last"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It will go back to the original main and then apply your commit on top (I think this is rebasing, but I'm not sure!)
> 
> I tried this and I think I just made it worse... Bah!

Hahaha! Maybe your upstream is out of date? I thought `git fetch` would handle it, but maybe you need to pull
```bash
git checkout main
git pull upstream main
git checkout reproject
git reset upstream/main
git commit -am "Your message"

```

erikmannerfelt:
Oh yeah and then you need to force push after modifying the history:
````bash
git push -f
```

PR #201, opened 2021-06-03
Refactor code to adhere to a number of standards

erikmannerfelt:
This was a big one!! Up until now, the code style in GeoUtils has been largely inconsistent, where each contributor has more or less had their own style. This PR brings an automatic check to validate the code using a bunch of tools, and a refactor of the entire code-base to adhere to these standards.

* The codebase now uses black and isort to format the code automatically
* An exception to black's strict rules is the 120 character limit instead of 88. This applies for all linters as well
* Incorrectly spelled words in comments and text are now shown as lints (valdate -> validate and so on)
* mypy is running in strict mode, with some exceptions. This means that all functions have to be type-annotated. I've needed to add the `# type: ignore` flag to some places where mypy doesn't understand the syntax (but it is still technically valid). In the future, we should try to avoid that flag, and could even add it as an anti-pattern (like e.g. pandas does). If mypy can't understand the syntax, it can most often be expressed in a more fool-proof way.
* Docstring type annotations are now flagged as erroneous, and the function/method definition type hints are suggested instead.
* An instruction of how to use these tools is added in `CONTRIBUTING.md`.

Tools that are used:
* `black`: Formats the code in a strict and systematic way
* `mypy`: Validates and enforces type hints
* `flake8`: Provides various lints
* `check-yaml`: Formats yaml files consistently
* `end-of-file-fixer`: Formats the end of files consistently
* `trailing-whitespace`:  Removes trailing whitespaces
* `codespell`: Fixes common spelling mistakes in documentation
* `absolufy-imports`: Convert relative imports to absolute (e.g. "from . import georaster" -> "from geoutils import georaster")
* `isort`: Sorts imports alphabetically and by origin (builtin, third party, local)
* `pyupgrade`: Automatically formats code to use the most modern syntax (set to fit 3.7+)
* Multiple rst lints
* `python-no-eval`: Avoid the use of `eval()` as it can execute arbitrary code.
* `relint`: Custom regex lints. Used currently to define docstring types as an anti-pattern.

Tools that could be added are docstring checks. I didn't bother to add missing docstrings (there are quite many) in this PR, but we should probably enforce them at some point.

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -1,0 +1,15 @@
+name: Linting and formatting (pre-commit)
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-python@v2
+    - uses: pre-commit/action@v2.0.0
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,11 +6,13 @@ repos:
                 - id: end-of-file-fixer
                   exclude: \.txt$
                 - id: trailing-whitespace  # Remove trailing whitespaces
+                - id: check-merge-conflict
         # Fix common spelling mistakes
         - repo: https://github.com/codespell-project/codespell
           rev: v2.0.0
           hooks:
                 - id: codespell
+                  args: [--ignore-words-list=alos]
                   types_or: [python, rst, markdown]
                   files: ^(geoutils|doc|tests)/
 
@@ -31,17 +33,30 @@ repos:
           rev: 3.9.2
           hooks:
                 - id: flake8
-                  args: [--max-line-length=120]
+                  args: [
+                          --max-line-length=120,
+                          --extend-ignore=PDF001,
+                  ]
                   additional_dependencies:
                           - flake8-comprehensions==3.1.0
                           - flake8-bugbear==21.3.2
                           - pandas-dev-flaker==0.2.0
+                  files: ^(geoutils|tests)
         # Lint the code using mypy
         - repo: https://github.com/pre-commit/mirrors-mypy
           rev: v0.812
           hooks:
                 - id: mypy
-                  args: [--strict, --ignore-missing-imports, --allow-untyped-calls]
+                  args: [
+                        --strict,
+                        --ignore-missing-imports,  # Don't warn about stubs since pre-commit runs in a limited env
+                        --allow-untyped-calls,  # Dynamic function/method calls are okay. Untyped function definitions are not okay.
+                        --show-error-codes,
+                        --no-warn-unused-ignores,  # Ignore 'type: ignore' comments that are not used.
+                        --disable-error-code=attr-defined,  # "Module has no attribute 'XXX'" occurs because of the pre-commit env.
+                        --disable-error-code=name-defined  # "Name 'XXX' is not defined" occurs because of the pre-commit env.
+
+                  ]
                   additional_dependencies: [tokenize-rt==3.2.0]
 
 
@@ -77,16 +92,9 @@ repos:
                 - id: python-no-eval
                 # Enforce the use of type annotations instead of docstring type comments
                 - id: python-use-type-annotations
-        
-        # Remove unneeded #noqa comments (in case the line conforms to standards already)
-        - repo: https://github.com/asottile/yesqa
-          rev: v1.2.3
+
+        # Add custom regex lints (see .relint.yml)
+        - repo: https://github.com/codingjoe/relint
+          rev: 1.2.0
           hooks:
-                - id: yesqa
-                  additional_dependencies:
-                          - flake8==3.9.2
-                          - flake8-comprehensions==3.1.0
-                          - flake8-bugbear==21.3.2
-
-                        
-
+                - id: relint
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,92 @@
+repos:
+        - repo: https://github.com/pre-commit/pre-commit-hooks
+          rev: v4.0.1
+          hooks:
+                - id: check-yaml
+                - id: end-of-file-fixer
+                  exclude: \.txt$
+                - id: trailing-whitespace  # Remove trailing whitespaces
+        # Fix common spelling mistakes
+        - repo: https://github.com/codespell-project/codespell
+          rev: v2.0.0
+          hooks:
+                - id: codespell
+                  types_or: [python, rst, markdown]
+                  files: ^(geoutils|doc|tests)/
+
+        # Replace relative imports (e.g. 'from . import georaster' -> 'from geoutils import georaster')
+        - repo: https://github.com/MarcoGorelli/absolufy-imports
+          rev: v0.3.0
+          hooks:
+                - id: absolufy-imports
+
+        # Format the code aggressively using black
+        - repo: https://github.com/psf/black
+          rev: 20.8b1
+          hooks:
+                  - id: black
+                    args: [--line-length=120]
+        # Lint the code using flake8
+        - repo: https://gitlab.com/pycqa/flake8
+          rev: 3.9.2
+          hooks:
+                - id: flake8
+                  args: [--max-line-length=120]
+                  additional_dependencies:
+                          - flake8-comprehensions==3.1.0
+                          - flake8-bugbear==21.3.2
+                          - pandas-dev-flaker==0.2.0
+        # Lint the code using mypy
+        - repo: https://github.com/pre-commit/mirrors-mypy
+          rev: v0.812
+          hooks:
+                - id: mypy
+                  args: [--strict, --ignore-missing-imports, --allow-untyped-calls]
+                  additional_dependencies: [tokenize-rt==3.2.0]
+
+
+
+        # Sort imports using isort
+        - repo: https://github.com/PyCQA/isort
+          rev: 5.8.0
+          hooks:
+                  - id: isort
+
+        # Automatically upgrade syntax to a minimum version
+        - repo: https://github.com/asottile/pyupgrade
+          rev: v2.18.3
+          hooks:
+                - id: pyupgrade
+                  args: [--py37-plus]
+
+        # Various formattings
+        - repo: https://github.com/pre-commit/pygrep-hooks
+          rev: v1.8.0
+          hooks:
+                # Single backticks should apparently not be used
+                - id: rst-backticks
+                # Check that all directives end with double colon
+                - id: rst-directive-colons
+                  types: [text]
+                  types_or: [python, rst]
+                # Inline code should not touch normal text
+                - id: rst-inline-touching-normal
+                  types: [text]
+                  types_or: [python, rst]
+                # Eval should never be used (can do arbitrary code execution)
+                - id: python-no-eval
+                # Enforce the use of type annotations instead of docstring type comments
+                - id: python-use-type-annotations
+        
+        # Remove unneeded #noqa comments (in case the line conforms to standards already)
+        - repo: https://github.com/asottile/yesqa
+          rev: v1.2.3
+          hooks:
+                - id: yesqa
+                  additional_dependencies:
+                          - flake8==3.9.2
+                          - flake8-comprehensions==3.1.0
+                          - flake8-bugbear==21.3.2
+
+                        
+
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -35,7 +35,7 @@ repos:
                 - id: flake8
                   args: [
                           --max-line-length=120,
-                          --extend-ignore=PDF001,
+                          --extend-ignore=E203,   # flake8 disagrees with black, so this should be ignored.
                   ]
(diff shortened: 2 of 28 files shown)


geoutils/geoviewer.py
```diff
@@ -1,98 +1,170 @@
...
-    parser.add_argument('-dpi', dest='dpi', type=str, default='default', help='int, dpi value to use when saving figure (Default is from rcParams).')
-    parser.add_argument('-nodata', dest='nodata', type=str, default='default', help='float, no data value (Default is read from file metadata).')
-    parser.add_argument('-noresampl', dest='noresampl', default=False, action='store_true', help='True or False, if False then allow dynamic image downscaling, if True, prevent it.')
+    parser = argparse.ArgumentParser(description="Visualisation tool for any image supported by GDAL.")
+
+    # Positional arguments
+    parser.add_argument("filename", type=str, help="str, path to the image")
+
+    # optional arguments
+    parser.add_argument(
```
adehecq, line 30: Why are so many spaces and breaklines added here? It looks a bit awful now.

adehecq:
Wow, great job!! Thanks Erik!
It will definitely make the code more readable and consistent.

I will have a look, but not now. As long as it doesn't break the tests, I think I'm happy!

geoutils/geoviewer.py
erikmannerfelt: `black` is subtitled "The Uncompromising Code Formatter" for a reason ;) The lengths help texts drive the expressions to not fit in one-liners, so `black` divides them accordingly.

geoutils/satimg.py
```diff
@@ -283,56 +306,61 @@ def __init__(self, filename_or_dataset, attrs=None, load_data=True, bands=None,
...
         """
         Attempts to pull metadata (e.g., sensor, date information) from fname, setting sensor, satellite,
         tile, datetime, and date attributes.
         """
-
         fname = self.filename
-        name_attrs = ['satellite', 'sensor', 'product', 'version', 'tile_name', 'datetime']
-        attrs = parse_metadata_from_fn(fname)
+        name_attrs = ["satellite", "sensor", "product", "version", "tile_name", "datetime"]
+        attrs = parse_metadata_from_fn(fname if fname is not None else "")
```
rhugonnet, line 328: wow, was this "if loop" made automatically for the cases with `fname=None`? :O 
would be impressive!

erikmannerfelt: No that was me haha. `parse_metadata_from_fn` raises a `ValueError` if None is given. So I made that fix to pass an empty string instead.

.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,15 @@
+name: Linting and formatting (pre-commit)
+
+on:
+  push:
+    branches: [ main ]
```
adehecq, line 5: Technically, this means that the operations will be run only when pushing to the main branch, so in a PR?
Or will a PR be rejected if it does not conform to the syntax?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,100 @@
...
+          rev: v4.0.1
+          hooks:
+                - id: check-yaml
+                - id: end-of-file-fixer
+                  exclude: \.txt$
+                - id: trailing-whitespace  # Remove trailing whitespaces
+                - id: check-merge-conflict
+        # Fix common spelling mistakes
+        - repo: https://github.com/codespell-project/codespell
+          rev: v2.0.0
```
adehecq, line 12: All those version numbers will possibly have to be updated regularly?

CONTRIBUTING.md
```diff
@@ -63,3 +63,22 @@ It is also recommended to try the tests from the parent directory, to validate t
 cd ../  # Change to the parent directory
 pytest geoutils
 ```
+
+### Formatting and linting
```
adehecq, line 67: Ok, this answers to my first question!

dev-requirements.txt
```diff
@@ -10,3 +10,4 @@ sphinxcontrib_programoutput
 sphinx_autodoc_typehints
 matplotlib
 scipy
+pre-commit
```
adehecq, line 13: Are black and the other linters not needed?

geoutils/__init__.py
```diff
@@ -2,20 +2,18 @@
...
 
-from . import georaster
-from . import geovector
-from . import datasets
-from . import satimg
-
-from .georaster import Raster
-from .geovector import Vector
-from .satimg import SatelliteImage
+from geoutils import datasets, georaster, geovector, satimg  # noqa
```
adehecq, line 5: what does noqa mean?

geoutils/geovector.py
```diff
@@ -68,64 +77,67 @@ def bounds(self) -> rio.coords.BoundingBox:
...
         left, bottom, right, top = rst.bounds
-        x1, y1, x2, y2 = warp.transform_bounds(
-            rst.crs, self.ds.crs, left, bottom, right, top)
+        x1, y1, x2, y2 = warp.transform_bounds(rst.crs, self.ds.crs, left, bottom, right, top)
         self.ds = self.ds.cx[x1:x2, y1:y2]
 
-    def create_mask(self, rst=None, crs=None, xres=None, yres=None, bounds=None):
+    def create_mask(
+        self,
+        rst: str | gu.georaster.RasterType | None = None,
```
adehecq, line 108: Is this format preferred over using Union?

geoutils/geoviewer.py
adehecq: Ok, sometimes I forget that the GitHub interface adds empty lines in the diffs. so there are actually no blank lines.
But I still don't understand why black does not add a line break when the line is close to 120 characters, rather than after each argument...

geoutils/geoviewer.py
```diff
@@ -1,98 +1,170 @@
...
 
 
-def main():
+def main() -> None:
 
     # Parse arguments
     args = getparser()
 
-    ## Load image metadata ##
+    # Load image metadata #
```
adehecq, line 130: I like to use 2 or 3 # to delineate larger code blocks... e.g.:
```
## Read input data set ##

# Read input 1
...

# Read input 2
...

## Process data ##

# Calculation 1
...

# Calculation 2
...
```
What would be the alternative?

geoutils/geoviewer.py
```diff
@@ -1,98 +1,170 @@
...
-    ## Set default parameters ##
+    # Set default parameters #
 
     # vmin
     if args.vmin is not None:
         try:
-            vmin = float(args.vmin)
-        except ValueError:   # Case is not a number
-            perc, _ = args.vmin.split('%')
+            vmin: float | None = float(args.vmin)
```
adehecq, line 161: Will it be needed to define the type of ALL variables??

adehecq:
I have a few questions before we can merge that, but overall it's an impressive effort, thanks !!
My biggest concern is: how much effort will it be for the developers to follow the rules? I guess the linters will fix many of the syntax issues, but some aspects (e.g. type hinting) will need to be understood by the developers, or else they will never manage to merge a PR, right? So basically, I need to understand what will contributors need to do to be able to merge a PR.
I think it's definitely great to have a clear syntax, but I also don't want it to be so complex that it might discourage potential contributors. Not everyone is as up-to-date with coding practices as you are @erikmannerfelt ;-)

.github/workflows/pre-commit.yml
erikmannerfelt: It will run on a PR. Currently, it will say "failed" but it's still possible to merge.

.pre-commit-config.yaml
erikmannerfelt: Yes, I don't really know why the versioning is fixed. I copied the pandas config and made sure all are updated.

dev-requirements.txt
erikmannerfelt: No, `pre-commit` makes its own separate environment to make sure no rogue configs and similar may interfere.

geoutils/__init__.py
erikmannerfelt: I don't know what it stands for, but it's for disabling `flake8` lints. In this case, it warns about unused imports (which is as it should in an `__init__.py` file)

geoutils/geovector.py
erikmannerfelt: It's something that `pyupgrade` did automatically. I do think it's cleaner though, after getting used to it!

geoutils/geovector.py
```diff
@@ -1,65 +1,74 @@
...
 import geopandas as gpd
 import numpy as np
 import rasterio as rio
 from rasterio import features, warp
+from rasterio.crs import CRS
+
+import geoutils as gu
 
+# This is a generic Vector-type (if subclasses are made, this will change appropriately)
+VectorType = TypeVar("VectorType", bound="Vector")
```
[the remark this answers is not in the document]
erikmannerfelt: It's for a `Vector` or any other potential subclass of it. The same (but for now actually relevant) one exists in georaster.py.

geoutils/geoviewer.py
erikmannerfelt: I get it, and it's more or less personal style. I think it's `flake8` that complains about this. It's all about consistent decisions in the end!

geoutils/geoviewer.py
erikmannerfelt: No, it's because `vmin` may be either a float or a None, so when it's first defined, it needs to be typed.

The incorrect way:
```python
vmin = 3.0

vmin = None  # this is wrong because vmin is supposed to be a float
```
The correct way
```python
vmin: float | None = 3.0

vmin = None  # This is now fine because the type is Optional[float]
```

In another case when the type does not change, types are dynamically understood:
```python
float_value = 1.1

float_value  = 1.7
```

erikmannerfelt:
> I have a few questions before we can merge that, but overall it's an impressive effort, thanks !!
> My biggest concern is: how much effort will it be for the developers to follow the rules? I guess the linters will fix many of the syntax issues, but some aspects (e.g. type hinting) will need to be understood by the developers, or else they will never manage to merge a PR, right? So basically, I need to understand what will contributors need to do to be able to merge a PR.
> I think it's definitely great to have a clear syntax, but I also don't want it to be so complex that it might discourage potential contributors. Not everyone is as up-to-date with coding practices as you are @erikmannerfelt ;-)

I get your point, and I understand that there needs to be some limit to strictness! As previously mentioned, this check doesn't necessarily need to pass in order to merge a PR (although I would strongly prefer to discuss this in the future). Also, running `pre-commit run --all-files` formats the code automatically. Then it's all about fixing the very descriptive lints provided by flake8 and mypy. We can of course also help with PR PRs.

.pre-commit-config.yaml
erikmannerfelt: I just had a look and it seems to be a decision in order to make sure the linting is consistent (if upstream is changed, it shouldn't change the lint unless specified)

geoutils/geovector.py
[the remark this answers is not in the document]
adehecq: Oh ok, so the argument bounds="Vector" refers to the parent class?

geoutils/geoviewer.py
adehecq: What do you use if you want to have a different comment "level" then?

geoutils/geoviewer.py
adehecq: Ok, but then we need to specify the type of any variable that may have several types...
Which of the pre-commit requires this?

geoutils/geovector.py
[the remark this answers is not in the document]
erikmannerfelt: Yes

geoutils/geoviewer.py
erikmannerfelt: I don't know! I'm sure there's documentation on that somewhere. One way is to just do:
```python
#
#
# NEW SECTION
# 
#
```
which I think is fine.

geoutils/geoviewer.py
erikmannerfelt: mypy

Imagine having a function that returns `Optional[float]` (read: it may return a NoneType). If a variable is already typed to `float`, we need to handle this potential NoneType and the error is therefore very useful.

adehecq: Yes, but vmin is just a local variable that is never used outside the function... Anyway, even if I don't quite get the point, that is not really the issue here.
The question is, if I write the same code without writing the type of vmin, what will happen? How will I know there's something wrong in the code and how to fix it?

adehecq:
> I get your point, and I understand that there needs to be some limit to strictness! As previously mentioned, this check doesn't necessarily need to pass in order to merge a PR (although I would strongly prefer to discuss this in the future). Also, running `pre-commit run --all-files` formats the code automatically. Then it's all about fixing the very descriptive lints provided by flake8 and mypy. We can of course also help with PR PRs.

Yes, a failure is not really useful if one keeps ignoring it...
Ok, well if you say that error messages are helpful enough for most contributors to be able to fix the issue themselves, then it should be ok. 
I think we should not spend our time trying to fix other people's PR though...
I guess we can give it a try and if it becomes too difficult to enforce, we can always deactivate some of the pre-commit checks, right?

adehecq (approved):

adehecq:
I will merge this now, as I need to fix a bug in geoutils and if I do it in parallel there might be a lot of conflicts.

PR #203, opened 2021-06-08
Fix bug in reproject with dst_ref. 

### LINKED ISSUE #198: Issue with Raster.reproject when using dst_ref
**Describe the bug**
Raster.reproject does not preserve the output resolution with argument dst_ref. 

**To Reproduce**
```
import geoutils as gu
from geoutils import datasets
r = gu.Raster(datasets.get_path("landsat_B4"))
r2 = gu.Raster(datasets.get_path("landsat_B4_crop"))
# Both r and r2 have the same resolution 30m, need to set a different one
r3 = r2.reproject(dst_res=60)
r4 = r.reproject(r3)
assert r4.transform == r3.transform
```

**Expected behavior**
`r4.transform == r3.transform` should be True, more precisely, `r3.res` and `r4.res` should be equal.

Irrelevant

adehecq:
Also improve tests to include two rasters with different resolution, which was the reason for not identifying this bug before.
Fixes #198.

### DIFF
--- a/geoutils/georaster.py
+++ b/geoutils/georaster.py
@@ -875,22 +875,23 @@ class Raster:
         else:
             dst_shape = (self.count, self.height, self.width)
 
-        # If dst_bounds is set, but not dst_res, will enforce dst_bounds
+        # If dst_bounds is set, will enforce dst_bounds
         if dst_bounds is not None:
 
-            # Calculate new raster size which ensures that pixels resolution is as close as possible to original
-            # Raster size is increased by up to one pixel if needed
-            yres, xres = self.res
-            dst_width = int(np.ceil((dst_bounds.right - dst_bounds.left) / xres))
-            dst_height = int(np.ceil(np.abs(dst_bounds.bottom - dst_bounds.top) / yres))
-            dst_size = (dst_width, dst_height)
+            if dst_size is None:
+                # Calculate new raster size which ensures that pixels resolution is as close as possible to original
+                # Raster size is increased by up to one pixel if needed
+                yres, xres = self.res
+                dst_width = int(np.ceil((dst_bounds.right - dst_bounds.left) / xres))
+                dst_height = int(np.ceil(np.abs(dst_bounds.bottom - dst_bounds.top) / yres))
+                dst_size = (dst_width, dst_height)
 
             # Calculate associated transform
-            dst_transform = rio.transform.from_bounds(*dst_bounds, width=dst_width, height=dst_height)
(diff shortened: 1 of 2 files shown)


geoutils/georaster.py
```diff
@@ -124,7 +124,7 @@ class Raster:
     transform: Affine
     crs: CRS
     nodata: int | float | None
-    res: list[float]
+    res: tuple[float]
```
erikmannerfelt, line 127: This should be `tuple[float, float]` as there are always just two items. Great that this is changed!

`tuple[float]` indicates just one item in the tuple. For a tuple with an unspecified length, use an ellipsis: `tuple[float, ...]`

This is why mypy starts complaining: 
```
tests/test_georaster.py:431: error: Tuple index out of range  [misc]
```

tests/test_georaster.py
```diff
@@ -379,11 +379,27 @@ def test_crop(self) -> None:
...
 
-        # Test reprojecting to dst_ref
+        # Reference raster to be used
         r = gr.Raster(datasets.get_path("landsat_B4"))
         r.set_ndv(0)  # to avoid warnings - will be used when reprojecting outside bounds
+
+        # A second raster with different bounds, shape and resolution
         r2 = gr.Raster(datasets.get_path("landsat_B4_crop"))
+        r2 = r2.reproject(dst_res=20)
+        assert r2.res == (20, 20)
```
erikmannerfelt, line 389: Mypy complains here because you compare floats with integers, the best fix is:
```python
assert r2.res == (20., 20.)
```
or if you (understandably) don't like this behaviour:
```python
assert r2.res == (20, 20)  # type: ignore
```

erikmannerfelt (commented):
Great that you're finding these bugs with the most basic methods! Mypy complains because of some tiny things, but otherwise it looks great!

tests/test_georaster.py
adehecq: Thanks for the feedback! I guess another solution is to indicate that res could also be a tuple of int. Would that be
`list[float | int, float | int]` ?

adehecq: The version with `Number` does not seem to work, but `tuple[float | int, float | int]` yes.

erikmannerfelt: `tuple[float | int, float | int]` would accept inconsistent types:
`(5, 5.5)`
`(3.3, 3)`
etc.

If this is fine by us, the type annotation is okay! Otherwise, it would be:
```python
tuple[float, float] | tuple[int, int]
```
I'm fine with either

erikmannerfelt (approved):

PR #208, opened 2021-06-10
Fix bugs in astype and merge astype and set_dtypes

adehecq:


adehecq:
@erikmannerfelt any help with the pre-commit issues welcome!

erikmannerfelt:
One pre-commit message is because I've carelessly copied from pandas!! Could you please remove line 43 in `pre-commit-config.yaml` and try again?:
https://github.com/GlacioHack/GeoUtils/blob/059f6a3025d0d508d6135a6daffeb1c28ffe563b/.pre-commit-config.yaml#L43

The other is a bit more problematic... Since the `astype` now may not return anything, one needs to treat it as such:

```python
changed_dtype = raster.astype("uint8")

if changed_dtype is None:
   raise ValueError
```
This is obviously not nice, as the `inplace` argument is linked to the return type, so the user will (in 99% of cases) know the return type. I was going to check how pandas does it (since we're borrowing the inplace terminology from them), but then I realised; can't we just have it always return a Raster, being either self or the converted one?

```python
raster.astype("uint8", inplace=True)  # returns itself

raster.astype("uint8", inplace=False)  # returns a new object
```
This would solve the type ambiguity, as you always know what it will return.

Thoughs?

erikmannerfelt:
It seems one should use `typing.overload` for this!

```python
from typing import overload, Literal

@overload
def astype(self, dtype: np.dtype | type | str,  inplace: Literal[True]) -> None: ....

@overload
def astype(self. dtype: np.dtype | type | str,  inplace: Literal[False]) -> Raster: ....

def astype(self. dtype: np.dtype | type | str,  inplace: Literal[False]) -> Raster | None:
    <actual implementation>

```
I havent' tested this yet, to be clear. I'll get back on it.

erikmannerfelt:
It works!! I tried a very simple approach, and the `overload` + `Literal` works well:

```python
from __future__ import annotations
from typing import overload, Literal


@overload
def func(noreturn: Literal[True]) -> None: ...

@overload
def func(noreturn: Literal[False]) -> int: ...


def func(noreturn: bool) -> int | None:
    if noreturn:
        return None
    return 1


new_val = func(noreturn=True) + 1  # This should not work as it's "None + 1"

new_val2 = func(noreturn=False)  + 1
```
mypy appropriately complains about `new_val` (but `new_val2` is fine)
```
hello.py:18: error: Unsupported operand types for + ("None" and "int")
Found 1 error in 1 file (checked 1 source file)
```

erikmannerfelt:
The pandas weirdness is fixed in the latest main (19ec8d8) so you just have to merge with upstream!

adehecq:
> The pandas weirdness is fixed in the latest main ([19ec8d8](https://github.com/GlacioHack/GeoUtils/commit/19ec8d8810f2ab2dc59c412cce0d4148cf768b37)) so you just have to merge with upstream!

Ok great thanks, I'll do that!

erikmannerfelt:
It seems to be an error on numpy's end if I'm not mistaken. What if you change the function signature to accept `dtype: ... | Any` in addition to the other types?

adehecq:
> It seems to be an error on numpy's end if I'm not mistaken. What if you change the function signature to accept `dtype: ... | Any` in addition to the other types?

You mean adding ' | Any' to the type list?
It didn't work...

adehecq:
> It seems to be an error on numpy's end if I'm not mistaken. What if you change the function signature to accept `dtype: ... | Any` in addition to the other types?

Weirdly, it recognize the new 'Any' but still suggest to add another one!

> geoutils/georaster.py:422: error: No overload variant of "astype" of "Raster" matches argument type "Any"  [call-overload]
> geoutils/georaster.py:422: note: Possible overload variants:
> geoutils/georaster.py:422: note:     def astype(self, dtype: Union[Any, type, str, Any], inplace: Literal[False]) -> Raster
> geoutils/georaster.py:422: note:     def astype(self, dtype: Union[Any, type, str, Any], inplace: Literal[True]) -> None

adehecq:
I can play a long time with mypy...

> geoutils/georaster.py:422: note:     def astype(self, dtype: Union[Any, type, str, **Any, Any**], inplace: Literal[False]) -> Raster
> geoutils/georaster.py:422: note:     def astype(self, dtype: Union[Any, type, str, **Any, Any**], inplace: Literal[True]) -> None

adehecq:
Or the other way around...

> geoutils/georaster.py:422: note:     def astype(self, dtype: Union[Any, Any, type, str], inplace: Literal[False]) -> Raster
> geoutils/georaster.py:422: note:     def astype(self, dtype: Union[Any, Any, type, str], inplace: Literal[True]) -> None

erikmannerfelt:
That's odd! Can you push the changes that trigger the error so I can see the whole script?

adehecq:
> That's odd! Can you push the changes that trigger the error so I can see the whole script?

Here we go, enjoy!

erikmannerfelt:
I may have narrowed it down a tiny bit. mypy and `Any` doesn't work well together, because `Any` is untyped. Numpy integration is not on by default (but it can quite easily be enabled), which allows for better typing. I don't know why it's not on by default, but that's why `np.dtype` becomes `Any`. It's a bug or feature that `Any` has trouble matching the overload variants that you put in.

My suggestion is to add `# type: ignore` to the three places for now, then try to fix the numpy integration in another PR.

adehecq:
Also the tests failed on Ubuntu Python 3.7 with the error `ImportError: cannot import name 'Literal' from 'typing' (/usr/share/miniconda/envs/geoutils/lib/python3.7/typing.py)`. 
Is that not available in Python 3.7??

erikmannerfelt:
Oh no!!

erikmannerfelt:
How about we skip the whole overload mess for now and just return `self` if `inplace=True`. We can fix these issues in a later PR?

erikmannerfelt:
The literal `Literal` mess should be solved with the latest main! Just merge from upstream and it should be solved.

#218

erikmannerfelt (approved):
Fantastic!

adehecq:
> Fantastic!

Thanks to you! ;-)

PR #209, opened 2021-06-11
Remove raster clip

### LINKED ISSUE #175: `Raster.clip()` method is empty
I don't know what the purpose of this method was supposed to be, but we should either give it one or remove it!

atedstone:
Closes #175 .


atedstone:
Hmm, not sure why other commits have appeared in here. Any ideas?

erikmannerfelt:
I approve of the title!

I think you can do this to fix the history:
```bash
git fetch upstream
git reset upstream/main  # "un-commit" your changes
git diff  #  double check that only your new changes are staged
git commit -m "your previous message"
git push -f

```

PR #214, opened 2021-06-30
Set SatImg as silent by default

rhugonnet:
All in the name.
To improve reading of tests, etc...


erikmannerfelt (approved):
YES!!!

adehecq:
So much goodness in one line change! ;-)

PR #216, opened 2021-07-02
Add `Raster.to_points()` method

erikmannerfelt:
This is one that I've been missing in many projects!

Many tools work with point clouds, and there's not been a great way to convert a `Raster` (or any other raster-type) into a point cloud. This is for example good for coregistration, as one might not want to load the entire raster.

## As a numpy array
A point cloud with x/y coordinate and all bands' values is returned.
```python
>>> img = gu.Raster(gu.datasets.get_path("landsat_rgb"))
>>> points = img.to_points(subset=10)
>>> points.shape  # It has the columns ["x", "y", "red", "green", "blue"]
(10, 5)
>>> points[0, :]
array([4.931650e+05, 3.106325e+06, 2.550000e+02, 2.550000e+02, 2.550000e+02])
``` 

## As a GeoDataFrame
A `GeoDataFrame` may be more convenient for some, allowing e.g. easier plotting, CRS conversions and much more.
```python
>>> points_frame = img.to_points(subset=3, as_frame=True)
>>> points_frame
      b1     b2     b3                        geometry
0  196.0  186.0  225.0  POINT (490855.000 3094145.000)
1  198.0  191.0  229.0  POINT (480505.000 3095075.000)
2  255.0  255.0  255.0  POINT (484525.000 3096245.000)
>>> points_frame.crs
<Projected CRS: EPSG:32645>
Name: WGS 84 / UTM zone 45N
Axis Info [cartesian]:
- [east]: Easting (metre)
- [north]: Northing (metre)
Area of Use:
- undefined
Coordinate Operation:
- name: UTM zone 45N
- method: Transverse Mercator
Datum: World Geodetic System 1984
- Ellipsoid: WGS 84
- Prime Meridian: Greenwich
```

This may be one large step toward https://github.com/GlacioHack/xdem/issues/134.

### DIFF
--- a/geoutils/georaster.py
+++ b/geoutils/georaster.py
@@ -1823,7 +1823,7 @@ to be cleared due to the setting of GCPs."
         rows = (choice / self.width).astype(int)
 
         # Extract the coordinates of the pixels and filter by the chosen pixels.
-        x_coords, y_coords = [np.array(a) for a in self.ij2xy(rows, cols, offset="center")]
+        x_coords, y_coords = (np.array(a) for a in self.ij2xy(rows, cols, offset="center"))
 
         # If the Raster is loaded, pick from the data, otherwise use the disk-sample method from rasterio.
         if self.is_loaded:
--- a/geoutils/georaster.py
+++ b/geoutils/georaster.py
@@ -8,8 +8,9 @@ import os
 import warnings
 from collections import abc
 from numbers import Number
-from typing import IO, Any, Callable, TypeVar
+from typing import IO, Any, Callable, TypeVar, overload
 
+import geopandas as gpd
 import matplotlib
 import matplotlib.pyplot as plt
 import numpy as np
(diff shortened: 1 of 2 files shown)


erikmannerfelt:
Upon further thought, this should probably be called `Raster.to_points()` and should convert the whole raster by default (`subset=1`)

erikmannerfelt:
And it should not generate x/y coordinates for all pixels; only the ones that will actually be used!

erikmannerfelt:
#218  should be merged before this so I can fix my very hacky `Literal` circumvention here!

tests/test_georaster.py
```diff
@@ -876,6 +876,35 @@ def test_resampling_str(self) -> None:
...
+        points = img1.to_points(1)
+
+        # Validate that 25 points were sampled (equating to img1.height * img1.width) with x, y, and band0 values.
+        assert isinstance(points, np.ndarray)
+        assert points.shape == (25, 3)
+        assert np.array_equal(np.asarray(points[:, 0]), np.tile(np.linspace(0.5, 4.5, 5), 5))
+
+        assert img1.to_points(0.2).shape == (5, 3)
+
+        img2 = gu.Raster(datasets.get_path("landsat_RGB"), load_data=False)
```
rhugonnet, line 896: Stopped commenting the tests here! (but it's easily understandable)

geoutils/georaster.py
```diff
@@ -1771,3 +1772,77 @@ def split_bands(self: RasterType, copy: bool = False, subset: list[int] | int |
...
+            subset = int(data_size * subset)
+        else:
+            subset = int(subset)
+        if subset > data_size:
+            raise ValueError(f"Subset cannot exceed the size of the dataset ({subset} vs {data_size})")
+
+        # If the subset is smaller than the max size, take a random subset of indices, otherwise take the whole.
+        choice = np.random.randint(0, data_size - 1, subset) if subset != data_size else np.arange(data_size)
+
+        cols = choice % self.width
```
rhugonnet, line 1822: I generally use `Raster.coords()` and subsample after flattening the array.
Here maybe you used `xy2ij` and `ij2xy` to account for the GDAL AREA or POINT attribute, thus better representing the location of the point data (left corner of pixel, or center of pixel)?! That is quite important actually!

geoutils/georaster.py
```diff
@@ -1771,3 +1772,77 @@ def split_bands(self: RasterType, copy: bool = False, subset: list[int] | int |
...
+            raise ValueError(f"Subset cannot be zero or negative (given value: {subset})")
+        # If the subset is equal to or less than 1, it is assumed to be a fraction.
+        if subset <= 1.0:
+            subset = int(data_size * subset)
+        else:
+            subset = int(subset)
+        if subset > data_size:
+            raise ValueError(f"Subset cannot exceed the size of the dataset ({subset} vs {data_size})")
+
+        # If the subset is smaller than the max size, take a random subset of indices, otherwise take the whole.
```
rhugonnet, line 1819: Here you could use Amaury's subsampling function for consistency?
But we'd need to move it to GeoUtils first. Maybe we can open an issue about this?

rhugonnet (approved):
Neat!

geoutils/georaster.py
erikmannerfelt: I first used `Raster.coords()` but this allocates two arrays for the whole dataset. Potentially, the user may only want 10 or so samples, so allocating two full-sized arrays would be unnecessary to say the least!

I can expose the location argument, but it should definitely default to "center"

geoutils/georaster.py
erikmannerfelt: Opened #219.

geoutils/georaster.py
erikmannerfelt: Added in efe3ed5.

geoutils/georaster.py
adehecq: Good point!

adehecq:
Nice functionality. I guess we could easily re-use that one in xdem!

Issue #217, opened 2021-07-02
Add `Literal` type for python 3.7


erikmannerfelt:
As of python 3.8, the `Literal` type hint exists by default. It is extremely useful, but cannot be used in geoutils directly as it is not supported by python 3.7! Luckily, there's a solution...

## Why it's important
If I have a function that returns a different type depending on a boolean flag, types should be annotated accordingly: 
```python
def func(flag: bool = True) -> int | str:
    if flag:
         return 1
    return "one"
```
The problem without py38+ is that the type will be `int | str` no matter what flag is given.
This will always return an `int` if `flag==True` and `str` if not. The solution is `overload` and `Literal`:
```python
from typing import overload, Literal

@overload
def func(flag: Literal[True])  -> int: ...

@overload
def func(flag: Literal[False])  -> str: ...

 # here goes the function definition
```
## The solution
[typing-extensions](https://pypi.org/project/typing-extensions) , the official backport of new typing functions!
Using conditionals in the pip and conda requirements, we can make sure that it is only installed for python <=3.7.

This has already been a big problem in #208 and #216 , so this should be fixed soon!

PR #218, opened 2021-07-03
Allow `Literal` for python 3.7 and lower

erikmannerfelt:
Motivated in #217.

This is integral for #208  and #216  to work properly!

The solution is to install+import `typing-extensions` for python versions lower than 3.8. This is an official backport of the new typing functions, so it should work as well!

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -8,3 +8,7 @@ dependencies:
   - pyproj
   - rasterio
   - scipy
+  - pip
+  - pip:
+
+    - "typing-extensions; python_version < '3.8'"
--- a/geoutils/georaster.py
+++ b/geoutils/georaster.py
@@ -8,7 +8,7 @@ import os
 import warnings
 from collections import abc
 from numbers import Number
-from typing import IO, Any, Callable, TypeVar
+from typing import IO, Any, Callable, Literal, TypeVar
 
 import matplotlib
 import matplotlib.pyplot as plt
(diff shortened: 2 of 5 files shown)


environment.yml
```diff
@@ -8,3 +8,7 @@ dependencies:
   - pyproj
   - rasterio
   - scipy
+  - pip
+  - pip:
+
+    - "typing-extensions; python_version < '3.8'"
```
rhugonnet, line 14: didn't know this pip syntax existed! neat

rhugonnet (approved):

rhugonnet:
Thanks for the detailed explanation in #217, I wouldn't have gotten everything otherwise!

geoutils/georaster.py
```diff
@@ -39,6 +45,8 @@
 
 RasterType = TypeVar("RasterType", bound="Raster")
 
+TRUE = Literal[True]  # This should be removed soon; it is for testing the import
```
adehecq, line 48: Is it time to remove it now?

adehecq:
Great! Can you clarify what is the meaning of the Literal? I still don't fully understand its use...

geoutils/georaster.py
erikmannerfelt: Yes, it's gone in the latest main!

erikmannerfelt:
> Great! Can you clarify what is the meaning of the Literal? I still don't fully understand its use...
@adehecq I added a minimal example in #217. Does that help?

Basically, `Literal[True]` means `given_argument == True` but for types. Thus, one can make overload signatures for different booleans or other flags such as integer flags and exact strings. I suggested another `Literal` in https://github.com/GlacioHack/GeoUtils/pull/185#discussion_r663717273 which might also provide some insight!

adehecq:
Ok, I think I get it now, thanks!

Issue #220, opened 2021-07-05
`Raster.reproject()` sometimes warns in a weird way


erikmannerfelt:
**Describe the bug**
When a `Raster` is loaded without a nodata value, `Raster.reproject` will almost 100% trigger a warning about it.
This is not fundamentally wrong, but I feel like the warning is kind of in the wrong place..

**To Reproduce**
```python
>>> import geoutils as gu
>>> img = gu.Raster(gu.datasets.get_path("landsat_B4_crop"))
>>> _ = img.reproject(dst_res=20)
/home/erik/Projects/ETH/GeoUtils/geoutils/georaster.py:827: UserWarning: No nodata set, will use 0
  warnings.warn("No nodata set, will use 0")
```

**Expected behavior**
I don't know exactly what the expected behavior would be, but this particular example feels wrong! Some ideas of solutions are:

- Set the nodata when loading the raster. Maybe this will have unforeseen consequences, however?
- Somehow separate these two scenarios: "There is no nodata-value" or "There may or may not be a nodata-value, but it doesn't exist for now".
- Allow no nodata value in `Raster.reproject()` (I think I prefer this one actually)

**System (please complete the following information):**
 - OS: linux
 - Environment: latest dev environment.

Thoughts, @adehecq @atedstone ?

adehecq:
The issue is particularly problematic with Byte data. With this data type, it's difficult to set a no data value for default as most values are used (0 could be a relevant value).
Initially, if no nodata value was set, it was set to None, and rasterio was taking care of setting it to default. According to what I wrote [here](https://github.com/GlacioHack/GeoUtils/blob/12bd607fe8024720ff264545abf0cc5ba0156ea6/geoutils/georaster.py#L850), it is set to 99999 by default, which has unexpected results for Byte dtype. Hence, I decided to set it to 0 and raise a warning, because this may have unexpected consequences.

The nodata value could indeed be an argument of reproject, but the issue will still be the same. If using a Byte image and choosing a default value may have unexpected consequences, so I think a warning is still the safest way.
The ideal solution would be to handle values and mask separately so that the nodata does not take up one value. But I don't know if it's possible to do. Rasterio has another way to handle no data masks that could be investigated more: https://rasterio.readthedocs.io/en/latest/topics/masks.html.
Sorry I don't have an easy solution to provide right now.

erikmannerfelt:
What would be the consequences of simply allowing `None` as nodata? That would more or less solve the problem, right?

[...]

PR #222, opened 2021-07-05
Resolve test warnings and make CI output cleaner

erikmannerfelt:
This PR includes three minor fixes:

- 8c124a6: When running CI, the output is currently littered with 49 PASSED lines (showing which ones that passed). This is quite unnecessary, as the only interesting information is if they failed or gave warnings! This commit removes the PASSED messages.
- bd73829: @rhugonnet your opinion might be valuable here. Currently, if a `SatelliteImage` is read from disk, `fn_meta` will be `None`, meaning it will try to run `SatelliteImage.__parse_metadata_from_file()`, triggering the "not implemented" warning. In this commit, I've added a check so that the function will not be run if `fn_meta` is None. This consequently removes a ton of test warnings.
- 78dfcd8: This heavily relates to #220. The nodata value has to be set, or else `Raster.reproject()` complains. I've now set nodatas so that this warning disappears.

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -50,4 +50,4 @@ jobs:
 
     - name: Test with pytest
       run: |
-        pytest -rA
+        pytest -ra
--- a/geoutils/satimg.py
+++ b/geoutils/satimg.py
@@ -297,7 +297,7 @@ class SatelliteImage(Raster):
         self.version = version
 
         # trying to get metadata from separate metadata file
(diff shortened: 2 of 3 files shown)


rhugonnet (approved):
BIG YES for the tests and the not-implemented SatImg functionality :)

erikmannerfelt:
> You need the warning filter even with the `set_ndv` fix?

Nope, I just wanted it to validate that there are no warnings anymore. There shouldn't be warnings, so I left it there!

PR #223, opened 2021-07-06
Improve nodata handling in Raster class.

adehecq:
An attempt to solve the issue brought up in #220.


adehecq:
With this fix, the default nodata value set by rasterio/gdal might be a value that is actually being used (e.g. 255 for Byte images).
There should be a warning in such cases, but it's not implemented as I couldn't find a way to know what is the default nodata value set by gdal.

adehecq:
Oh noooo. Currently, if nodata is set to None, i.e. using rasterio's default value, it will not be set properly in the output Raster at line https://github.com/adehecq/GeoUtils/blob/fe29706c9cd7cc985d28bc38285b0d7976e49367/geoutils/georaster.py#L989

geoutils/georaster.py
```diff
@@ -941,15 +953,15 @@ def reproject(
...
         ):
-            if nodata == self.nodata:
+            if (dst_nodata == self.nodata) or (dst_nodata is None):
                 if not silent:
                     warnings.warn("Output projection, bounds and size are identical -> return self (not a copy!)")
                 return self
 
-            else:
+            elif dst_nodata is not None:
                 warnings.warn("Only nodata is different, running self.set_ndv instead")
```
erikmannerfelt, line 962: I know it was not introduced in this particular PR, but this kind of reads like it will return self, but will in fact return a copy. Could this be reformulated appropriately?

Gets me thinking, there should be a method called `Raster.mirror(other_raster)` or something similar that will reprpject (if needed), nodata, resolution, bounds, dtype(?) etc. without warning about if it's actually needed or not! Should I add this as an issue maybe? Basically a function that does: "make sure that these two rasters can now be directly compared".

tests/test_georaster.py
```diff
@@ -616,6 +616,10 @@ def test_set_ndv(self) -> None:
         # Check that the number of no data value is correct
         assert np.count_nonzero(ndv_index.data) == 112088
 
+        # Check that nodata can also be set upon loading
+        r = gr.Raster(datasets.get_path("landsat_B4"), nodata=0)
```
erikmannerfelt, line 620: Is there not somewhere where the nodata will be automatically set to 0? Perhaps best to set it to something else, like 1, and then test it?

erikmannerfelt (commented):
Nice! Just some small comments

tests/test_georaster.py
erikmannerfelt: Also just curious, what happens if a float nodataval is set on an int raster?

adehecq:
> Oh noooo. Currently, if nodata is set to None, i.e. using rasterio's default value, it will not be set properly in the output Raster at line https://github.com/adehecq/GeoUtils/blob/fe29706c9cd7cc985d28bc38285b0d7976e49367/geoutils/georaster.py#L989

Minimum working example:
```
# The image is converted to float only because the default nodata will be > 255. This works also with uint8, but can only be checked visually
import geoutils as gu
img = img.astype('float32')
print(np.max(img.data))
```
-> 255
```
dst = img.reproject(dst_crs='EPSG:4326')
print(np.max(dst.data))
```
-> 1e20
The pixels in areas not covered by the input data are set to the default nodata value 1e20 for Float32, but dst.nodata is not properly set, as the a priori nodata value is unknown.

tests/test_georaster.py
adehecq: Good point for the nodata value. Solved in 730c86d.

adehecq: Regarding your second question, apparently, it accepts it, but nothing happens as no value is equal to the nodata. 
I would say this is a reasonable behavior... what do you think?

geoutils/georaster.py
adehecq: Good point... What do you think, should it rather return self? Or leave it as it is but rephrase the warning?

adehecq: Regarding the mirror method, I feel like this overlaps almost exactly with reproject, except for the nodata value. But is there any reason to absolutely want the two rasters to have the same nodata value?
In any case, you would get what you want with just:
`raster.reproject(dst_ref=raster_ref, dst_nodata=raster_ref.nodata)`
which seems simple enough for not having a dedicated function for it.

adehecq:
> > Oh noooo. Currently, if nodata is set to None, i.e. using rasterio's default value, it will not be set properly in the output Raster at line https://github.com/adehecq/GeoUtils/blob/fe29706c9cd7cc985d28bc38285b0d7976e49367/geoutils/georaster.py#L989
> 
> Minimum working example:
> 
> ```
> # The image is converted to float only because the default nodata will be > 255. This works also with uint8, but can only be checked visually
> import geoutils as gu
> img = img.astype('float32')
> print(np.max(img.data))
> ```
> 
> -> 255
> 
> ```
> dst = img.reproject(dst_crs='EPSG:4326')
> print(np.max(dst.data))
> ```
> 
> -> 1e20
> The pixels in areas not covered by the input data are set to the default nodata value 1e20 for Float32, but dst.nodata is not properly set, as the a priori nodata value is unknown.

Fixed in 8edfc88 and tested in 2c00b4d.

geoutils/georaster.py
```diff
@@ -844,22 +880,25 @@ def reproject(
...
+        # Set source nodata if provided
+        if src_nodata is None:
+            src_nodata = self.nodata
+
+        # Set destination nodata if provided. This is needed in areas not covered by the input data.
+        # If None, will use GeoUtils' default, as rasterio's default is unknown, hence cannot be handled properly.
+        if dst_nodata is None:
+            dst_nodata = self.nodata
+            if dst_nodata is None:
+                dst_nodata = default_ndv(dtype)
```
erikmannerfelt, line 892: Is dtype really always a string here? Can't it be a `np.dtype` as well? I'm asking because your `default_ndv` function seems to require a `str`.

geoutils/georaster.py
```diff
@@ -69,6 +69,31 @@ def _resampling_from_str(resampling: str) -> Resampling:
     return resampling_method
 
 
+# Function to set the default nodata values for any given dtype
+# Similar to GDAL for int types, but without absurdly long nodata values for floats.
+# For unsigned types, the maximum value is chosen (with a max of 99999).
+# For signed types, the minimum value is chosen (with a min of -99999).
+def default_ndv(dtype: str) -> int:
```
erikmannerfelt, line 76: Could this be hidden from the public interface, aka: `_default_ndv()`?

geoutils/georaster.py
erikmannerfelt: I would do something like:

```python
"Only nodata is different. Consider using the 'set_ndv()' method instead."
```
if this warning is even necessary (I don't personally).

Also, I think this warning should respect the `silent` flag.

erikmannerfelt: Regarding the mirror, nevermind. Because as you say, there's no use for synchronizing all properties that I know of!

geoutils/georaster.py
erikmannerfelt: Or it won't make much of a difference if `Raster` is moved, come to think of it.

tests/test_georaster.py
erikmannerfelt: What happens for an integer raster when given `1.0` as a nodata value? Does it correctly register as `1`? I think it may be worth warning if a `float` ndv is given to an `int` raster.

geoutils/georaster.py
adehecq: Mmmh good point. The argument of reproject is flagged as a np.dtype. Do you know an easy way to convert from one to another?

geoutils/georaster.py
```diff
@@ -69,6 +69,31 @@ def _resampling_from_str(resampling: str) -> Resampling:
...
+    Set the default nodata value for any given dtype, when this is not provided.
+    """
+    default_ndv_lookup = {
+        "uint8": 255,
+        "int8": -128,
+        "uint16": 65535,
+        "int16": -32768,
+        "uint32": 99999,
+        "int32": -99999,
+        "float32": -99999,
```
erikmannerfelt, line 87: Isn't the norm `-9999`? Not saying it's better, I'm just considering consistency.

geoutils/georaster.py
adehecq: Hidden in ae955e5 anyway.

tests/test_georaster.py
adehecq: It works fine if one sets a float instead of an int:
```
import geoutils as gu
print(np.min(img.data))
img.set_ndv(13.0)
print(np.min(img.data))
```
yields 13 then 14.
It seems like the value is rounded up when compared to an integer.
`img.set_ndv(12.9999); np.min(img.data)` -> 13
`img.set_ndv(13.9999); np.min(img.data)` -> 14

adehecq:
Additional bug found:
```
img = gu.Raster(gu.datasets.get_path("landsat_RGB"))
img.set_ndv([2,3, 4])
```
fails with error

> ValueError: Given nodata value, [2, 3, 4], is beyond the valid range of its data type, uint8.

adehecq:
> Additional bug found:
> 
> ```
> img = gu.Raster(gu.datasets.get_path("landsat_RGB"))
> img.set_ndv([2,3, 4])
> ```
> 
> fails with error
> 
> > ValueError: Given nodata value, [2, 3, 4], is beyond the valid range of its data type, uint8.

Apparently, this is more of a rasterio issue, as this appears during the rasterio.open call. rasterio.open does not accept several nodata values for multiband rasters.
I've raised an issue on rasterio's project: https://github.com/mapbox/rasterio/issues/2234

tests/test_georaster.py
adehecq: Now an issue is raised when the nodata value is incompatible with self.dtypes. See f366f03.

geoutils/georaster.py
adehecq: Fixed in ae955e5

geoutils/georaster.py
adehecq: It was just weird to have a nodata value that was smaller (in absolute) than the nodata value for int16 for example. And I've seen this value used when I was searching for default GDAL nodata values, but I can't find out where anymore.
Anyway, we could change that easily if needed.

adehecq:
The pre-commit issue is now fixed. If you're happy with all fixes, can you please approve the PR @erikmannerfelt ?

rhugonnet (approved):
Well I approve, impressively thorough :+1:

erikmannerfelt (approved):
Great!

Issue #207, opened 2021-06-10
Some functions require numpy > 1.19.0


adehecq:
**Describe the bug**
The comparison between two raster, using the overload `__eq__` does not work with numpy < 1.19.

**To Reproduce**
Use numpy < 1.19, then
```
import geoutils.georaster as gr
from geoutils import datasets
r = gr.Raster(datasets.get_path("landsat_B4"))
r == r
```
raises the error:

> TypeError: _array_equal_dispatcher() got an unexpected keyword argument 'equal_nan'

**Expected behavior**
This should not raise an error.
To avoid this, either find an alternative to 'equal_nan' or specify the numpy version needed in the requirements.

**System (please complete the following information):**
numpy version '1.18.1'

erikmannerfelt:
Nice that you found out! I'm sure there's a way around it. Worst case, we can just go back to `np.nansum(np.abs(r - r2)) < *dtype_min*`

erikmannerfelt:
Actually, it seems it's just possible to compare with zero:
```python
>>> np.float32(0.11111111111) - np.float32(0.111111111111111111111111111111)
0.0
```

So it should be as simple as `np.nansum(np.abs(r - r2)) == 0.0`. I wonder what happens with masks though...

Issue #224, opened 2021-07-07
Add numpy plugin for mypy


erikmannerfelt:
I'm sure we've all faced some related issue by now, and I think it's a pretty easy fix.

Without the numpy plugin, all `np.ndarray`s and friends evaluate to `Any` in `mypy`s eyes. Enabling the mypy numpy plugin and then fixing some type annotations should fix this.

erikmannerfelt:
After tons of research, I've found that this only works properly with numpy>=1.20... This conflicts with the aim of #207  which is to keep backward compatibility with numpy!

What are your thoughts on backward compatibility, @rhugonnet @adehecq @atedstone.

For context, numpy 1.20 was released on 30 January 2021

adehecq:
I think allowing only numpy > 1.20 might be quite limiting and could conflict with other libraries. So I would be in favor of keeping backward compatibility, especially since this is needed *only* for mypy and not for a core functionality.
Would your ArrayLike type in #227 solve the issue?

erikmannerfelt:
Solutions like the `ArrayLike` in #227 would more or less solve the issue, yes!

erikmannerfelt:
I now added the `_typing.py` module in #227, so I'll close this, as it is more or less not necessary anymore!

In maybe two years, we could reopen the issue to consider stricter numpy types...

PR #227, opened 2021-07-07
Enable doctests with pytest and add `array_equal` function for np<=1.18 compatibility

erikmannerfelt:
Solves #207.

Solves https://github.com/GlacioHack/xdem/issues/171 but for GeoUtils.

Numpy's `array_equal()` got the `equal_nan` keyword in 1.19. This means that before 1.19, all tests fail! This new function works for 1.18 (and probably before), and has an additional `tolerance` argument that may be useful in the future.

In addition, I enabled doctests to run with pytest. I had to disable (not remove, just skip) a few doctests that were not mine as they didn't work!


adehecq:
Looks good to me!
If array_equal is only meant for testing, there is no need to check for performance, compared to numpy's array_equal. But we should make sure it performs well computationally/memory-wise if we intend to use it in the code directly, no?

adehecq:
> In addition, I enabled doctests to run with pytest. I had to disable (not remove, just skip) a few doctests that were not mine as they didn't work!

Should we try to fix the doctests rather than skipping them?

erikmannerfelt:
I checked the performance between numpy and this function (basically C vs python...):
```python
import geoutils as gu
import geoutils.misc
import time
import numpy as np

start_time = time.time()
for _ in range(100):
    np.array_equal(img.data, img.data.copy())
duration = (time.time() - start_time) / 100

start_time = time.time()
for _ in range(100):
    geoutils.misc.array_equal(img.data, img.data.copy())
duration2 = (time.time() - start_time) / 100


print(f"Numpy: {duration:.5f}s/it, geoutils: {duration2:.5f}s/it")
```
prints:
```
Numpy: 0.00008s/it, geoutils: 0.01123s/it
```

So it's definitely slower (but still pretty fast). Keep in mind, however, that `np.array_equal` doesn't respect masks, for example, so it's not a perfect comparison. This new function does, (and needs to call `masked_array.filled()` which probably increases the time consumed) Technically, `array_equal(masked_array, masked_array)` has been wrong from the start!

erikmannerfelt:
I fixed one doctest, @adehecq.

The other two are:

-  The example datasets, which will change depending on the machine, so this can't be doctested cleanly.
- `Raster.value_at_coords()` whose example describes the output in a nice sketch instead of just numbers, which I think is great (so it's an example, just not a "doctestable" example)

erikmannerfelt:
As you brought up in #224, @adehecq, the `ArrayLike` type shortcut and others like it could replace the `numpy.typing` module (making all of this np backwards compatible). However..: `geoutils.misc` cannot currently be imported into `geoutils.georaster` (to use the `ArrayLike` type) because `geoutils.misc.array_equal` needs `RasterType` from `geoutils.georaster` in its signature! Thus, it would be a circular dependency...

May I suggest to add a `_typing.py` or `typing.py` module (I prefer the "hidden" one) where all of these type hints like `ArrayLike` and `DtypeLike` can reside? What do you think about that, @adehecq @atedstone @rhugonnet ?

adehecq:
> I checked the performance between numpy and this function (basically C vs python...):
> 
> ```python
> import geoutils as gu
> import geoutils.misc
> import time
> import numpy as np
> 
> 
> start_time = time.time()
> for _ in range(100):
>     np.array_equal(img.data, img.data.copy())
> duration = (time.time() - start_time) / 100
> 
> start_time = time.time()
> for _ in range(100):
>     geoutils.misc.array_equal(img.data, img.data.copy())
> duration2 = (time.time() - start_time) / 100
> 
> 
> print(f"Numpy: {duration:.5f}s/it, geoutils: {duration2:.5f}s/it")
> ```
> 
> prints:
> 
> ```
> Numpy: 0.00008s/it, geoutils: 0.01123s/it
> ```
> 
> So it's definitely slower (but still pretty fast). Keep in mind, however, that `np.array_equal` doesn't respect masks, for example, so it's not a perfect comparison. This new function does, (and needs to call `masked_array.filled()` which probably increases the time consumed) Technically, `array_equal(masked_array, masked_array)` has been wrong from the start!

Ok, thanks for looking into this! The sample data is relatively small, so I don't know how it scales up for larger raster (also memorywise). But as you say, it has all the features we need to exactly compare two masked arrays, so let's keep that.

adehecq:
> I fixed one doctest, @adehecq.
> 
> The other two are:
> 
>     * The example datasets, which will change depending on the machine, so this can't be doctested cleanly.
> 
>     * `Raster.value_at_coords()` whose example describes the output in a nice sketch instead of just numbers, which I think is great (so it's an example, just not a "doctestable" example)

Great !

adehecq:
> As you brought up in #224, @adehecq, the `ArrayLike` type shortcut and others like it could replace the `numpy.typing` module (making all of this np backwards compatible). However..: `geoutils.misc` cannot currently be imported into `geoutils.georaster` (to use the `ArrayLike` type) because `geoutils.misc.array_equal` needs `RasterType` from `geoutils.georaster` in its signature! Thus, it would be a circular dependency...
> 
> May I suggest to add a `_typing.py` or `typing.py` module (I prefer the "hidden" one) where all of these type hints like `ArrayLike` and `DtypeLike` can reside? What do you think about that, @adehecq @atedstone @rhugonnet ?

If that solves the type issues, then why not. But I agree it should be a hidden module.
Also should we have a document, either directly in _typing.py, or on a Wiki page, where we list common types and how to use them? I find it difficult to handle the type issues sometimes and it would be useful for possible future contributors too. I think it would also help setting the type more consistently throughout the code (e.g. like with the Iterable vs Sequence case).

erikmannerfelt:
> > As you brought up in #224, @adehecq, the `ArrayLike` type shortcut and others like it could replace the `numpy.typing` module (making all of this np backwards compatible). However..: `geoutils.misc` cannot currently be imported into `geoutils.georaster` (to use the `ArrayLike` type) because `geoutils.misc.array_equal` needs `RasterType` from `geoutils.georaster` in its signature! Thus, it would be a circular dependency...
> > May I suggest to add a `_typing.py` or `typing.py` module (I prefer the "hidden" one) where all of these type hints like `ArrayLike` and `DtypeLike` can reside? What do you think about that, @adehecq @atedstone @rhugonnet ?
> 
> If that solves the type issues, then why not. But I agree it should be a hidden module.
> Also should we have a document, either directly in _typing.py, or on a Wiki page, where we list common types and how to use them? I find it difficult to handle the type issues sometimes and it would be useful for possible future contributors too. I think it would also help setting the type more consistently throughout the code (e.g. like with the Iterable vs Sequence case).

We could have a documentation section "For developers" or something!

erikmannerfelt:
Speaking of our `Number` + `mypy` issues, @adehecq, I've now added an `AnyNumber` alias in `geoutils._typing.py`!

erikmannerfelt:
I also added a warning wrapper and associated test that literally removes around 2000 warnings!! There's a deprecation warning in GeoPandas which I can't seem to reproduce outside of GeoUtils, so now I just added a warning catcher to `Vector`.

For reference, here's a warning summary showing 2436 warnings in a previous passed CI run:
```
=============================== warnings summary ===============================
../../../../../usr/share/miniconda/envs/geoutils/lib/python3.7/site-packages/geopandas/_vectorized.py:142: 1084 warnings
tests/test_datasets.py: 86 warnings
tests/test_georaster.py: 86 warnings
tests/test_geovector.py: 1091 warnings
tests/test_projtools.py: 86 warnings
  /usr/share/miniconda/envs/geoutils/lib/python3.7/site-packages/geopandas/_vectorized.py:142: DeprecationWarning: An exception was ignored while fetching the attribute `__array_interface__` from an object of type 'Polygon'.  With the exception of `AttributeError` NumPy will always raise this exception in the future.  Raise this deprecation warning to see the original exception. (Warning added NumPy 1.21)
    aout[:] = out

../../../../../usr/share/miniconda/envs/geoutils/lib/python3.7/site-packages/pandas/core/dtypes/cast.py:1981
../../../../../usr/share/miniconda/envs/geoutils/lib/python3.7/site-packages/pandas/core/dtypes/cast.py:1981
  /usr/share/miniconda/envs/geoutils/lib/python3.7/site-packages/pandas/core/dtypes/cast.py:1981: DeprecationWarning: An exception was ignored while fetching the attribute `__array_interface__` from an object of type 'Polygon'.  With the exception of `AttributeError` NumPy will always raise this exception in the future.  Raise this deprecation warning to see the original exception. (Warning added NumPy 1.21)
    result[:] = values

tests/test_georaster.py::TestRaster::test_astype
  /home/runner/work/GeoUtils/GeoUtils/geoutils/georaster.py:502: UserWarning: dtype conversion will result in a loss of information. uint8 is the minimum type to represent the data.
    "dtype conversion will result in a loss of information. "

-- Docs: https://docs.pytest.org/en/stable/warnings.html
====================== 51 passed, 2436 warnings in 28.16s ======================
```

adehecq:
> Speaking of our `Number` + `mypy` issues, @adehecq, I've now added an `AnyNumber` alias in `geoutils._typing.py`!

Fantastic!

adehecq:
> We could have a documentation section "For developers" or something!

We definitely should! I would use it a lot :wink:

atedstone:
@erikmannerfelt, this PR looks good to go pending resolving a couple of conflicts on `import` lines - or is it still awaiting further work?

geoutils/misc.py
```diff
@@ -0,0 +1,92 @@
...
+        if isinstance(arr, Raster):  # If a Raster subclass, take its data. I don't know why mypy complains here!
+            arr = arr.data  # type: ignore
+        if isinstance(arr, np.ma.masked_array):  # If a masked_array, replace the masked values with nans
+            if "float" not in np.dtype(arr.dtype).name:
+                arr = arr.astype(float)
+            arrays.append(arr.filled(np.nan))  # type: ignore
+        else:
+            arrays.append(np.asarray(arr))
+
+    if np.shape(arrays[0]) != np.shape(arrays[1]):
```
adehecq, line 79: Could this test, and probably the next one, be tested at the very beginning? This way the whole block before that duplicates the data is not needed.

tests/test_georaster.py
```diff
@@ -47,9 +48,9 @@ def test_init(self) -> None:
 
         assert np.logical_and.reduce(
             (
-                np.array_equal(r.data, r2.data, equal_nan=True),
-                np.array_equal(r2.data, r3.data, equal_nan=True),
-                np.array_equal(r3.data, r4.data, equal_nan=True),
+                geoutils.misc.array_equal(r.data, r2.data, equal_nan=True),
```
adehecq, line 51: Would gu.misc not work? Otherwise, just 'from geoutils import misc' to shorten?

adehecq: Or directly import array_equal is it's all that's needed.

adehecq:
It's been a while since our last discussions. I believe it's good, apart maybe some optimization on the array_equal function to speed-up and/or reduce memory usage in certain cases.

adehecq:
Do you prefer to merge as it is, and possibly add the few recommendations as issues? Or will you have time to implement them today?

adehecq (approved):
Nothing more to add than the comments made a few months ago. This looks great !

geoutils/misc.py
```diff
@@ -0,0 +1,92 @@
...
+        if any(s in np.dtype(type(np.asanyarray(arr)[0])).name for s in ("<U", "str")):
+            strings_compared = True
+        if isinstance(arr, Raster):  # If a Raster subclass, take its data. I don't know why mypy complains here!
+            arr = arr.data  # type: ignore
+        if isinstance(arr, np.ma.masked_array):  # If a masked_array, replace the masked values with nans
+            if "float" not in np.dtype(arr.dtype).name:
+                arr = arr.astype(float)
+            arrays.append(arr.filled(np.nan))  # type: ignore
+        else:
+            arrays.append(np.asarray(arr))
```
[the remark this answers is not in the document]
rhugonnet: This is solved by our conversation, I think

geoutils/misc.py
rhugonnet: +1

rhugonnet (commented):
Only one comment from me:
Could be great to add a `test_misc.py` where you check that the new `array_equal` behaves as expected on a few cases.
Rest looks good!

geoutils/misc.py
erikmannerfelt: This is not possible, as we don't know before this point if the inputs are arrays. For example `Raster.shape` and `Raster.data.shape` are different. This needs to be put here.

geoutils/misc.py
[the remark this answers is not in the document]
erikmannerfelt: This does not duplicate the data if it has an `__array_interface__`, as mentioned in our chat earlier.

erikmannerfelt:
> Only one comment from me: Could be great to add a `test_misc.py` where you check that the new `array_equal` behaves as expected on a few cases. Rest looks good!

Thanks! There are lots of doctests that cover the functionality. Doctests are now automatically run with pytest.

geoutils/misc.py
adehecq: self.shape won't work, but np.shape should work, whether it's a Raster or a list.

erikmannerfelt: Nope!
```python
>>> import geoutils as gu
>>> import numpy as np
>>> r = gu.Raster(gu.datasets.get_path("landsat_B4"))
>>> np.shape(r)
(655, 800)
>>> np.shape(r.data)
(1, 655, 800)
```

adehecq: Huh! Why is that ??

erikmannerfelt: Don't know! It probably does some check first to see if the object has a `.shape` attribute, and returns that if so, maybe..

rhugonnet: Yes I noticed that a few days ago as well haha :stuck_out_tongue:

rhugonnet: Let's open a new issue! :laughing:

Issue #193, opened 2021-05-26
Include a "buffer" option in create_mask


adehecq:
Currently, Vector.create_mask will use the raw mask. It would be nice to have an option to automatically create a buffer around the mask, using either scipy's or opencv's morphological dilation.

adehecq:
Here's a quick example on how to do it:
```
from scipy.ndimage import morphology
import numpy as np

# Create a single 1 pixel within 0s
mask = np.zeros((11, 11))
mask[5, 5] = 1

# Dilate the mask with a buffer of size 2
buffer_size = 2
struct = np.ones((buffer_size*2 +1, buffer_size*2 + 1))
dil_mask = morphology.binary_dilation(mask, structure=struct)

```

Would be interesting to see how it compares with opencv's.

rhugonnet:
The issue I see here is that the buffer will be "limited" by the resolution of your raster.
Wouldn't it make more sense to apply a `.buffer()` to the Vector object, and then rasterize it? The buffer functions run hyper fast, even on a shapefile with hundreds of thousands of geometries!

adehecq:
> The issue I see here is that the buffer will be "limited" by the resolution of your raster.
> Wouldn't it make more sense to apply a `.buffer()` to the Vector object, and then rasterize it? The buffer functions run hyper fast, even on a shapefile with hundreds of thousands of geometries!

Good point! I didn't think it could be that fast, but if so, yes that's a good idea.

PR #228, opened 2021-07-08
Add a buffer option geovector.create_mask.

adehecq:


geoutils/geovector.py
```diff
@@ -162,15 +165,31 @@ def create_mask(
...
         else:
-            out_shape = rst.shape  # type: ignore
-            transform = rst.transform  # type: ignore
-            crs = rst.crs  # type: ignore
+            raise ValueError("`rst` must be either a str, geoutils.Raster or None")
+
+        # Crop vector geometries to avoid issues when reprojecting
+        left, bottom, right, top = bounds  # type: ignore
+        x1, y1, x2, y2 = warp.transform_bounds(crs, self.ds.crs, left, bottom, right, top)
+        self.ds = self.ds.cx[x1:x2, y1:y2]
```
erikmannerfelt, line 179: This is permanent, so the Vector will be modified after the function call!

geoutils/geovector.py
```diff
@@ -162,15 +165,31 @@ def create_mask(
...
 
         # Reproject vector into rst CRS
         # Note: would need to check if CRS are different
```
erikmannerfelt, line 182: I know this wasn't added now, but GeoPandas checks the CRS so there's no need for us to do that as well. This comment can be removed.

geoutils/geovector.py
```diff
@@ -162,15 +165,31 @@ def create_mask(
...
         vect = self.ds.to_crs(crs)
 
+        # Create a buffer around the features
+        if not isinstance(buffer, (int, float, np.number)):
+            raise ValueError(f"`buffer` must be a number, currently set to {type(buffer)}")
+        if buffer != 0:
+            vect.geometry = [geom.buffer(buffer) for geom in vect.geometry]
```
erikmannerfelt, line 189: What's the performance/quality advantages/disadvantages to dilating/eroding the mask after rasterization?

geoutils/geovector.py
adehecq: Good catch! I need to add a copy.

geoutils/geovector.py
adehecq: What do you mean? I added this step because I had issues after reprojecting certain RGI outlines to a local UTM zone. Some geometries turned into inf values.

erikmannerfelt: I mean the comment! If a `GeoDataFrame.to_crs()` is called with the same CRS as it currently has, it will just skip any computations. Therefore, there's no need to check if the CRSs are different, because the call below will reproject if needed, and won't if not needed.

geoutils/geovector.py
adehecq: No idea... But @rhugonnet suggested to do it this way in #193, saying buffer was "hyper fast". 
At least it is independent of the raster size. My guess is that for very small raster, erosion/dilation might be faster, but it will be much slower for larger rasters.

geoutils/geovector.py
adehecq: Fixed in eb21601

geoutils/geovector.py
adehecq: Oh ok, I thought you meant the lines above... Thanks for the clarification, will delete then!

geoutils/geovector.py
erikmannerfelt: Okay!

erikmannerfelt (approved):
Nice!! I feel like the tests could have been parametrized somehow instead of repeating some commands, but that is simply optional as the  tests look solid!

adehecq:
> Nice!! I feel like the tests could have been parametrized somehow instead of repeating some commands, but that is simply optional as the tests look solid!

It could, but it would require having a different function to test different features of create_mask. But as we say... perfect is the enemy of good! :wink:

PR #248, opened 2021-10-07
Fix development contribution instructions

### LINKED ISSUE #229: `CONTRIBUTING.md` install instructions are wrong (add `dev-environment.yml`?)
The current installation approach in [CONTRIBUTING.md](https://github.com/GlacioHack/GeoUtils/blob/main/CONTRIBUTING.md) does not work.

It says to run:
```bash
git clone https://github.com/GlacioHack/geoutils.git
cd geoutils
conda create -f environment.yml  # add '-n custom_name' if you want.
conda activate geoutils  # or any other name specified above
pip install -e .
```
Then "pytest can be run":
```bash
pytest
```

This is false, as not even pytest is installed with the `environment.yml`!

## Packages that are missing
- pytest
- pytest-xdist (kind of optional)
- sphinx 
- numpydoc 
- sphinx_rtd_theme 
- sphinx-autodoc-typehints 
- sphinxcontrib-programoutput
- flake8
- matplotlib 
- pre-commit

Add a `dev-environment.yml` with the aforementioned and base packages.

An alternative is to expand `environment.yml`, but this would install lots of unnecessary stuff for users that clone the latest main. A question, however, is how many **users** will clone the latest main? One? Two? Perhaps it's okay to just have one environment file! Dependencies are already defined in four places, so one more would be quite annoying:
- environment.yml
- setup.py
- dev-requirements.txt
- [meta.yaml](https://github.com/conda-forge/geoutils-feedstock/blob/master/recipe/meta.yaml)

## My preferred fix
Add the development dependencies to `environment.yml`!

atedstone:
This PR resolves #229. But, instead of the preferred fix noted there of creating an additional .yml file describing the dev environment or adding the dev dependencies to environment.yml, I propose simply using pip with the dev-requirements.txt file already in place.

I have also tried to improve the sequencing/readability of the document.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,15 +1,31 @@
 # How to contribute
 
+## Overview: making a contribution
+For more details, see the rest of this document.
+
+1. Fork and clone the repository.
+2. Set up the development environment.
+3. Create a branch for the new feature or bug fix.
+4. Make your changes, ensuring they are conventions-compliant.
+5. Commit, making sure to run `pre-commit` separately if not installed as git hook.
+6. Push.
+7. Open a Pull Request to discuss and eventually merge.
+8. You can now delete your feature branch.
+
+
 ## Rights
 The license (see LICENSE) applies to all contributions.
+
 
 ## Issue Conventions
 When submitting bugs, please fill out the bug report template in full.
 
 Please search existing issues, open and closed, before creating a new one.
 
+
 ## Git conventions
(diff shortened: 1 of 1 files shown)


adehecq:
This seems much more understandable, thanks !
My only concern is whether it's stable on the longer term to mix pip and conda install? If one wants to later update some packages that were installed with pip, or with pip dependencies, does it work?

erikmannerfelt:
That is true, @adehecq. It may be risky business to mix in general! We should have a dev-environment.yml I think

Issue #215, opened 2021-07-02
Re-structure georaster.py and move functions from xdem


adehecq:
These are two related issues:
1. Some features of xdem should be moved to geoutils since they are rather generic. This was brought up in [xdem issue #157](https://github.com/GlacioHack/xdem/issues/157). 
2. As discussed with @atedstone and @erikmannerfelt, the functions should be saved in `georaster.py`. But to avoid georaster to turn into a mess, the definition of the `Raster` class should be moved to a separate file (`_raster.py` is suggested). 

**Consequence:** After the change, the `Raster` class will not be available via `geoutils.georaster.Raster` anymore, but only via `geoutils.Raster`.

atedstone:
To be done:

- [ ] `georaster.py` to be replaced by `georaster` folder/submodule.
- [ ] Add `__init__.py`
- [ ] Move `interpolate.py` in from xdem
- [ ] Move generic spatial tools in from xdem (probably in `_.py` type files)
- [ ] Move `Raster` to `_raster.py`

PR #251, opened 2021-10-14
Create georaster submodule

atedstone:
Resolves the geoutils-only aspects of #215 , moving functionality from `xdem` will be dealt with in a separate PR.


adehecq:
Should the raster.py file be named _raster.py?

adehecq (approved):

Issue #247, opened 2021-10-06
Improve user documentation and add examples


atedstone:
We need to plan and write improved user documentation, including code examples in the style of the xdem documentation.

PR #252, opened 2021-10-14
Improve the documentation backend

erikmannerfelt:
One step toward #247.

## Changes
- Code examples can now be added!
- Function and classes are dynamically linked between the examples, the text, and the API
- Tests are added so docs-breaking changes cannot be made without failing CI.
- numpy, rasterio and matplotlib functions are now linked to their associated docs (more libraries can very easily be added)

## Going forward
The actual documentation needs to be improved! This is just a PR to fix the skeleton of the docs.

There's currently one thing that bothers me: The init modules are ignored. The consequence is that the `gu.*` like `gu.Raster` classes have to be added manually in the API. It's not hard, but can be forgotten if we make new classes. I'll work on this in the future, but I see it as a non-vital improvement.

You can see the proposed documentation structure here:
https://geoutils-erikm.readthedocs.io/en/docs-backend/index.html


erikmannerfelt:
Reathedocs is happy now!

PR #254, opened 2021-10-14
Correct the width/height calculation in geovector.rasterize

### LINKED ISSUE #199: Issue with rasterize when 'bounds' is set
In this line and next, width and height have been mixed up. This is an easy fix.
https://github.com/GlacioHack/GeoUtils/blob/4933b2a39269af82ec3ac8d3bc990bcb9691b077/geoutils/geovector.py#L227

A test could be added to check this.

atedstone:
Resolves #199 . I don't currently have any good ideas for implementing a test for this, but I have nevertheless added a rasterize() test which does, at least, allow us to check that the basic function works correctly.

### DIFF
--- a/geoutils/geovector.py
+++ b/geoutils/geovector.py
@@ -258,8 +258,8 @@ the provided raster file.
 
             # Calculate raster shape
             left, bottom, right, top = bounds
-            height = abs((right - left) / xres)
-            width = abs((top - bottom) / yres)
+            width = abs((right - left) / xres)
+            height = abs((top - bottom) / yres)
 
             if width % 1 != 0 or height % 1 != 0:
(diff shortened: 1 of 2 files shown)


adehecq:
> Resolves #199 . I don't currently have any good ideas for implementing a test for this, but I have nevertheless added a rasterize() test which does, at least, allow us to check that the basic function works correctly.

Cool !
Fort the test, would it not make sense to just hard-code the value? It will work until we change the reference data set...

atedstone:
I unearthed a problem with the rasterizing along the way - if a destination crs is requested then, as far as I can tell, the dataset needs to first be transformed to that before the bounds of the output can be determined. So I've sorted that as part of this PR. But, it shows that the test coverage for rasterize() requires further improvement - I'll raise as a separate issue.

Issue #256, opened 2021-10-15
Numpy representation of Rasters causes issue with arithmetic overloading


adehecq:
This is a bug on a yet non-existing code, but related to on-going work on issue #73.
The idea is to enable `rout = r1 + r2` with r1 being a Raster and r2 being either a Raster, a numpy array, or a number. This should return a Raster.
This works great in most cases, except in the situation when r1 is a numpy array and r2 a Raster. In that case, numpy's `__add__` method is used instead, which returns a numpy array. This is counter-intuitive in my opinion and should be avoided.
I believe the reason behind it is the numpy's interface, related to PR #210. When issue #73 is solved, one needs to check how to have both features working together.

erikmannerfelt:
Is this still an issue?

Also perhaps this should be done in conjunction with #261

rhugonnet:
This actually works now with #302 (the numpy interface is not taking over anymore):
```python
test = np.ones(r.shape) + r
type(test)
Out[15]: geoutils.georaster.raster.Raster
test = r + np.ones(r.data.shape)
type(test)
Out[17]: geoutils.georaster.raster.Raster
```
However, as you can notice, I have to change the shape to `.data.shape` on one side, and not another, otherwise:
```python
ValueError: Both rasters must have the same shape.
```
I'll leave this issue to be closed by #302 and open a new one on this.

PR #257, opened 2021-10-15
Add additional overloading of arithmetic functions

adehecq:
This PR adds the overloading of most useful arithmetic operations: +, -, *, /, //, % and **.
The arithmetic operation works whether the first or second operand is a Raster object, except if the first operand is a numpy array (see issue #256).


atedstone:
@adehecq , this looks superb. Very nice test functionality too! Looks good to go from my side.

rhugonnet (approved):
This looks amazing indeed! A solid core for geoutils arithmetics.
Congrats on this :partying_face:

PR #263, opened 2021-12-13
Fix several issues in spatial_tools.stack_rasters

adehecq:
**Issue 1:** stack_rasters was failing when providing an xdem.DEM instance as input, because the class does not allow multiband rasters
**Issue 2:** Progress bar was not displaying when intended, and also not propagated to merge_raster.

Reproducible example:
```
import xdem
import geoutils as gu
dem = xdem.DEM(gu.datasets.get_path("landsat_B4"))
raster = gu.Raster(gu.datasets.get_path("landsat_B4"))

# Progress bar not displaying (issue 2)
stacked_rasters = gu.spatial_tools.stack_rasters([raster, raster], progress=True)

# No progress bar option (issue 2)
merged_raster = gu.spatial_tools.merge_rasters([raster, raster])

# Error with DEM instance (issue 1)
# Fails with error "ValueError: DEM rasters should be composed of one band only"
stacked_rasters = gu.spatial_tools.stack_rasters([dem, dem])
```

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,5 +1,5 @@
 rasterio
-geopandas
+geopandas >= 0.10.0
 pyproj
 flake8
 pytest
--- a/environment.yml
+++ b/environment.yml
@@ -3,7 +3,7 @@ channels:
   - conda-forge
 dependencies:
   - python=3
-  - geopandas
+  - geopandas >= 0.10.0
   - matplotlib
   - pyproj
   - rasterio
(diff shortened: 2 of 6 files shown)


atedstone (approved):
These changes look good to me :)

adehecq:
Needed to force the geopandas version to >= 0.10.0 to fix an issue in geovector.buffer_without_overlap, that probably occurs due to a package upgrade.

Issue #240, opened 2021-08-16
Fix pickle error and allow `Raster`s to be serializable


erikmannerfelt:
Serialization and pickling are often useful for caching and sending data here and there. This should be supported for `Raster`s, but it currently isn't!

```python
>>> import geoutils as gu
>>> import pickle
>>> import json
>>> img = gu.Raster(gu.datasets.get_path("landsat_B4"))
>>> _ = pickle.dumps(img)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "stringsource", line 2, in rasterio._io.DatasetReaderBase.__reduce_cython__
TypeError: self._hds cannot be converted to a Python object for pickling
>>> _ = json.dumps(img)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/erik/.local/share/conda/geoutils/lib/python3.9/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
  File "/home/erik/.local/share/conda/geoutils/lib/python3.9/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/home/erik/.local/share/conda/geoutils/lib/python3.9/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/home/erik/.local/share/conda/geoutils/lib/python3.9/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type Raster is not JSON serializable
```

**Describe the solution you'd like**
Allow Rasters to be pickled for later use, and allow them to be described as JSON and other "serde" formats.

I assume that this requires a decoupling from the rasterio memoryfile (which is why it currently raises an error I think) when writing.

For normal `Raster`s, loading and saving can of course be achieved using the `load()` and `save()` functions. For derivations, however (like `SatelliteImage` or `dDEM`), metadata are lost in the process. Therefore, it would be great with serialization support.

Issue #122, opened 2021-03-15
Interp_points does not function if data is not loaded


rhugonnet:
Same as other is_loaded issues

adehecq:
Maybe better if you merge PR #120 first.

PR #265, opened 2022-01-11
Remove the memoryfile

### LINKED ISSUE #50: Method to close links to underlying dataset reader
By default, there are at least two links to the underlying rasterio reader saved within a Raster object: `.ds`, `.dataset_mask` . `.memfile` may also count.

There are some edge cases when this is problem. The one I've identified is that Rasters with these readers still open cannot be pickled when they are being distributed to compute nodes in multiprocessing situations.

Manually calling a method to close these links once the Raster has been created is a possible solution to this problem.

### LINKED ISSUE #153: GeoUtils or Rasterio doesn't close files properly.
**Describe the bug**
`Raster`'s don't get unloaded when they go out of scope. I don't know if this is a Rasterio bug or GeoUtils, but it should definitely be solved within GeoUtils somehow.

**To Reproduce**

```python
import resource

import geoutils as gu


def read_data():
    """Read a value from the data and return it."""
    r = gu.Raster(gu.datasets.get_path("landsat_B4"))
    return r.data[0, 0, 0]


ram = []
for _ in range(10000):
    read_data()
    ram.append(resource.getrusage(resource.RUSAGE_SELF).ru_maxrss)


plt.plot(ram)
plt.ylabel("RAM usage (KiB)")
plt.xlabel("Iterations")
plt.show()
```
Screenshot_20210412_124248

EDIT: The memory magnitude is off, probably by `4096` since I didn't consider the page size. The `resource` calls are honestly over my head!

**Expected behavior**
RAM usage should not increase linearly by that amount (only by the amount equivalent to the RAM usage statistics). The dataset should be close once it goes out of scope in `read_data()`.

 - OS: Linux
 - Environment: Latest source

### LINKED ISSUE #159: Allow `Raster.from_array` and `Raster.split_bands()` to work without copying.

### LINKED ISSUE #226: Raster.set_ndv does not work with multiple nodata values
**Describe the bug**
Raster.set_ndv supposedly accept several nodata values, but this hasn't been tested and it fails.
Related to #223 .

**To Reproduce**
```
import geoutils as gu
img = gu.Raster(gu.datasets.get_path("landsat_RGB"))
img.set_ndv([2,3, 4])
```
fails with error
>  ValueError: Given nodata value, [2, 3, 4], is beyond the valid range of its data type, uint8.

**Expected behavior**
This should not raise an error.
Apparently, this is not possible to create a raster with several nodata values at the moment with rasterio. This has been raised as an issue: https://github.com/mapbox/rasterio/issues/2234
In the meantime, a quick fix would be to add a NotImplementedError if several nodata values are passed.

irrelevant

**Additional context**
None

### LINKED ISSUE #238: `Raster.reproject()` raises confusing error when data are not loaded
**Describe the bug**
If the data are not loaded before calling reproject, it fails. The error is also not very useful...

**To Reproduce**
```python
import geoutils as gu

img1 = gu.Raster(gu.datasets.get_path("landsat_B4_crop"), load_data=True)
img2 = gu.Raster(gu.datasets.get_path("landsat_B4"), load_data=False)

img2.crop(img1)

#img2.load()

img2_reproj = img2.reproject(img1)
```
This raises an exception:
```
  File "/home/erik/temp/hello.py", line 8, in <module>
    img2_reproj = img2.reproject(img1)
  File "/home/erik/Projects/ETH/GeoUtils/geoutils/georaster.py", line 1022, in reproject
    dst_data, dst_transformed = rio.warp.reproject(self.data, **reproj_kwargs)
  File "/home/erik/.local/share/conda/geoutils/lib/python3.9/site-packages/rasterio/env.py", line 387, in wrapper
    return f(*args, **kwds)
  File "/home/erik/.local/share/conda/geoutils/lib/python3.9/site-packages/rasterio/env.py", line 610, in wrapper
    return f(*args, **kwds)
  File "/home/erik/.local/share/conda/geoutils/lib/python3.9/site-packages/rasterio/warp.py", line 368, in reproject
    _reproject(
  File "rasterio/_warp.pyx", line 410, in rasterio._warp._reproject
  File "rasterio/_warp.pyx", line 404, in rasterio._warp._reproject
ValueError: Invalid source
```
I presume what the error means is that `None` is an invalid source.

If `img2.load()` is uncommented in the above example, reprojection works fine.

**Expected behavior**
Like most other parts of the `Raster` class, `Raster._data` should probably be silently loaded in the `reproject()` function. An alternative (but I think much worse) solution is to just provide an actual descriptive error.

 - OS: Linux
 - Environment: latest main

**Additional context**
The reason I found this, and also why it bothers me, is that I often have a large reference DEM that I want to compare with a smaller DEM. Then I could do something like:
```python
small_dem = xdem.DEM("...")
# Crop the DEM first and then use `reproject()` to conform the resolution.
large_dem = xdem.DEM("...", load_data=False).crop(small_dem, inplace=False)
large_dem.load()
large_dem = large_dem.reproject(small_dem, resample="bilinear")
```
This could have been made in a one-liner, if this "bug" is fixed:
```python
small_dem = xdem.DEM("...")
# Crop the DEM first and then use `reproject()` to conform the resolution.
large_dem = xdem.DEM("...", load_data=False).crop(small_dem, inplace=False).reproject(small_dem, resample="bilinear")
```
This gets me thinking that maybe the `Raster()` instantiation could have a crop functionality already. Something like:
```python
large_dem = xdem.DEM("...", crop_to=small_dem)
```
... which would open the dataset, crop it, and then load the data.

### LINKED ISSUE #246: Are memfiles needed?
First, Issues #153 and #50 present problems encountered with memory management and closing Rasterio datasets. There has been some discussion about whether MemoryFiles (which usually originate from the class method `from_array`) might be responsible for some of these problems. 

Second, regardless of these Issues, it may be desirable to simplify the Raster class by removing dependencies on MemoryFiles. Currently, we suspect that data are duplicated (once in the memfile, once in .data), and also memfiles are likely to be an impediment to any future change in backend, e.g. rioxarray. @erikmannerfelt can probably comment further here. 

Remove mandatory use of MemoryFiles. This requires a broader conceptual shift - at the moment, various attributes are read directly from the underlying rio.DatasetReader regardless of whether disk file or memfile. So, the solution probably requires new property methods to yield `nx`, `ny`, etc.

N/A.

**Additional context**
N/A.

### LINKED ISSUE #259: `Raster.value_at_coords()` always reads from disk/memoryfile, not from `self._data`
**Describe the bug**
I found this as part of the go-away-MemoryFile PR. The function will never read from `self._data` but instead from the memoryfile or from disk. These are more or less kept in sync in the current GeoUtils (so it's just risky, not necessarily wrong), but with the upcoming PR, this will be an actual problem!

**To Reproduce**
```python
>>> import geoutils as gu
>>> import numpy as np
>>> img = gu.Raster(gu.datasets.get_path("landsat_B4"))
>>> x = np.mean([img.bounds.left, img.bounds.right])
>>> y = np.mean([img.bounds.bottom, img.bounds.top])
>>> img.value_at_coords(x, y)
255.0
>>> img -= 50  # This updates the memoryfile accordingly.
>>> img.value_at_coords(x, y)   # This produces the correct answer (right now)
205.0
>>> img = gu.Raster(gu.datasets.get_path("landsat_B4"))
>>> img._data -= 50  # This does not updated the memoryfile
>>> img.value_at_coords(x, y)  (this does not work because the memoryfile is not updated)
255.0
```

**Expected behavior**
I expect value_at_coords to read from the numpy array.

erikmannerfelt:
**Minor edits.**

Just kidding...

The memoryfile is gone! This means that the simplest explanation of a Raster is:

"A container of the transform object, CRS object, on-disk filename (if any), metadata dictionary and (if loaded) data ndarray of a raster."

The transform, CRS, metadata and data can be modified freely by the user, and all associated properties and attributes will (read "should") get updated accordingly.

Some advantages of this approach are:

- Data are not duplicated as they potentially were before (data in the ndarray and in the memoryfile)
- Operations are more transparent as they edit the ndarray, transform, CRS and/or metadata dict and nothing else (because there is nothing else!)
- After I/O operations, the in-memory version of the raster is completely disconnected from the disk version until another I/O operation is made, meaning that it will not lock the entire file.
- Easier to debug: The memoryfile was always a bit of a mystery!
- Easier to completely switch backends: `rio` is now only used for reading, writing and some builtin functions that work on numpy arrays + transform + CRS objects. This means that `rioxarray` could take its place for I/O if we want.
- Easier to pickle / serialize Rasters, as they are not bound to always-open I/O files

### Related issues
Closes #50 
Closes #226
Closes #238
Closes #246
Closes #259
Fixes the pickle part of #240 (json doesn't work yet)

### Issues that I need to find out may be solved:
Maybe c loses #122 (EDIT: Nope, still an issue)
Maybe closes #159 (EDIT: It does work!!)
Probably closes #153 (EDIT: It seems to be gone!! I did some quick load 6.6Gb raster loops and I can't replicate the OOM crashes anymore).

# But wait!
It's not finished...

There's one test that I need help with, probably from @adehecq  or @atedstone. 
`tests/test_georaster.py::TestRaster::test_reproj` fails, and I don't understand why...
I temporarily disabled it to make sure that everything else works fine, but we obviously have to find out why it's not working before merging!
Help!!


geoutils/georaster/raster.py
```diff
@@ -195,31 +210,97 @@ def __init__(
 
         :return: A Raster object
         """
+        self.driver: str | None = None
```
rhugonnet, line 213: What is the benefit of defining the Class attributes + type hints inside `__init__` instead of outside here?

geoutils/georaster/raster.py
```diff
@@ -195,31 +210,97 @@ def __init__(
...
-            # Save the absolute on-disk filename
-            self.filename = os.path.abspath(filename_or_dataset)
-            if as_memfile:
-                # open the file in memory
-                memfile = MemoryFile(open(filename_or_dataset, "rb"))
-                # Read the file as a rasterio dataset
-                self.ds = memfile.open()
+        elif isinstance(filename_or_dataset, (str, rio.io.DatasetReader, rio.io.MemoryFile)):
+
+            with ExitStack():
```
rhugonnet, line 248: Here a comment to shortly describe the behaviour of ExitStack() would help!

geoutils/georaster/raster.py
```diff
@@ -229,69 +310,92 @@ def __init__(
...
+            return self._disk_shape[0]
+        return int(self.data.shape[0])
 
-        # Save _masked attribute to be used by self.load()
-        self._masked = masked
+    @property
+    def count(self) -> int:
+        if self._disk_shape is not None:
+            return self._disk_shape[0]
+        return self.nbands
```
rhugonnet, line 323: Wait... here our `nbands` corresponds to rasterio `ds.data.shape[0]`, and then `count` corresponds to `nbands` again? I'm not sure I get the difference! Bit out of touch with those attributes, comments would help!

rhugonnet (approved):
Amazing! :partying_face: 
And impressive how little needed to be changed in the tests, in comparison to how much was modified in the Class! 
Out of curiosity: did you try to check locally if everything worked fine in `xdem` as well?

Unfortunately I don't have so much time to get into the details of the test failing with the thesis writing, I leave that to @adehecq and @atedstone! And other than that it all looks very good to me :smile: 

EDIT: in general, a couple more comments here and there would help quite a bit! Especially when using functions that we hardly use everyday (even something like `rio.transform.rowcol`, but for more unusual stuff also!)

geoutils/georaster/raster.py
erikmannerfelt: I would very much like input from @adehecq and/or @atedstone here, because I have no idea what the difference is either! I tried to mimic the old behaviour, but I'm not 100% sure it came out right

geoutils/georaster/raster.py
erikmannerfelt: Uuh probably no difference. There's a difference if no default argument is given:
```python
class Raster:
    driver: str | None   # This is only a type signature

Raster().driver   # Will raise an error because driver does not exist (but the expected type is known)
```
But when a default is defined, I honestly don't know the difference.

geoutils/georaster/raster.py
atedstone: `nbands` is a relic from the old georaster. `count` makes an appearance now because it's an attribute loaded by rasterio. In practical terms it looks like there probably isn't a difference between them. But, there could be some edge cases associated with the user specifying only certain bands to load, in which case `nbands` and `count` might not match. From a quick look, the former appears to be set according to the number of bands loaded, whereas `count` always matches rasterio I think.

atedstone: There is definitely some work needed here to iron out the differences. Perhaps deprecate `nbands` and, if found to be needed by testing, add an equivalent variable named e.g. `count_loaded`. But we should check that such an attribute is actually important to the functionality of the class before re-implementing it!!

adehecq: Yes, the difference between the two is the number of bands in the raster vs the number of bands loaded. I believe count is the former, but I'm not 100% sure without testing it.
I think both are useful in case one loads only a part of the data, and one of the names should be coherent with rasterio's naming convention. But the other one could be renamed if it makes things clearer.

geoutils/geoviewer.py
```diff
@@ -140,7 +140,8 @@ def main() -> None:
         new_shape = None
 
     # Read image #
-    img.load(out_shape=new_shape)
+    img.load()
+    img = img.reproject(dst_size=new_shape)  # type: ignore
```
adehecq, line 144: Just a note here. This sounds like one needs to load the whole array before being able to reproject, particuarly downsample it?
If so, this should be improved at some point.

tests/test_georaster.py
```diff
@@ -95,8 +95,9 @@ def test_info(self) -> None:
             "transform",
             "width",
         ]
-        for attr in default_attrs:
-            assert r.__getattribute__(attr) == r.ds.__getattribute__(attr)
+        with rio.open(datasets.get_path("landsat_B4")) as dataset:
```
adehecq, line 98: This is minor, but to be sure it's not broken in the future, maybe the path should be saved in a variable, and the same variable used at this line and L 80?

geoutils/georaster/raster.py
```diff
@@ -87,6 +86,47 @@ def _default_ndv(dtype: str | np.dtype | type) -> int:
...
+
+
+    \*\*kwargs: any additional arguments to rasterio.io.DatasetReader.read.
+    Useful ones are:
+    .. hlist::
+    * out_shape : to load a subsampled version
+    * window : to load a cropped version
+    * resampling : to set the resampling algorithm
+    """
+    if transform is not None and shape is not None:
```
adehecq, line 111: I don't really understand how shape and transform are to be used. Can you elaborate a bit?

geoutils/georaster/raster.py
adehecq: + 1

geoutils/georaster/raster.py
```diff
@@ -195,31 +210,97 @@ def __init__(
...
+                    self.filename = None
+
+                self.transform = ds.transform
+                self.crs = ds.crs
+                self.nodata = ds.nodata
+                self.name = ds.name
+                self.driver = ds.driver
+                self.tags.update(ds.tags())
+
+                self._disk_shape = (ds.count, ds.height, ds.width)
```
adehecq, line 266: Why rename them and use the underscore prefix? Those are metadata that can be useful to the user.

adehecq: By "them" I mean shape, indexes and dtypes.

geoutils/georaster/raster.py
```diff
@@ -137,40 +177,15 @@ class Raster:
...
-    name: str
-
     def __init__(
         self,
-        filename_or_dataset: str | RasterType | rio.io.DatasetReader | rio.io.MemoryFile,
+        filename_or_dataset: str | RasterType | rio.io.DatasetReader | rio.io.MemoryFile | dict[str, Any],
         bands: None | int | list[int] = None,
         load_data: bool = True,
         downsample: AnyNumber = 1,
         masked: bool = True,
```
adehecq, line 186: This is somewhat unrelated to this PR, but should we not enforce that self.data is a masked array?

geoutils/georaster/raster.py
```diff
@@ -195,31 +210,97 @@ def __init__(
...
-        # Or, image is already a Memory File.
-        elif isinstance(filename_or_dataset, rio.io.MemoryFile):
-            self.ds = filename_or_dataset.open()
+                down_width = int(np.ceil(self.width / downsample))
+                down_height = int(np.ceil(self.height / downsample))
+                out_shape = (nbands, down_height, down_width)
+                res = tuple(np.asarray(self.res) * downsample)
+                self.transform = rio.transform.from_origin(self.bounds.left, self.bounds.top, res[0], res[1])
+
+            if load_data:
```
adehecq, line 294: there seem to be two identical if statement in a row!

geoutils/georaster/raster.py
```diff
@@ -229,69 +310,92 @@ def __init__(
...
+    @property
+    def res(self) -> tuple[float | int, float | int]:
+        return self.transform[0], abs(self.transform[4])
+
+    @property
+    def bounds(self) -> rio.coords.BoundingBox:
+        left = self.transform[2]
+        top = self.transform[5]
+        res = self.res
+        return rio.coords.BoundingBox(
```
adehecq, line 352: It would be more straightforward to use this: https://rasterio.readthedocs.io/en/latest/api/rasterio.transform.html#rasterio.transform.array_bounds

geoutils/georaster/raster.py
adehecq: Nevermind, I see those are defined as properties!

geoutils/georaster/raster.py
```diff
@@ -329,47 +433,15 @@ def from_array(
...
-            nodata=nodata,
-            driver="GTiff",
-        ) as ds:
-
-            ds.write(data)
-
-        # Initialise a Raster object created with MemoryFile.
-        # (i.e., __init__ will now be run.)
-        return cls(mfh)
+        return cls({"data": data, "transform": transform, "crs": crs, "nodata": nodata})
```
adehecq, line 436: This is so simple that it seems somewhat magical 😉

geoutils/georaster/raster.py
```diff
@@ -491,7 +563,7 @@ def __add__(self: RasterType, other: RasterType | np.ndarray | Number) -> Raster
         out_data = self_data + other_data
 
         # Check that if no ndv was set, a default value is used
-        if (np.sum(out_data.mask) > 0) & (ndv is None):
+        if (np.sum(getattr(out_data, "mask", 0)) > 0) & (ndv is None):
```
adehecq, line 566: why was this change needed?

geoutils/georaster/raster.py
```diff
@@ -917,46 +907,59 @@ def crop(
...
-            crop_bbox = Polygon([(xmin, ymin), (xmax, ymin), (xmax, ymax), (xmin, ymax)])
-
-            crop_img, tfm = rio.mask.mask(self.ds, [crop_bbox], crop=True, all_touched=True)
-            meta.update(
-                {
-                    "height": crop_img.shape[1],
-                    "width": crop_img.shape[2],
-                    "transform": tfm,
-                }
+            # crop_bbox = Polygon([(xmin, ymin), (xmax, ymin), (xmax, ymax), (xmin, ymax)])
```
adehecq, line 911: All of this seems like a lot of changes. I'm worried that some of those changes might break stuff or introduce errors. I know we have a test suite, but as we've seen in the past, it's not bullet-proof ;-)
Why can't we use rasterio's functions anymore here? If a DatasetReader is needed, can't we even just temporarily open it?

tests/test_georaster.py
```diff
@@ -638,7 +649,7 @@ def test_set_ndv(self) -> None:
         assert np.all(ndv_index_2 == ndv_index)
 
         # Check that the number of no data value is correct
-        assert np.count_nonzero(ndv_index.data) == 112088
+        assert np.count_nonzero(ndv_index.data) == 111688
```
adehecq, line 652: why did this change?

tests/test_georaster.py
```diff
@@ -1028,7 +1039,7 @@ def test_numpy_functions(dtype: str) -> None:
     transform = rio.transform.from_origin(0, 5, 1, 1)
 
     # Create a raster from the array
-    raster = gu.Raster.from_array(array, transform=transform, crs=4326)
+    raster = gu.Raster.from_array(np.ma.masked_array(array), transform=transform, crs=4326)
```
adehecq, line 1042: Why is np.ma.masked_array needed each time? If so, why isn't it enforced in the `from_array` method directly? It's a bit annoying to have to write it each time...

geoutils/geoviewer.py
erikmannerfelt: This is the case in the current main as well. I agree that we should try to fix that in the future!

geoutils/georaster/raster.py
erikmannerfelt: Because it's not necessarily a masked array. Related to your suggestion if it should maybe always be one.

tests/test_georaster.py
erikmannerfelt: I was hoping that you wouldn't catch that hahaha :clown_face:. All of the other parts of the test work, so I assume that this has to do with the new handling of old nodata values. I can get to the bottom of this before merging if you wish.

tests/test_georaster.py
erikmannerfelt: We can enforce masked arrays! But I feel like that has to be a unanimous decision.

geoutils/georaster/raster.py
erikmannerfelt: Oh no, then I have to remember why it's there in the first place!!

geoutils/georaster/raster.py
erikmannerfelt: It's just to make sure it's actually actually loaded.





Just kidding, I'll fix it!

geoutils/georaster/raster.py
erikmannerfelt: Because then we have to write a whole memoryfile, basically copying the whole raster

geoutils/geoviewer.py
adehecq: So why previously this was one line and now two?

tests/test_georaster.py
adehecq: That would be great ! If we have to re-write the tests to make them pass, then there's no point in tests anymore 😉 At least it would be great to understand why that is.

tests/test_georaster.py
adehecq: And in the meantime, would the tests not pass without the conversion to a masked array?

geoutils/georaster/raster.py
adehecq: Yes, I thought that too at some point, but looking at the code of rasterio.mask.mask, I believe you can manage that with geometry_mask: https://github.com/rasterio/rasterio/blob/1.1.3/rasterio/mask.py#L108-L109

adehecq: Or actually geometry_window: https://github.com/rasterio/rasterio/blob/830fe9d59abe0040e6ac2a8af1db4ebc7eef3676/rasterio/features.py#L379

adehecq: Oh, but geometry_window requires a dataset... It's frustrating that there is not a function that would work with just transform and shape.

adehecq: What about this? https://github.com/rasterio/rasterio/blob/830fe9d59abe0040e6ac2a8af1db4ebc7eef3676/rasterio/windows.py#L255
(Sorry for the pile of message!)

tests/test_georaster.py
erikmannerfelt: I added these because you call the mask as if it most definitely exists, which it (right now) doesn't necessarily. For example (a bit further down):

```python
if np.sum(r3.data.mask) == 0:
```

geoutils/georaster/raster.py
erikmannerfelt: That looks great! Thanks for finding it! I think I looked for quite a while but ended up giving up on the search.

erikmannerfelt:
To make this PR simpler, should we start enforcing masked arrays? Note that this probably doesn't increase memory usage, as the mask can be empty (a single `False` value). Anyone against that, @adehecq, @rhugonnet, @atedstone ?

geoutils/georaster/raster.py
erikmannerfelt: Wouldn't geometry_mask work?

tests/test_georaster.py
adehecq: Stupid past me ! Fortunately I'm getting brighter everyday 😆

adehecq:
> To make this PR simpler, should we start enforcing masked arrays? Note that this probably doesn't increase memory usage, as the mask can be empty (a single `False` value). Anyone against that, @adehecq, @rhugonnet, @atedstone ?

I'm all up for it ! It would be more consistent. The only issue is that some regular numpy functions might not work as expected anymore...
I wonder why numpy arrays are not masked arrays with a single mask value set to False by default...

rhugonnet:
> To make this PR simpler, should we start enforcing masked arrays? Note that this probably doesn't increase memory usage, as the mask can be empty (a single `False` value). Anyone against that, @adehecq, @rhugonnet, @atedstone ?

Up for it as well! :+1:

erikmannerfelt:
> > To make this PR simpler, should we start enforcing masked arrays? Note that this probably doesn't increase memory usage, as the mask can be empty (a single `False` value). Anyone against that, @adehecq, @rhugonnet, @atedstone ?
> 
> Up for it as well! +1

Nice! If @atedstone is okay with it, I'll make it so.

geoutils/georaster/raster.py
```diff
@@ -195,31 +210,97 @@ def __init__(
...
+                out_shape = (nbands, down_height, down_width)
+                res = tuple(np.asarray(self.res) * downsample)
+                self.transform = rio.transform.from_origin(self.bounds.left, self.bounds.top, res[0], res[1])
+
+            if load_data:
+                if load_data:
+                    self._data = _load_rio(ds, bands=bands, masked=masked, out_shape=out_shape)
+                if isinstance(filename_or_dataset, str):
+                    self._is_modified = False
+                    self._disk_hash = hash((self.data.tobytes(), self.transform, self.crs, self.nodata))
```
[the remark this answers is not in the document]
erikmannerfelt: It generates a checksum (hash) to allow the `self.is_modified` property to properly respond if the data have been changed compared to those on-disk. Example:
```python
>>> img = gu.Raster(gu.datasets.get_path("landsat_B4"))  # Here, a checksum is generated on the data that are loaded
>>> img.is_modified  # This compares the saved checksum above with the current checksum
False
>>> img.data += 1
>>> img.is_modified  # This does the comparison again, which will be unequal
True
```

geoutils/geoviewer.py
erikmannerfelt: Ah right. Sorry I responded to your question but not to the actual issue. It was because I removed `**kwargs` from `Raster.load()`. I've taken it back in a commit, so I will resolve this comment soon.

geoutils/georaster/raster.py
erikmannerfelt: Implemented in def1811

geoutils/georaster/raster.py
erikmannerfelt: Fixed in b059bfb

geoutils/georaster/raster.py
erikmannerfelt: Elaborated in b059bfb

geoutils/geoviewer.py
erikmannerfelt: Fixed in 5d7fcfc

tests/test_georaster.py
erikmannerfelt: Changed in 74b4f9a. I honestly don't fully get the advantage, but now it's done!

geoutils/georaster/raster.py
erikmannerfelt: What about this:
- `nbands`: Deprecate?
- `count`: The number of bands in the current data (`self.data.shape[0]` if loaded, otherwise `disk_count`)
- `disk_count`: The number of bands in the disk (also the default for `count` if data are not loaded)
- `indexes`: The rio indexes if the Raster was made from a rio dataset, otherwise just 1,2,3 etc.
- `bands`: Deprecate?

geoutils/georaster/raster.py
erikmannerfelt: Described in b059bfb

erikmannerfelt:
> Amazing! partying_face And impressive how little needed to be changed in the tests, in comparison to how much was modified in the Class! Out of curiosity: did you try to check locally if everything worked fine in `xdem` as well?
> 
> Unfortunately I don't have so much time to get into the details of the test failing with the thesis writing, I leave that to @adehecq and @atedstone! And other than that it all looks very good to me smile
> 
> EDIT: in general, a couple more comments here and there would help quite a bit! Especially when using functions that we hardly use everyday (even something like `rio.transform.rowcol`, but for more unusual stuff also!)

Just so everyone knows, this is with 100% certainty not compatible with the latest xdem. Once this PR is finished, I'll jump straight into fixing xdem to work.

geoutils/georaster/raster.py
erikmannerfelt: Hmm, I checked `geometry_mask` (which basically just calls `rasterize`). It's a bit unnecessarily complex because rasterization is not needed to just get the min/max bounds of the features. I'd prefer to not use `geometry_mask` in favour of speed.

tests/test_georaster.py
erikmannerfelt: Updated the test and documented it properly in 5008906 (which by the way is the first commit hash that I've seen only started with numbers!).

Basically, the previous nodata functionality was that `set_ndv()` just enlarged the mask with the new nodata values, e.g. the pseudocode below:
```python
>>> r.set_ndv(255)  # This masks, say, 300 values
>>> np.count_nonzero(r.data.mask)
300
>>> r.set_ndv(254)  # This masks, say 100 NEW values
>>> np.count_nonzero(r.data.mask)
400
```
In the new case, the `255` values would be unmasked before masking the new `254` value, so the nodata count after this code would be `100`. The functionality for ndarrays (which will be made redundant in this PR), values are not masked but are set to `nan`. There, the previous behaviour of "enlarging the nodata domain" still happens. But yeah, pure ndarrays will most likely not be supported pretty soon anyway.

geoutils/georaster/raster.py
erikmannerfelt: It is now enforced in d690b1d

tests/test_georaster.py
erikmannerfelt: These changes are reverted in 3e84043 as `Raster.data` is now always a masked_array

geoutils/georaster/raster.py
erikmannerfelt: Reverted this change in 3f95a7e since it's no longer needed.

erikmannerfelt:
Okay! In summary, what is left on this PR is:

1. The cropping test needs to be re-enabled and needs to work. I need help on this one!
2. The count/nbands/indexes/etc mess needs to be cleared up.

After this, I think it's ready!

erikmannerfelt:
@adehecq, do you think I shall update geoutils to 0.0.6 directly in this PR?

Earlier, we have incremented versions in separate PRs or pushes directly to main. I don't really see the advantage of this, however, so I might as well change it here directly.

adehecq:
> @adehecq, do you think I shall update geoutils to 0.0.6 directly in this PR?
> 
> Earlier, we have incremented versions in separate PRs or pushes directly to main. I don't really see the advantage of this, however, so I might as well change it here directly.

I would make it in a separate PR, just for clarity. And also because you need to include a summary of changes for the whole version, and you won't know that until this PR is closed 😉

erikmannerfelt:
> > @adehecq, do you think I shall update geoutils to 0.0.6 directly in this PR?
> > Earlier, we have incremented versions in separate PRs or pushes directly to main. I don't really see the advantage of this, however, so I might as well change it here directly.
> 
> I would make it in a separate PR, just for clarity. And also because you need to include a summary of changes for the whole version, and you won't know that until this PR is closed wink

The summary of changes is for tagging the commit that marks the version increment, so that comes later either way. But sure, I can wait with it.

adehecq:
Related to issue #272, @erikmannerfelt, with this PR, would we still have access to the rasterio dataset?
It seems like we won't. 
Would it be possible to keep self.ds in case the raster was directly loaded from file. Of course, it won't work if the raster was modified afterward. This can be useful to perform some operations like cropping, reprojecting etc, without loading the data into memory...

adehecq:
Check that the latest rasterio release does not solve the memory issue? 
https://rasterio.groups.io/g/main/topic/92199259, mostly [#2141](https://github.com/rasterio/rasterio/pull/2141).

adehecq:
> There's one test that I need help with, probably from @adehecq or @atedstone.
> `tests/test_georaster.py::TestRaster::test_reproj` fails, and I don't understand why...
> I temporarily disabled it to make sure that everything else works fine, but we obviously have to find out why it's not working before merging!
> Help!!

I figured out why it failed. You updated the self.data property. Before, it used to just return `self._data`, but now you've added a condition to check that data was loaded:
```
        if not self.is_loaded and self._data is None:
            raise ValueError("Data are not loaded")
 ```
 However, in several instances, we are running the test `if self.data is not None`, which raises an error in case the data is not loaded with your update.
I will try updating those cases.

geoutils/georaster/raster.py
adehecq: I think we should stay consistent with rasterio's naming convention and hence have `count` and `indexes` match the definition of rasterio:
- `count`: band count on disk
- `indexes`: The 1-based indexes of each band in the dataset

We can have an equivalent of each for bands loaded, i.e. `count_loaded` and `indexes_loaded` (not sure the latter is really needed...).
What do you think @erikmannerfelt, @atedstone @rhugonnet ? 
Others should be deprecated.

rhugonnet: Sounds good :+1:

Issue #233, opened 2021-07-28
Raster.save does not take into account mask.


adehecq:
**Describe the bug**
When modifying the mask of a raster, without setting the corresponding values to nodata, the mask is not taken into account upon saving.

**To Reproduce**
```
import numpy as np
import geoutils as gu
# Load data
r = gu.Raster(gu.datasets.get_path('landsat_B4'))
outlines = gu.Vector(gu.datasets.get_path('glacier_outlines'))
# Mask glaciers
glacier_mask = outlines.create_mask(r)
r.data[glacier_mask] = np.ma.masked
# Check that it worked and save
r.show()
r.save('test.tif')
# Load the saved raster and check that masked values have been ignored
test = gu.Raster('test.tif')
test.show()
```

**Expected behavior**

Saved raster should have masked values set to nodata.

**Ideas of solution**

There are several options:
1) apply mask to both masked values and nodata values
2) raise a warning if mask and nodata disagree, then 1) 
3) raise an error if mask and nodata disagree
What do you think @atedstone @erikmannerfelt ?

erikmannerfelt:
That's definitely not optimal! Another thing that's been bugging me for a long while is that rasterio and gdal don't really agree on `nan` handling. Some GDAL calculations don't work at all with nans, right? They "are supposed to have a nodata value", so mixing nodata with nan (and then the mask on top!) makes a true mess.

One solution would be to add a keyword argument to `Raster.save()` that does "convert all nans to the nodata value (if any)" which is True by default (so it can be disabled if you know what you're doing).

So my suggestion would be:
1. Convert the mask AND `nan`s to the nodata value. (default behavior)
2. Skip the `nan` check but trigger a warning (or raise an error, but I'd prefer warning) if the mask is inconsistent. (this behavior is triggered with a keyword argument)

atedstone:
I'm not really got experience with (GDAL-)masked rasters, so cannot comment in detail here. But, I think that @erikmannerfelt 's development of the proposed options sounds like a good way forward.

Issue #235, opened 2021-07-29
Consistent masking of Raster object


rhugonnet:
Implement a consistent way of masking Raster objects, to ensure codes are easily reproducible with ndarray, masked arrays and Raster subclasses.
See below.

It's bad practice to use NaNs with the masked arrays. The values should be masked instead. I just read on the numpy.ma module that the best way to do it is
```
dh.data[mask_glacier] = np.ma.masked
```
The only issue is that dh.data.data is unchanged, so if one were e.g. to save the new array, the mask would not be preserved. To ensure that, one must also change the data attribute:
```
dh.data.data[mask_glacier] = dh.nodata
```
Which is not very intuitive I must admit.

_Originally posted by @adehecq in https://github.com/GlacioHack/xdem/pull/159#discussion_r678332285_

adehecq:
This is quite related to #233. I think it should just be masked as suggested in the first example. The second line is actually only needed because save does not take into account the mask, only the nodata value.

rhugonnet:
Yes, but the issue I see is that probably less than 5% of programmers will know that the first example is the way to go, it's really not intuitive/widely used.

Maybe we should encapsulate this in the `data.setter`? 
And, for instance, if the users is calling `dh.data = np.nan`, then we run the first example? There could be a couple convenient cases where we update the mask like this?

adehecq:
Well, we should make it clear somewhere in the documentation.
I don't think the data.setter is used when changing only a view of self.data (e.g. `self.data[mask] = 0`) but only when updating the whole attribute (e.g. `self.data = np.zeros((10, 10))`) but I'll double check.

adehecq:
Another option that came to my mind is to create a method like `gu.Raster.mask` that reads in a boolean array and does the masking. This would be more intuitive than having to edit `self.data.mask`.

PR #271, opened 2022-05-11
Add a masking functionality

adehecq:
This is in reply to issue #235.
I added a new method to Raster to set the data mask.

Example:
```
import geoutils as gu
r = gu.Raster(gu.datasets.get_path(dataset))
mask = (r.data == 126)
r.set_mask(mask)
r.show(cmap='gray')
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,7 +24,7 @@ repos:
 
         # Format the code aggressively using black
         - repo: https://github.com/psf/black
-          rev: 21.6b0
+          rev: 22.3.0
           hooks:
                   - id: black
                     args: [--line-length=120]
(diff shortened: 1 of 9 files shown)


rhugonnet (approved):
Looks great! This will be super useful :+1:. Doing this operation for masked arrays (especially within another object) is not something intuitive for all. 
I have only a little comment on the tests: we could add an assertion using `np.array_equal` to check that the exact content of the mask is preserved? Right now if I got it right, it's only the counts.

adehecq:
> I have only a little comment on the tests: we could add an assertion using `np.array_equal` to check that the exact content of the mask is preserved? Right now if I got it right, it's only the counts.

Good point ! Especially when testing that the previous mask is preserved. Summing the count works if the two masks do not overlap (which is the case here), otherwise not...
Fixed with c2fe2a1.

adehecq:
While I was on it, I also fixed issue #233.

Issue #273, opened 2022-05-18
Raster.reproject raises unclear error when data is not loaded


adehecq:
**Describe the bug**
The error raised by Raster.reproject when self.data is not loaded is not intuitive.

**To Reproduce**
```
import geoutils as gu
import rasterio as rio
raster = gu.Raster(gu.datasets.get_path("landsat_B4"), load_data=False)
raster.reproject(dst_crs=rio.crs.CRS.from_epsg(4326))
```
raises error

> TypeError: cannot unpack non-iterable NoneType object

**Expected behavior**
It should clearly state that self.data should be loaded.

**System (please complete the following information):**
Irrelevant

PR #274, opened 2022-05-18
Make reproject work even with data not loaded

adehecq:
This is mostly in reply to issue #273.

### DIFF
--- a/geoutils/georaster/raster.py
+++ b/geoutils/georaster/raster.py
@@ -996,7 +996,7 @@ Must be a Raster, np.ndarray or single number."
         dst_nodata: int | float | None = None,
         src_nodata: int | float | None = None,
         dtype: np.dtype | None = None,
-        resampling: Resampling | str = Resampling.nearest,
+        resampling: Resampling | str = Resampling.bilinear,
         silent: bool = False,
         n_threads: int = 0,
         memory_limit: int = 64,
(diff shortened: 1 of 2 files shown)


adehecq:
Sorry @erikmannerfelt, another conflict with your PR...

erikmannerfelt:
Looking good! There's just a test that needs fixing. Because you change the default reprojection algorithm (why by the way? I agree with the change but you haven't said why bilinear should be default!), some tests fail sporadically because the values are no longer exact. For some reason that I can't explain, only the Windows tests fail!

Here's the line that fails in the Windows CI:

https://github.com/GlacioHack/GeoUtils/blob/c101c3ed1e7424e0492b453b08f34aaa52b0569a/tests/test_georaster.py#L508

If you change this to use `np.isclose` instead, it should work. It's comparing `13.048369 == 13.0` so a one-decimal tolerance should be sufficient.

adehecq:
> Looking good! There's just a test that needs fixing. Because you change the default reprojection algorithm (why by the way? I agree with the change but you haven't said why bilinear should be default!), some tests fail sporadically because the values are no longer exact. For some reason that I can't explain, only the Windows tests fail!
> 

You're right, that is the reason for the tests failing. I will revert to the previous behavior and make the change in a separate PR, to be more transparent.
The test also need to be updated to pass only for nearest interpolation.

PR #275, opened 2022-08-12
Set default resampling to bilinear instead of nearest in reproject

adehecq:
I think 'nearest' should not be the default resampling algorithm in reproject. For non integer warping, it will create artificial shifts between the two rasters. 
Bilinear is the safest option as it is relatively fast, preserves range and will have an effect even for subpixel shifts.
Any thought against that?

### DIFF
--- a/geoutils/georaster/raster.py
+++ b/geoutils/georaster/raster.py
@@ -996,7 +996,7 @@ Must be a Raster, np.ndarray or single number."
         dst_nodata: int | float | None = None,
         src_nodata: int | float | None = None,
         dtype: np.dtype | None = None,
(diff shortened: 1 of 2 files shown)


rhugonnet:
Fully agree. Maybe it should only be "nearest" when the dtype is boolean?

rhugonnet (approved):

adehecq:
> Fully agree. Maybe it should only be "nearest" when the dtype is boolean?

Yes I agree, but boolean types do not exist with GTiff, so there's no need for that specific case. Anyway, bilinear would yield the same result, it would just be longer...

PR #284, opened 2022-08-29
Move examples to `geoutils-data` including an additional `float32` raster with NaNs

### LINKED ISSUE #117: Download sample data rather than having in repo
Similarly to xdem, we could have a script to download the data from archives, like google cloud?
Here's an example of link to Landsat image:
gdalinfo /vsicurl/https://storage.googleapis.com/gcp-public-data-landsat/LE07/01/140/041/LE07_L1TP_140041_20001030_20170209_01_T1/LE07_L1TP_140041_20001030_20170209_01_T1_B2.TIF

### LINKED ISSUE #190: Revisit the Landsat dataset
**Describe the bug**
As spotted by @erikmannerfelt, the "landsat_B4_crop" dataset has a larger extent, towards the east, than "landsat_B4". This is not very intuitive and should be fixed.

**To Reproduce**
```
import geoutils as gu
r1 = gu.Raster(gu.datasets.get_path('landsat_B4'))
r2 = gu.Raster(gu.datasets.get_path('landsat_B4_crop'))
r2.bounds.right > r1.bounds.right
```

**Expected behavior**
landsat_b4_crop should have an extent contained within landsat_B4.

 - OS: irrelevant
 - Environment : irrelevant

rhugonnet:
This PR:

- Removes the `geoutils/datasets` folder that contained example data;
- Adds a `geoutils/examples.py` module that downloads example data from a fixed commit in the repository [geoutils-data](https://github.com/GlacioHack/geoutils-data) created just for this purpose;
- Adds an example based on an ASTER DEM on Exploradores glacier, Northern Patagonian Icefield which brings diversity to the testing suite because: it is coded in `float32` instead of `uint8` for the Everest example, contains nodata when the other does not, has a nodata value of `-9999` when the other has none;
- Modifies the cropped Everest Landsat file to have smaller extent in all directions (see #190);
- Updates all occurences of `geoutils.datasets` into `geoutils.examples`;
- Runs all tests on two examples whenever possible using `@pytest.mark.parametrize()`.

Updating the tests made one bug appear: #286, the related `test_info` of `test_georaster.py` is temporarily skipped.

Resolves #117
Resolves #190

With the new tests, found 2 bugs:

- The `set_ndv` function does not update array, opened issue #286;
- The `crop` function has a wrong right bound, opened issue #287.


geoutils/examples.py
```diff
@@ -0,0 +1,74 @@
...
+        tmp_dir_name = os.path.join(temp_dir.name, [dirname for dirname in os.listdir(temp_dir.name) if os.path.isdir(os.path.join(temp_dir.name, dirname))][0],
+        "data", dir_name)
+
+        # Copy the temporary extracted data to the example directory.
+        copy_tree(tmp_dir_name, os.path.join(EXAMPLES_DIRECTORY, dir_name))
+
+def get_path(name: str) -> str:
+    """
+    Get path of example data
+    :param name: Name of test data (listed in xdem/examples.py)
```
adehecq, line 66: It would be great to have a way to list all the options available. For example, by returning the keys of FILEPATHS_DATA. I always need to go back to the source code to remember what is the exact name. And since the names are quite long, it would be useful!

rhugonnet: You mean in the description of the function? I generally just do `print(examples.FILEPATHS_DATA)`

adehecq: Well, yes, but it's not straightforward that the keys are found in examples.FILEPATHS_DATA. In datasets, there used to be a dictionary called `available`, that returns the available dataset.
We should at least state here in the docstring that the list can be retrieved with `list(examples.FILEPATHS_DATA.keys())`. But ideally, in examples one would have two functions:
- one listing the available examples (missing)
- one providing the paths to a given example (already existing)

rhugonnet: Not sure if we should mirror the previous `available` dict: it is basically the same dictionary as `FILEPATHS_DATA`, but linked the example names to the basename path instead of the full path. Seems like duplicating things...
Or `available` could just be a list of names, and not a dict?

rhugonnet: I added `available` as a list, pointed to it in the description of `get_path`, and added `_..` in front of the other variables.

geoutils/examples.py
```diff
@@ -0,0 +1,76 @@
...
+        "data", dir_name)
+
+        # Copy the temporary extracted data to the example directory.
+        copy_tree(tmp_dir_name, os.path.join(_EXAMPLES_DIRECTORY, dir_name))
+
+def get_path(name: str) -> str:
+    """
+    Get path of example data. List of available files can be found in "examples.available".
+
+    :param name: Name of test data (listed in xdem/examples.py)
```
adehecq, line 68: It's detail, but the part in the brackets can be removed as it's redundant with the previous sentence now.

tests/test_examples.py
```diff
@@ -0,0 +1,48 @@
...
+    "everest_landsat_b4_cropped": "0e63d8e9c4770534a1ec267c91e80cd9266732184a114f0bd1aadb5a613215e6",
+    "everest_landsat_rgb": "7d0505a8610fd7784cb71c03e5b242715cd1574e978c2c86553d60fd82372c30",
+    "everest_rgi_outlines": "d1a5bcd4bd4731a24c2398c016a6f5a8064160fedd5bab10609adacda9ba41ef",
+    "exploradores_aster_dem": "dcb0d708d042553cdd2bb4fd82c55b5674a5e0bd6ea46f1a021b396b7d300033",
+    "exploradores_rgi_outlines": "19c2dac089ce57373355213fdf2fd72f601bf97f21b04c4920edb1e4384ae2b2"
+}
+
+
+@pytest.mark.parametrize(
+    "example", examples.FILEPATHS_DATA
```
adehecq, line 36: This needs to be updated, either with the underscore, or better with `available`.

tests/test_georaster.py
```diff
@@ -638,42 +681,43 @@ def test_value_at_coords(self) -> None:
...
         """
-        Read Landsat dataset and set 255 to no data. Save mask.
-        Then, set 254 as new no data (after setting 254 to 0). Save mask.
+        Read dataset and set a certain value (e.g., 255 or -9999) to no data. Save mask.
+        Then, set the value minus one as new no data (e.g., 254 or -10000), after rewriting the previous nodata to 0. Save mask.
         Check that both no data masks are identical and have correct number of pixels.
         """
-        # Read Landsat image and set no data to 255
-        r = gr.Raster(self.landsat_b4_path, masked=True)
+        # Read image and set no data to 255
```
adehecq, line 706: Need to update the comment

tests/test_georaster.py
```diff
@@ -638,42 +681,43 @@ def test_value_at_coords(self) -> None:
...
-        r.set_ndv(ndv=255, update_array=True)
+        # Set the nodata value to default ndv (e.g., 255) and update the mask accordingly.
+        r.set_ndv(ndv=_default_ndv(r.dtypes[0]), update_array=True)
         # Save the mask for validation
         ndv_index = r.data.mask.copy()
 
         # Now set to 254, after changing 254 to 0.
-        r.data[r.data == 254] = 0
+        r.data[r.data == _default_ndv(r.dtypes[0]) - 1] = 0
         # This will unset the mask of all masked 255 values
```
adehecq, line 718: See my comment regarding your issue #286. The tests show that this is the expected behavior.

tests/test_georaster.py
```diff
@@ -638,42 +681,43 @@ def test_value_at_coords(self) -> None:
...
         # Copy the original data to validate the mask later
         original_data = r.data.copy()
-        # Set the nodata value to 255 and update the mask accordingly.
-        r.set_ndv(ndv=255, update_array=True)
+        # Set the nodata value to default ndv (e.g., 255) and update the mask accordingly.
+        r.set_ndv(ndv=_default_ndv(r.dtypes[0]), update_array=True)
         # Save the mask for validation
         ndv_index = r.data.mask.copy()
 
         # Now set to 254, after changing 254 to 0.
```
adehecq, line 716: Update comment

adehecq (approved):
I approve so we can move on, but I think we should have created only a single data repository instead of a second one.

geoutils/examples.py
rhugonnet: Done

tests/test_examples.py
rhugonnet: Nice catch! Done

rhugonnet:
Thanks for the review. All checks passing, merging! :partying_face: 
Not sure it makes any difference for the data repos, I find it fine like that. In any case, we can easily change in the future: it's 2 lines for defining tarball + commit in `examples.py`, and 5min to create the data repo and copy the data in. A minimal effort now that it is set up.

Issue #287, opened 2022-08-29
`Raster.crop()` has an erroneous right bound


rhugonnet:
Looks like the right bound of `crop` is shifted by exactly one pixel. This was hidden because the tests could not check this bound due to #190.

The test is skipped starting in #284 as it fails:

```python
r = gr.Raster(landsat_b4_path)
r2 = gr.Raster(landsat_b4_crop_path)

# Get expected bounds
b = r.bounds
b2 = r2.bounds
b_minmax = (max(b[0], b2[0]), max(b[1], b2[1]), min(b[2], b2[2]), min(b[3], b2[3]))

# Crop to the smaller extent
r.crop(r2)

assert b_minmax == b_crop
[Out]: False
```
`r.bounds.right[2]` is shifted by 30 (= 1 pixel) compared to `r2.bounds.right[2]` which it should have been cropped to.

adehecq:
I don't see the issue. 
First, there is a line missing in your example, what is `b_crop`?
If I run this on my install, I get:

> In [30]: print(r.bounds)
> BoundingBox(left=484060.0, bottom=3098030.0, right=502000.0, top=3103250.0)
> 
> In [31]: print(b_minmax)
> (484060.0, 3098030.0, 502000.0, 3103250.0)
>
> In [35]: list(r.bounds) == list(b_minmax)
> Out[35]: True

So exactly the same...

Issue #285, opened 2022-08-29
Masked pixels not properly taken into account in Raster.reproject


adehecq:
**Describe the bug**
Raster.reproject does not take into account masked values if `nodata` or `fill_value` not properly set.

**To Reproduce**
```
import geoutils as gu
from geoutils import datasets
import numpy as np

# Load raster
landsat_b4_path = datasets.get_path("landsat_B4")
r = gu.Raster(landsat_b4_path)

# Mask pixels to the left
r.data[0, :, :100] = np.ma.masked

# Reprojet
r2 = r.reproject(dst_res=20)
r2.show()
```
-> the masked pixels are no longer masked. Also the default nodata value of 255 is chosen which masks valid pixels.

Not that the result is the same if using `set_mask`:
```
mask = np.zeros(r.data.shape)
mask[0, :, :100] = 1
r.set_mask(mask)
```

**Expected behavior**
1) The masked array should be used as reference, not only the nodata value.
2) If nodata was not set in the input but is required, a warning should be raised to state that the default nodata was used (which can be a problem for UInt8 rasters).

**System (please complete the following information):**
 Irrelevant

Issue #282, opened 2022-08-26
Fix issues in from_array


adehecq:
Despite our best efforts, PR #265 seems to have introduced bugs that are not caught by current tests.
The backbone is in `from_array` and some checks seem to have been lost in the PR after [this line](https://github.com/GlacioHack/GeoUtils/pull/265/files#diff-706a5e63234acd08def533a19aae4dc002438ad09b809802ddd09b1d51dff280L333), e.g.:
- [ ] the dimension of data is not forced to be 3D (as in rasterio)
- [ ] the mask is likely not always preserved

rhugonnet:
Also potentially after [this line](https://github.com/GlacioHack/GeoUtils/pull/265/files#diff-706a5e63234acd08def533a19aae4dc002438ad09b809802ddd09b1d51dff280L769) of `_update`, that doesn't exist anymore.

adehecq:
Actually, the issue does not seem to be in `from_array`, or at least I was not able to find one. But an issue occurs with reproject for rasters with masked values, e.g.:
```
import geoutils as gu
# Take any raster with gaps
r = gu.Raster('/Users/adehecq/work/RAGMAC/ragmac_xdem/data/raw/experiment_2/PK_Baltoro/DEM_2002.75631975.tif')
r2 = r.reproject(dst_res=20)
print(r2.data)
print(r.data)
```
-> `r2.data` and `r.data` have different fill_value. This might cause issues later.

rhugonnet:
Maybe the added lines in `reproject` [here](https://github.com/GlacioHack/GeoUtils/pull/265/files#diff-706a5e63234acd08def533a19aae4dc002438ad09b809802ddd09b1d51dff280R1243)?
More tomorrow...

Issue #266, opened 2022-01-14
Test python 3.10 (and 3.11?) in CI


erikmannerfelt:
Python 3.10 is now bundled in most distributions, and we should make sure that it works properly. Maybe it doesn't make sense to test four versions, so I suggest these:

- python 3.7 (for compatibility)
- python 3.9 (many will still use this)
- python 3.10 (all new users will use this)

Many will of course also be on 3.8, but if it works on 3.7 and 3.9, it probably works for 3.8 too.

A heads-up in case I'm not the one implementing this:
- rasterio 1.2 does not work with 3.10! rio 1.3 will work, and their latest master works, but the 1.3a2 does not. https://github.com/rasterio/rasterio/commit/329c481bfdad4fa3272d07a548001f5ec1e8bda4 fixed this on 26 November 2021, so any release after that should work.
- I've also had issues with fiona. This was also fixed (https://github.com/Toblerity/Fiona/commit/3b1e80bfba017651151942b46fa1744e7fee6842; 22 November 2021), but has not yet been found in a release.

erikmannerfelt:
In addition, it would be great to somehow stay ahead of the curve and test python 3.11 as well! (It's in alpha 3 at the time of writing)

adehecq:
Decided to add 3.10 but to include 3.11 only when it is stable.

PR #289, opened 2022-08-30
Fix several issues that arose since the MemoryFile PR #265

adehecq:
Since the MemoryFile PR #265, several functionalities seem to have been broken. Some bugs were revealed by current tests, but it also highlighted the lack of robust tests for certain functionalities.

- [x] solve #287
- [x] fix an issue with Raster.crop when using mode='match_extent' (an untested functionality)
- [x] improve the tests for Raster.crop
- [x] solve #285
- [x] improve the tests for Raster.reproject
- [x] solve #282
- [x] update the CI to test python versions 3.8 to 3.10, as discussed in #266


tests/test_georaster.py
```diff
@@ -409,48 +411,147 @@ def test_masking(self, example: str) -> None:
...
+        r = gr.Raster(raster_path)
+
+        # -- Test with cropGeom being a list/tuple -- ##
+        cropGeom = list(r.bounds)
+
+        # Test with same bounds -> should be the same #
+        cropGeom2 = [cropGeom[0], cropGeom[1], cropGeom[2], cropGeom[3]]
+        r_cropped = r.crop(cropGeom2, inplace=False)
+        assert r_cropped.bounds == r.bounds
+        assert gu.misc.array_equal(r.data, r_cropped.data)
```
rhugonnet, line 429: The `equal_nan` functionality of `np.array_equal` was introduced in NumPy 1.19.0, compatible with Python 3.6.
It was relevant to have this for backward compatibility a year ago when it was relatively new, but now we only use Python >3.6 anyway, and Numpy is at the 1.23 release already.
We should deprecate `gu.misc.array_equal` and switch fully to `np.array_equal` for simplicity. I'm opening an issue.

tests/test_georaster.py
```diff
@@ -409,48 +411,147 @@ def test_masking(self, example: str) -> None:
...
+        r_cropped = r.crop(cropGeom2, inplace=False)
+        assert list(r_cropped.bounds) == list(cropGeom2)
+        assert gu.misc.array_equal(r.data[:, rand_int:, :], r_cropped.data)
+
+        # -- Test with CropGeom being a Raster -- #
+        r_cropped2 = r.crop(r_cropped, inplace=False)
+        assert r_cropped2.bounds == r_cropped.bounds
+        assert gu.misc.array_equal(r_cropped2.data, r_cropped)
+
+        # -- Test with inplace=False (Default) -- #
```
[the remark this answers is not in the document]
adehecq: Good catch, fixed in fb7ca73.

tests/test_georaster.py
```diff
@@ -409,48 +411,147 @@ def test_masking(self, example: str) -> None:
...
+        # outlines = outlines.query(f"index == {np.argmax(outlines.ds.geometry.area)}")
+        # Crop and validate that it got smaller
+        r_cropped = r.crop(outlines, inplace=False)
+        assert r.data.size > r_cropped.data.size  # type: ignore
 
-        # Crop the raster to the outline and validate that it got smaller
-        r_outline_cropped = r.crop(outlines, inplace=False)
-        assert r.data.size > r_outline_cropped.data.size  # type: ignore
+        # b = r.bounds
+        # b2 = r2.bounds
```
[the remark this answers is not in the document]
adehecq: The problem is that it will not necessarily match, because of differences in resolution, and because the raster's extent might be smaller than the outlines extent.
I will try to find a way to check that using intersection.

tests/test_georaster.py
adehecq: See my comment in #291.

tests/test_georaster.py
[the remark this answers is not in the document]
adehecq: Solved with 9d22fc9

.github/workflows/python-app.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     strategy:
       matrix:
         os: ["ubuntu-latest", "macos-latest", "windows-latest"]
-        python-version: ["3.7", "3.8", "3.9"]
+        python-version: ["3.8", "3.9", "3.10"]
```
rhugonnet, line 20: You also need to change the minimum python version from `3.7` to `3.8` in:

- `.readthedocs.yml`
- `dev-environement` and `environment` files
- `setup.py`

geoutils/georaster/raster.py
```diff
@@ -1136,6 +1136,13 @@ def reproject(
             dst_nodata = self.nodata
             if dst_nodata is None:
                 dst_nodata = _default_ndv(dtype)
+                # if dst_nodata is already being used, raise a warning.
+                # TODO: for uint8, if all values are used, apply rio.warp to mask to identify invalid values
+                warnings.warn(
+                    f"For reprojection, dst_nodata must be set. Default chosen value {dst_nodata} exist in \
+self.data. This may have unexpected consequences. Consider setting a different nodata with \
```
rhugonnet, line 1143: Did you mean to write "Default chosen value **might** exist in self.data", or is there a test to add here before triggering the warning?

geoutils/georaster/raster.py
```diff
@@ -1252,7 +1259,14 @@ def reproject(
...
         if self.is_loaded:
-            dst_data, dst_transformed = rio.warp.reproject(self.data, **reproj_kwargs)
+
+            # All masked values must be set to a nodata value for rasterio's reproject to work properly
+            # TODO: another option is to apply rio.warp.reproject to the mask to identify invalid pixels
+            if src_nodata is None and np.sum(self.data.mask) > 0:
+                raise ValueError("No nodata set, use `src_nodata`.")
+
+            # Mask not taken into account by rasterio, need to fill with src_nodata
+            dst_data, dst_transformed = rio.warp.reproject(self.data.filled(src_nodata), **reproj_kwargs)
```
rhugonnet, line 1269: Tiens je connaissais pas `np.ma.filled`, y a tellement de choses différentes avec les masked_arrays ! C'est solide mais ça limite aussi l'accès aux utilisateurs, ça sera cool de finaliser rapidement les fonctions style `get_array` pour faciliter l'accès à une `np.ndarray` classique :)

tests/test_georaster.py
```diff
@@ -544,33 +544,83 @@ def test_reproject(self, example: str) -> None:
...
-        new_data = np.ones(r.shape)
-        transform = list(r.transform)
-        new_transform = rio.transform.Affine(
-            transform[0] / 2,
-            transform[1],
-            (r.bounds.left + r.bounds.right) / 2,
-            transform[3] / 2,
-            transform[4],
-            (r.bounds.bottom + r.bounds.top) / 2,
+        # -- Check proper errors are raised if nodata are not set -- #
```
rhugonnet, line 547: Is the "--" to make the comment a bit of a "header" a PEP8 convention?

tests/test_georaster.py
```diff
@@ -544,33 +544,83 @@ def test_reproject(self, example: str) -> None:
...
+        r_ndv = r.copy()
+        r_ndv.nodata = None
+        default_ndv = _default_ndv(r_ndv.dtypes[0])
+        r_ndv.data.data[~r_ndv.data.mask][0] = default_ndv
+
+        # 1 - if no src_nodata is set and masked values exist, raises an error
+        if np.sum(r_ndv.data.mask) > 0:
+            with pytest.raises(ValueError, match="No nodata set, use `src_nodata`"):
+                r_ndv.reproject(dst_res=r_ndv.res[0] / 2, dst_nodata=0)
```
rhugonnet, line 556: Do you need to specify `dst_nodata=0` here? Won't the `ValueError` be raised in any case?

tests/test_georaster.py
```diff
@@ -544,33 +544,83 @@ def test_reproject(self, example: str) -> None:
...
+        if r.nodata is None:  # specific for the landsat test case, default nodata 255 cannot be used, so use 0
+            r.set_ndv(0)
+
+        # -- Additional tests -- #
+
+        # - Create 2 artificial rasters -
+        # for r2b, bounds are cropped to the upper left by an integer number of pixels (i.e. crop)
+        # for r2, resolution is also set to 2/3 the input res
+        min_size = min(r.shape)
+        rand_int = np.random.randint(min_size / 10, min(r.shape) - min_size / 10)
```
rhugonnet, line 575: We probably want to add `np.random.seed(42)` above that line, to have the tests fully consistent.

tests/test_georaster.py
```diff
@@ -584,6 +634,27 @@ def test_reproject(self, example: str) -> None:
...
             plt.show()
 
+        # - Check that if mask is modified afterwards, it is taken into account during reproject - #
+        # Create a raster with (additional) random gaps
+        r_gaps = r.copy()
+        data_1d = r_gaps.data.reshape(-1)
+        valids = np.where(~np.ma.atleast_1d(data_1d).mask)[0]
+        assert len(valids) > 0  # sanity check
+        nsamples = 200
+        data_1d[np.random.choice(valids, nsamples, replace=False)] = np.ma.masked
```
rhugonnet, line 644: Same comment here on fixing a random state. Could even use the `subsample_raster` function of GeoUtils.

tests/test_georaster.py
```diff
@@ -584,6 +634,27 @@ def test_reproject(self, example: str) -> None:
...
+        assert np.sum(r_gaps.data.mask) - np.sum(r.data.mask) == nsamples  # sanity check
+
+        # reproject raster, and reproject mask. Check that both have same number of masked pixels
+        # TODO: should test other resampling algo
+        r_gaps_reproj = r_gaps.reproject(dst_res=dst_res, resampling="nearest")
+        mask = gu.Raster.from_array(
+            r_gaps.data.mask.astype("uint8"), crs=r_gaps.crs, transform=r_gaps.transform, nodata=None
+        )
+        mask_reproj = mask.reproject(dst_res=dst_res, dst_nodata=255, resampling="nearest")
+        tot_masked_true = np.sum(mask_reproj.data.mask) + np.sum(mask_reproj.data == 1)
```
rhugonnet, line 655: Can you comment a bit more why you sum those two? Also on the creating of random gaps above, some functions such as `atleast_1d` are not so common, so it'd avoid one of us Googling that in the future if we need to re-work/understand these lines :slightly_smiling_face:.

Side note: I also typically use `np.count_nonzero` and not `np.sum`. Here it's unimportant, but I thought it was always a detail until I realized that `np.count_nonzero` always outputs an integer, which is quite practical to make `typing` happy and consistent in some instances (who knew I'd eventually become typing-enthusiast :laughing: )

rhugonnet (approved):
Amazing work!! :partying_face: 
In particular the new tests which will make things a lot more robust. It's exhausting work  :muscle:, and not so rewarding in the short-term, but so much worth it in the long-term! 

We've come quite a way since a year and a half ago, I think we're really getting close to a stable, solid core that we (hopefully) won't have to change so much in the future :smiling_face_with_tear:. Then we can focus on adding new features (and the science that goes with it)!

.github/workflows/python-app.yml
adehecq: Yes, I saw that you have included this in xdem. But in geoutils, we don't have any restriction on versions for Python apart from >3. I'm unsure about adding it. We cannot guarantee that geoutils will be stable with version below 3.8, but on the other hand, it does not strictly require Python>=3.8. It works with 3.7 and maybe below, but some rounding issues seem to happen...

geoutils/georaster/raster.py
adehecq: Good catch, I think I forgot to add an if statement!

geoutils/georaster/raster.py
adehecq: `np.ma.filled` is used quite often throughout our code 😉 
The problem is that there is not a unique solution to go from a masked array to a regular array. Nodata can be set to NaN, but only for compatible dtypes, or it can be set to nodata, but then it is not ignored in calculations.
Here it's simply because rasterio expects a regular array with gaps filled to the set nodata. No need for the user to know that.

tests/test_georaster.py
adehecq: I don't think it is a convention. I used to have ## or ### like in markdown, but flake8 or black don't like it.
I never found anything about writing some sort of headers in comments, but I do like to split my code in larger blocks. I think I've seen this elsewhere and found it was a good idea.

tests/test_georaster.py
adehecq: If `dst_nodata` is not set, it will raise a warning. That's case 2 below.

tests/test_georaster.py
adehecq: The test should work regardless of the value of that number, if the range is set correctly. I would leave it purely random, unless it becomes an issue later.

tests/test_georaster.py
adehecq: I forgot about `subsample_raster` ! 🙃 I will use that functionality instead then. I have basically just replicated what `subsample_raster` is doing...

.github/workflows/python-app.yml
rhugonnet: It's not tested for python <3.7, so we absolutely need pip/conda to solve the environment with this condition. It's a web of packages out there we depend on (and might depend on us eventually), if they all left untested versions possible in their environment, all hell would break loose and nothing would work! :exploding_head: 

People can force another version of python manually quite easily if they want to try it.

adehecq: Ok, done in 8946ae7

geoutils/georaster/raster.py
adehecq: Fixed in f3f5f7a

tests/test_georaster.py
adehecq: Use `subsample_raster` since 0781e4b.

PR #295, opened 2022-09-05
Add test coverage from Coveralls in CI

rhugonnet:
This PR:

- Adds the workflow for Coveralls;
- Adds a `dev-environment.yml` file to facilitate the installation for developers and feed to CI, to have the exact same environment locally;
- Adds `pytest-xdist` in development dependencies to distribute `pytest` on several CPUs when possible;
- Changes the use of `TemporaryDirectory()` as a context manager in `examples.py` to avoid failures during CI tests.


adehecq (commented):
Great ! 🎉

PR #297, opened 2022-09-05
Check that `environment.yml` can import geoutils, and use diff with `dev-environment.yml` during CI

### LINKED ISSUE #296: Consistently test `environment.yml` and `dev-environment.yml`
**Problem common to GeoUtils and xDEM:**

Right now, during CI, we have no way of verifying that `environment.yml` actually works with the package because we use directly `dev-environment.yml`, which is needed in order to consistently reproduce the environment locally. It would lose a lot of time to test both.

One solution could be, during CI:

1. We install `environment.yml` first,
2. We run `python -c "import xdem"` to check that there are no error when importing the package with `environment.yml`,
3. We also run `flake8` there,
4. (Tricky point) We compute the difference between `dev-environment.yml` and `environment.yml` (separating pip and conda channels if necessary),
5. We update the CI environment with the packages specific to `dev-environment.yml`,
6. We run the tests and documentation building of CI normally.

rhugonnet:
All in the title

Resolves #296

### DIFF
--- a/geoutils/misc.py
+++ b/geoutils/misc.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 
 import functools
 import warnings
+import yaml
 
 import numpy as np
 import rasterio as rio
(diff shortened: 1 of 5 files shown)


geoutils/misc.py
```diff
@@ -154,6 +155,75 @@ def new_func(*args, **kwargs):  # type: ignore
     return deprecator_func
 
 
+def diff_environment_yml(fn_env: str, fn_devenv: str, print_dep: str = "both") -> None:
```
adehecq, line 158: Ok see my comment in the related issue. You can try, but this was probably not needed 😥

Issue #292, opened 2022-08-31
Ensure that masked values are always set to nodata in Raster


adehecq:
At many occasions, one can update the mask in `Raster.data` but the underlying values are unchanged.
This can lead ton inconsistencies when `self.data.data` is read, e.g. by other modules like rasterio. It would probably be safer to always update the values in `self.data.data`.
The first thing is to modify `set_mask` to ensure that. It will require setting a nodata value in all cases though, is that desired?

Here is a draft of set_mask:

```
    def set_mask(self, mask: np.ndarray, nodata: int | float | None = None) -> None:
        """
        Mask all pixels of self.data where `mask` is set to True or > 0.

        Masking is performed in place.
        `mask` must have the same shape as loaded data, unless the first dimension is 1, then it is ignored.

        :param mask: The data mask
        :param nodata: The nodata value to be used (by default uses self.nodata)
        """
        # Check that mask is a Numpy array
        if not isinstance(mask, np.ndarray):
            raise ValueError("mask must be a numpy array.")

        # Check that new_data has correct shape
        if self.is_loaded:
            orig_shape = self.data.shape
        else:
            raise AttributeError("self.data must be loaded first, with e.g. self.load()")

        if mask.shape != orig_shape:
            # In case first dimension is empty and other dimensions match -> reshape mask
            if (orig_shape[0] == 1) & (orig_shape[1:] == mask.shape):
                mask = mask.reshape(orig_shape)
            else:
                raise ValueError(f"mask must be of the same shape as existing data: {orig_shape}.")

        # Check that a nodata value is set, otherwise, need to set one
        # It should not be needed, but some functionnalities like rasterio's reproject require a nodata to be set.
        if self.nodata is None:
            if nodata is None:
                raise ValueError("A nodata must be set. Use self.set_ndv()")
            else:
                self.nodata = nodata

        # Mask array
        self.data[mask > 0] = np.ma.masked

        # Replace the masked values by nodata
        self.data = self.data.filled(self.nodata)
```

rhugonnet:
**Adapted from a comment in #300 for the record:**
My opinion would be to not address this issue at all: if one wants to work with a nodata-proof array, he can (1) use directly the masked array in `.data` or (2) use `get_nanarray()` to get a NaN array, or (3) run his calculation directly on the Raster object once I have fixed the `array_interface` to apply to the masked array.

We use to work with `.data.data` at the beginning, but we should not anymore, especially with the numpy array interface!

[...]

PR #300, opened 2022-09-05
Improve `data.setter` of `Raster` for consistent array output

### LINKED ISSUE #294: Wrong behaviour of `copy()` method with `new_array` from np.ndarray with NaNs
The `copy` method expects a `np.ndarray` as input for the `new_array` argument. However, if that array contains NaNs, the `Raster` object created will have as `.data` a `np.ma.masked_array` which contains unmasked NaNs.

This results in a lot of issues: applying `np.ma` functions, writing a Raster to disk in rasterio, propagating nodata when applying arithmetic functions to `Raster` objects, etc... Strangely, some systems support the writing of these unmasked NaNs to disk in TIFF, while it result in a break down on others.

We should probably check the input type in `copy`, or upstream in `from_array` i.e. in `Raster.__init__` if the issue is global to these routines, in order to create a masked_array that masks the NaNs automatically.

### LINKED ISSUE #298: Crop tests failing for `rasterio`<1.3
Sinces #289, tests fail for `rasterio`<1.3 due to bug fixes on the crop function. And lower version of RasterIO are still used during environment solving.

If we force `rasterio`>=1.3, we'll get the right behaviour and no issue, so let's go with that. Will need to fix #293, though.

rhugonnet:
This PR corrects the behaviour of `Raster.data.setter`, which is directly used by `Raster.__init__()` or `Raster.from_array()` to construct/update the `data` attribute. It also adds tests to ensure this behaviour will be constant.

**The previous behaviour was:**

1. A unmasked `np.ndarray` or a `np.ma.masked_array` can both be used to set a new `data`,
2. The shape was checked: if 2D broadcast to 3D, otherwise an error was raised,
3. The dtype was checked: if it didn't match the previous `dtype`, an error was raised.
4. The array was simply cast to a `np.ma.masked_array` with `fill_value` set from the `nodata`.

**The new behaviour modifies point 4. Now:**

- If the array is unmasked and has non-finite values, a mask of these non-finite values is created and passed to `np.ma.masked_array`. If the nodata was not defined to pass to fill_value, the default for the dtype is used.
- If the array is masked, but has non-finite values that are unmasked, the mask is extended to non-finite values. If the nodata was not defined to pass to fill_value, the default for the dtype is used.
- Otherwise, the array is cast normally to a np.ma.masked_array.

This behaviour directly affects: `__init__()`, `from_array()`, `copy()` and arithmetic operations that all use the `data.setter`.
**Points 1, 2 and 3 are now also thoroughly tested, in addition to point 1.**

For these reasons, this PR:

- Removes the setting of `nodata` within arithmetic functions, as this is now always done consistently by the `data.setter` after the `from_array` operation triggered by arithmetic operations,
- Inverts the lines in `__init__()` for setting `self.nodata` and `self.data` from a `dict` (passed by `from_array()`), putting the nodata first so it can be used by `self.data`, and so that `self.data` can possibly set a new one afterwards with the new behaviour,
- Adds `nodata` as a property so it cannot be set directly, as this is the role of `set_ndv()`,
- Adds None as a possible input/output of `nodata` in `set_ndv()`, for the sake of testing,
- Adds a `test_data_setter()` function that tests the behaviour of `data.setter` for all data types, for nodata as None or default value, and for all types of input (masked array with mask, masked array without mask, classic array). The function also checks that errors are properly raised for different `dtype` or `shape`,
- Adds tests to `test_copy` for the `new_array` attribute. Among other, it checks that `copy(new_array=)` can change the `dtype` of the output,
- Modifies the error of `data.setter` when trying to set a different `dtype` into "New data must be of the same type as existing data: {}. Use copy() to set a new array with different dtype, or astype() to change type.".
- Adds a warning that is raised when the nodata is set during the `data.setter` because of unmasked non-finite values.

Addionally, this PR:

- Forces `rasterio` > 1.3 to avoid failure of tests related to cropping on older versions,
- Allows None to be passed to `crs` for testing reasons,
- Prints the CRS directly in `info` instead of with `to_epsg()` that can trigger error for CRS that do not have an EPSG code,
- Removes some warnings.

This PR:
Resolves #298
Resolves #294

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -3,6 +3,7 @@ geopandas >= 0.10.0
 pyproj
 flake8
 pytest
+pytest-xdist
 pytest-lazy-fixture
 sphinx
 sphinx_rtd_theme
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -14,7 +14,7 @@ sphinx_autodoc_typehints
 matplotlib
 scipy
 pre-commit
-typing-extensions; python_version < '3.8'
+typing-extensions
 sphinx-gallery
 scikit-image
 
--- a/geoutils/georaster/raster.py
+++ b/geoutils/georaster/raster.py
@@ -798,7 +798,15 @@ Must be a Raster, np.ndarray or single number."
         """
         Set the contents of .data.
 
-        new_data must have the same shape as existing data! (bands dimension included)
+        The data setter behaviour is the following:
+
+        1. Writes the data in a masked array, whether the input is a classic array or a masked_array,
+        2. Reshapes the data in a 3D array if it is 2D that can be broadcasted, raises an error otherwise,
+        3. Raises an error if the dtype is different from that of the Raster, and points towards .copy() or .astype(),
+        4. Sets a new nodata value to the Raster if none is set and if the provided array contains non-finite values
+            that are unmasked (including if there is no mask at all, e.g. NaNs in a classic array),
+        5. Masks non-finite values that are unmasked, whether the input is a classic array or a masked_array. Note that
+            these values are not overwritten and can still be accessed in .data.data.
 
         :param new_data: New data to assign to this instance of Raster
 
@@ -841,6 +849,11 @@ Must be a Raster, np.ndarray or single number."
             and self.nodata is None
             and np.count_nonzero(~np.isfinite(new_data.data[~new_data.mask])) > 0
         ):
+            warnings.warn(
+                "Setting default nodata {:.0f} to mask non-finite values found in the array, as "
+                "no nodata value was defined.".format(_default_ndv(dtype)),
+                UserWarning,
+            )
             self._nodata = _default_ndv(dtype)
 
         # If the new data is not masked (a classic ndarray) and contains non-finite values such as NaNs, define a mask
@@ -901,7 +914,7 @@ Must be a Raster, np.ndarray or single number."
             f"Size:                 {self.width}, {self.height}\n",
             f"Number of bands:      {self.count:d}\n",
             f"Data types:           {self.dtypes}\n",
-            f"Coordinate System:    {self.crs}\n",
+            f"Coordinate System:    {[self.crs.to_string() if self.crs is not None else None]}\n",
             f"NoData Value:         {self.nodata}\n",
             "Pixel Size:           {}, {}\n".format(*self.res),
             "Upper Left Corner:    {}, {}\n".format(*self.bounds[:2]),
--- a/geoutils/georaster/raster.py
+++ b/geoutils/georaster/raster.py
@@ -825,7 +825,7 @@ Must be a Raster, np.ndarray or single number."
         if str(new_data.dtype) != dtype:
             raise ValueError(
                 "New data must be of the same type as existing\
- data: {}".format(
+ data: {}. Use copy() to set a new array with different dtype, or astype() to change type.".format(
                     dtype
                 )
             )
--- a/geoutils/georaster/raster.py
+++ b/geoutils/georaster/raster.py
@@ -239,11 +239,11 @@ class Raster:
         """
         self.driver: str | None = None
         self.name: str | None = None
-        self.nodata: int | float | list[int] | list[float] | None = None
         self.filename: str | None = None
         self.tags: dict[str, Any] = {}
 
         self._data: np.ma.masked_array | None = None
+        self._nodata: int | float | list[int] | list[float] | None = nodata
         self._bands = bands
         self._masked = masked
         self._disk_hash: int | None = None
@@ -254,10 +254,11 @@ class Raster:
 
         # This is for Raster.from_array to work.
         if isinstance(filename_or_dataset, dict):
+            # Important to pass the nodata before the data setter, which uses it in turn
+            self._nodata = filename_or_dataset["nodata"]
             self.data = filename_or_dataset["data"]
             self.transform: rio.transform.Affine = filename_or_dataset["transform"]
             self.crs: rio.crs.CRS = filename_or_dataset["crs"]
-            self.nodata = filename_or_dataset["nodata"]
             for key in filename_or_dataset:
                 if key in ["data", "transform", "crs", "nodata"]:
                     continue
@@ -290,7 +291,7 @@ class Raster:
 
                 self.transform = ds.transform
                 self.crs = ds.crs
-                self.nodata = ds.nodata
+                self._nodata = ds.nodata
                 self.name = ds.name
                 self.driver = ds.driver
                 self.tags.update(ds.tags())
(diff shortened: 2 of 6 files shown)


dev-requirements.txt
```diff
@@ -14,7 +14,7 @@ sphinx_autodoc_typehints
 matplotlib
 scipy
 pre-commit
-typing-extensions; python_version < '3.8'
```
adehecq, line 17: Before removing it, do you have an idea why this was enforced? Maybe @erikmannerfelt would know.

geoutils/georaster/raster.py
```diff
@@ -736,10 +719,7 @@ def __pow__(self: RasterType, power: int | float) -> RasterType:
         # Calculate the product of arrays and save to new Raster
         out_data = self.data**power
         ndv = self.nodata
```
adehecq, line 721: You can skip setting `ndv` and use directly `self.nodata` in the following line.

geoutils/georaster/raster.py
```diff
@@ -846,7 +835,24 @@ def data(self, new_data: np.ndarray | np.ma.masked_array) -> None:
                 f"New data must be of the same shape as existing data: {orig_shape}. Given: {new_data.shape[1:]}."
             )
 
-        self._data = np.ma.masked_array(new_data, fill_value=self.nodata)
+        # If the new data is not masked and has non-finite values, we define a fill_value
```
adehecq, line 838: Maybe you can update the docstring of data.setter to explain all this?

geoutils/georaster/raster.py
```diff
@@ -846,7 +835,24 @@ def data(self, new_data: np.ndarray | np.ma.masked_array) -> None:
...
+        if (not np.ma.is_masked(new_data) and self.nodata is None and np.count_nonzero(~np.isfinite(new_data)) > 0) or (
+            np.ma.is_masked(new_data)
+            and self.nodata is None
+            and np.count_nonzero(~np.isfinite(new_data.data[~new_data.mask])) > 0
+        ):
+            self._nodata = _default_ndv(dtype)
```
adehecq, line 844: Maybe we should raise a warning in case the default value is used in the array.

geoutils/georaster/raster.py
```diff
@@ -895,7 +901,7 @@ def info(self, stats: bool = False) -> str:
             f"Size:                 {self.width}, {self.height}\n",
             f"Number of bands:      {self.count:d}\n",
             f"Data types:           {self.dtypes}\n",
-            f"Coordinate System:    EPSG:{self.crs.to_epsg()}\n",
+            f"Coordinate System:    {self.crs}\n",
```
adehecq, line 904: But `self.CRS` is a rasterio object, so it does not really print nicely... If epsg does not exist, we should use `self.crs.to_string`.
If `self.crs` is not always a rasterio CRS (not so great...) then add a condition.

tests/test_georaster.py
```diff
@@ -195,6 +196,162 @@ def test_loading(self) -> None:
...
+            randint_dtype = dtype
+        else:
+            val_min = -99999
+            val_max = 99999
+            randint_dtype = "int32"
+
+        # Fix the random seed
+        np.random.seed(42)
+        arr = np.random.randint(low=val_min, high=val_max, size=(1, width, height), dtype=randint_dtype).astype(dtype)
+        mask = np.random.randint(0, 2, size=(1, width, height), dtype=bool)
```
adehecq, line 234: You can also mask e.g. the min values, so they are random and you know they exist.

tests/test_georaster.py
```diff
@@ -195,6 +196,162 @@ def test_loading(self) -> None:
...
+        assert r1.nodata == nodata
+        assert r2.nodata == nodata
+        assert r3.nodata == nodata
+
+        # Compare the consistency of the data setter whether it is passed a masked_array or an unmasked one
+        assert np.array_equal(r1.data.data, arr)
+        assert not r1.data.mask
+        assert np.array_equal(r2.data.data, arr)
+        assert not r2.data.mask
+        assert np.array_equal(r3.data.data, arr)
```
adehecq, line 266: `gu.misc.array_equal` checks that the array **and** mask are equal.

adehecq:
Nice ! 
Good idea to convert nodata to a property.

The only issue I see with setting all NaNs to nodata is that I have had cases of GTiff with certain values set to nodata, others to NaN, and they had slightly different meaning (e.g. nodata for pixels outside of the area observed, NaN where the processing did not work = gaps). With this change, it means we lose that information.
I feel that this may also hide some improper handling of nodata. If there is a NaN value and we know there should not, then we made a mistake. But now it's just going to be ignored.
I don't have a strong opinion on what is the best strategy, but I think we should weigh the pros and cons.

rhugonnet:
The values are not set to nodata, they are simply masked. So you can still extract the array in `self.data.data` and see where there are NaNs or normal values.
I guess this solves your issue :wink:

adehecq:
> The values are not set to nodata, they are simply masked. So you can still extract the array in `self.data.data` and see where there are NaNs or normal values. I guess this solves your issue 😉

Well... until we address [this issue](https://github.com/GlacioHack/geoutils/issues/292) ! 🙃 
Maybe a simple solution would be to make it an argument, like `mask_nans` which defaults to True. The argument should be available also in __init__ to prevent from overwriting actual NaNs upon loading.

rhugonnet:
My opinion would be to not address this issue at all: if one wants to work with a nodata-proof array, he can (1) use directly the masked array in `.data` or (2) use `get.array()`, or (3) even run the ufunc/array_function on the `Raster` object directly once I have fixed the __array_interface__. 

We use to do it at the beginning, but no one should work with `.data.data` anymore, especially with the numpy array interface!

dev-requirements.txt
rhugonnet: My guess is to support typing for Python < 3.7 which was probably not enabled by default in typing-extensions, as typing was not so widespread then. Should be good now.

geoutils/georaster/raster.py
rhugonnet: Done

geoutils/georaster/raster.py
rhugonnet: Done + test.

geoutils/georaster/raster.py
rhugonnet: Done with a condition

tests/test_georaster.py
rhugonnet: Seeds are recommended for unitests even when they should pass for any random value, e.g. to reproduce locally if there is ever an error, so it works nicely like this

tests/test_georaster.py
rhugonnet: Actually it uses the mask to replace masked values with NaNs, so this functionality already exists as `np.ma.allequal`. I'm not sure there any added value to it now that we are always using NumPy > 1.20 that has good internal functions. Adding this to the related issue for closing.

Regarding the current case: it cannot be handled by a single NumPy (or GeoUtils) function:  the mask is not defined, so `data.mask` is `False` without any shape. I think it can only be compared this way, which is not too much trouble.

adehecq (approved):
All good !!

PR #301, opened 2022-09-05
Remove `yaml` method for diff calculation of `environment.yml` and `dev-env.yml` 

rhugonnet:
Diff of environment files can be done more easily directly in conda, following @adehecq's comment in #297 

However, we lose one thing when using conda compared to the previous PR: we do not test anymore if there are dependencies listed in `environment.yml` that are not indicated in `dev-environment.yml`.

I have thus moved the function previously coded in `misc.py` into `test_misc.py`, and it will run during the test suite to verify that we have written the `environment.yml` and `dev-environment.yml` correctly. Coding this was not for nothing! :smile:


adehecq (approved):
The master of recycling! 😆

geoutils/misc.py
```diff
@@ -155,75 +154,6 @@ def new_func(*args, **kwargs):  # type: ignore
     return deprecator_func
 
 
-def diff_environment_yml(fn_env: str, fn_devenv: str, print_dep: str = "both") -> None:
```
adehecq, line 158: Just thinking, we could also test that the pip and conda environments are the same?

adehecq: I meant checking that the content of `requirement.txt` (or rather `setup.py` I guess?) is consistent with environment.yml. 
I'm mostly concerned about installing from source, which requires a `pip install -e .` command. If any requirement listed in `setup.py` is not in environement.yml, then it would cause pip to install it.

Issue #277, opened 2022-08-24
Add a `get_nan_array` method to the Raster class.


adehecq:
Currently, to convert the masked_array in `self.data` to a regular array with Nans, one can use the function in spatial_tools `get_array_and_mask`. It would be easier to make it a method of the Raster class directly.
I suggest naming it `Raster.get_nan_array`, unless there are other suggestions?

One might need to consider an equivalent for `get_mask` too.

PR #302, opened 2022-09-06
Improve array interface of `Raster` for consistency with `np.ma.masked_array` operations

### LINKED ISSUE #256: Numpy representation of Rasters causes issue with arithmetic overloading

### LINKED ISSUE #261: Inconsistent nodata handling through `array_interface` for `RasterType` objects
Linked to #235 and #231
The nodatas are not accounted for consistently when running `numpy` functions over `RasterType` or its `.data`.
Reproducible example:
```python
import geoutils as gu
import numpy as np

s = gu.Raster(gu.datasets.get_path("landsat_B4"))

mask = np.zeros(np.shape(s.data), dtype=bool)
mask[0, 5, 5] = True
s.data[0, 5, 5] = 0
s.data = np.ma.masked_array(s.data, mask)

print(np.min(s))
print(np.ma.min(s))
print(np.min(s.data))
print(np.ma.min(s.data))
```
>0
>0
>13
>13

### LINKED ISSUE #304: Asymmetry of arithmetic operations when broadcasting between 2D and 3D arrays
This actually works now with #302 (the numpy interface is not taking over anymore):
```python
test = np.ones(r.shape) + r
type(test)
Out[15]: geoutils.georaster.raster.Raster
test = r + np.ones(r.data.shape)
type(test)
Out[17]: geoutils.georaster.raster.Raster
```
However, as you can notice, I have to change the shape to `.data.shape` on one side, and not another, otherwise:
```python
ValueError: Both rasters must have the same shape.
```
I'll leave this issue to be closed by #302 and open a new one on this.

_Originally posted by @rhugonnet in https://github.com/GlacioHack/geoutils/issues/256#issuecomment-1240589747_

rhugonnet:
## Summary

This PR adds a **nodata-aware array interface** to work with `Raster` objects and their subclasses. **All of the ~90 universal functions of NumPy are now supported** (see ufuncs: https://numpy.org/doc/stable/reference/ufuncs.html), **and also ~25 NumPy array functions** defined on a per-need basis (such as np.min, np.max, np.mean, np.median, np.std, np.var, np.percentile, np.quantile). All these functions are wrapped in a way that accounts for possible `nodata` in `Raster.data`.

## From `__array_interface__` to `__array_ufunc__` and `__array_function__`

In #210, we added `__array_interface__` as a property to perform operations on `Raster` classes as if on array. However, this interface points to the array of the masked array `.data.data`, and therefore gives inconsistent results depending on `nodata` values.
For example:
```python
np.min(raster.data) # minimum value is around 340, nodata value is -9999
Out[1]: 341.2

np.min(raster) 
Out[2]: -9999
```

This PR tries to expand the addition of @erikmannerfelt to apply to the masked array (Note: there is a `mask` key in the dict of `__array_interface__`, but I didn't understand how this is supposed to work, see https://numpy.org/doc/stable/reference/arrays.interface.html). 
Instead, this PR uses `__array__`, `__array_ufunc__` and `__array_function__` methods of NumPy, that allow to define the comportement of the `Raster` class when functions are applied to it, respectively, `np.array()`, any universal functions `ufunc`, and any array function `function`. See https://numpy.org/doc/stable/user/basics.dispatch.html#basics-dispatch.

## Choosing the behaviour of NumPy functions on `Raster` objects

Important note to start: I did not find a way to define the comportement of `np.ma` functions (which are not covered by `__array_function__` of NumPy, and no equivalent seem to exists in the `ma` module). So this PR only concerns NumPy functions with the format `np.function`.

For defining behaviour, I thought the best option would be to enforce the exact same behaviour as when applying to the masked array `.data`. NumPy function generally account for masks even if not called from `np.ma`. However, there are a couple exceptions that ignore the mask (raising a `UserWarning`), those are:

- `np.median`
- `np.quantile`
- `np.percentile`

And so, because of those three, I thought preferable to make the behaviour of NumPy functions on `Raster` always **nodata-aware**. This means that the behaviour is:

1. All NumPy functions except the 3 listed above are cast to the masked array `Raster.data`. Then, if the function preserves shape (all ufuncs), the output masked array is encapsulated in a `Raster` object.
2. For "median", the NumPy function is redefined into `np.ma.median` to be nodata-aware and cast on the masked array `.data`,
3. For "quantile" and "percentile", no `np.ma` function exists, so the NumPy function is left the same and cast on valid values `.data.data[~.data.mask]`

Seeing how NumPy has made most classic functions outside of `np.ma` account for mask, it could be only a matter of years until they add `np.median`, `np.percentile` and `np.quantile`. Then the behaviour of our array interface would become fully consistent without the need for specific cases.

## For documentation

In the documentation (once we set it up a bit better), we would recommend three options for array calculations, that always remain fully consistent:

1. Applying any `np.function` to the `Raster` object,
2. Applying any `np.ma.function` to the masked array `.data` of the `Raster` object,
3. Applying any `np.nanfunction` to `Raster.get_nanarray()` once implemented (see #277).

## Tests

A specific test class `TestArrayInterface` is added.
It tests all combinations of: 

- `np.function` (all ufuncs and supported array funcs), 
- `dtype` supported by `Raster` (possibly two loops for two inputs),
- `nodata`, either None or the default value of a `dtype` (possibly two nodata for two inputs),

which are applied to `Raster` object with a 5x5 masked array with random values of the `dtype`, and with a mask containing at least one `True` and one `False` value.

For all combinations, the test check that the output of the NumPy function applied to `Raster.data` is fully consistent with the same function applied to `Raster`, at the exception of the special case of "median", "percentile" and "quantile" checked separately. 

Additionally, the PR checks that NumPy `TypeError` or `ValueError` are always raised through application to `Raster` when supposed to, i.e. for a wrong combination of input `dtypes`, and that this error behaviour is fully consistent with the one for the function being applied to `Raster.data`.

The many combinations increases our number of test reported by `pytest` by about 13,500. This is a lot but helps tracing quickly exactly which combination is failing (which occured quite a bit during the drafting phase of the PR). If we feel it's too much, we could move some of these test loops inside the `test_` functions instead of as `@pytest.mark.parametrize` loops.

## Little addition

I have uncommented tests on reflective arithmetic operations with 3D np.ndarrays that use to fail due to #256. I have also added 2D tests and updated the `_overloading_check` function to enable #304.

## Linked issues

Resolves #261
Resolves #256
Resolves #304

### DIFF
--- a/geoutils/georaster/raster.py
+++ b/geoutils/georaster/raster.py
@@ -939,12 +939,36 @@ Must be a Raster, np.ndarray or single number."
 
         return cp
 
-    @property
-    def __array_interface__(self) -> dict[str, Any]:
+    # Not sure yet how this behaves with __array__, __array_ufunc__ and __array_function__ implemented
+    # @property
+    # def __array_interface__(self) -> dict[str, Any]:
+    #     if not self.is_loaded:
+    #         self.load()
+    #
+    #     return self._data.__array_interface__  # type: ignore
+
+    # Set up array interface for our class container
+    # Inspired from https://numpy.org/doc/stable/user/basics.dispatch.html#basics-dispatch
+    def __array__(self):
+
+        return self._data
+
+    def __array_ufunc__(self, ufunc, method, *inputs, **kwargs) -> RasterType:
+
         if not self.is_loaded:
             self.load()
 
-        return self._data.__array_interface__  # type: ignore
+        return  self.__class__({"data": ufunc(self._data, *inputs, **kwargs),
+                               "transform": self.transform,
+                               "crs": self.crs,
+                               "nodata": self.nodata}) # type: ignore
+
+    def __array_function__(self, func, types, args, kwargs):
+
+        if not self.is_loaded:
+            self.load()
+
+        return func(self._data, *kwargs) # type: ignore
 
     # Note the star is needed because of the default argument 'mode' preceding non default arg 'inplace'
     # Then the final overload must be duplicated
--- a/geoutils/georaster/raster.py
+++ b/geoutils/georaster/raster.py
@@ -1065,7 +1065,9 @@ Must be a Raster, np.ndarray or single number."
                     {"data": output[1], "transform": self.transform, "crs": self.crs, "nodata": self.nodata}
                 )
 
-    def __array_function__(self, func: Callable[[np.ndarray, Any], Any], types: tuple[type], args: Any, kwargs: Any) -> Any:
+    def __array_function__(
+        self, func: Callable[[np.ndarray, Any], Any], types: tuple[type], args: Any, kwargs: Any
+    ) -> Any:
         """
         Method to cast NumPy array function directly on a Raster object by applying it to the masked array.
         A limited number of function is supported, listed in raster._HANDLED_FUNCTIONS.
--- a/geoutils/georaster/raster.py
+++ b/geoutils/georaster/raster.py
@@ -1027,42 +1027,37 @@ Must be a Raster, np.ndarray or single number."
             # If the universal function has only one output
             if ufunc.nout == 1:
                 return self.from_array(
-                    {
-                        "data": getattr(ufunc, method)(inputs[0].data, **kwargs),  # type: ignore
-                        "transform": self.transform,
-                        "crs": self.crs,
-                        "nodata": self.nodata,
-                    }
-                )
+                        data=getattr(ufunc, method)(inputs[0].data, **kwargs),
+                        transform=self.transform,
+                        crs=self.crs,
+                        nodata=self.nodata)
 
             # If the universal function has two outputs (Note: no ufunc exists that has three outputs or more)
             else:
-                output = getattr(ufunc, method)(inputs[0].data, **kwargs)  # type: ignore
+                output = getattr(ufunc, method)(inputs[0].data, **kwargs)
                 return self.from_array(
-                    {"data": output[0], "transform": self.transform, "crs": self.crs, "nodata": self.nodata}
+                    data=output[0], transform=self.transform, crs=self.crs, nodata=self.nodata
                 ), self.from_array(
-                    {"data": output[1], "transform": self.transform, "crs": self.crs, "nodata": self.nodata}
+                    data=output[1], transform=self.transform, crs=self.crs, nodata=self.nodata
                 )
 
         # If the universal function takes two inputs (Note: no ufunc exists that has three inputs or more)
         else:
             if ufunc.nout == 1:
                 return self.from_array(
-                    {
-                        "data": getattr(ufunc, method)(inputs[0].data, inputs[1].data, **kwargs),  # type: ignore
-                        "transform": self.transform,
-                        "crs": self.crs,
-                        "nodata": self.nodata,
-                    }
+                    data=getattr(ufunc, method)(inputs[0].data, inputs[1].data, **kwargs),
+                        transform=self.transform,
+                        crs=self.crs,
+                        nodata=self.nodata
                 )
 
             # If the universal function has two outputs (Note: no ufunc exists that has three outputs or more)
             else:
-                output = getattr(ufunc, method)(inputs[0].data, inputs[1].data, **kwargs)  # type: ignore
+                output = getattr(ufunc, method)(inputs[0].data, inputs[1].data, **kwargs)
                 return self.from_array(
-                    {"data": output[0], "transform": self.transform, "crs": self.crs, "nodata": self.nodata}
+                    data=output[0], transform=self.transform, crs=self.crs, nodata=self.nodata
                 ), self.from_array(
-                    {"data": output[1], "transform": self.transform, "crs": self.crs, "nodata": self.nodata}
+                    data=output[1], transform=self.transform, crs=self.crs, nodata=self.nodata
                 )
 
     def __array_function__(
--- a/geoutils/georaster/raster.py
+++ b/geoutils/georaster/raster.py
@@ -1026,7 +1026,7 @@ Must be a Raster, np.ndarray or single number."
         if ufunc.nin == 1:
             # If the universal function has only one output
             if ufunc.nout == 1:
-                return self.__class__(
+                return self.from_array(
                     {
                         "data": getattr(ufunc, method)(inputs[0].data, **kwargs),  # type: ignore
                         "transform": self.transform,
@@ -1038,16 +1038,16 @@ Must be a Raster, np.ndarray or single number."
             # If the universal function has two outputs (Note: no ufunc exists that has three outputs or more)
             else:
                 output = getattr(ufunc, method)(inputs[0].data, **kwargs)  # type: ignore
-                return self.__class__(
+                return self.from_array(
                     {"data": output[0], "transform": self.transform, "crs": self.crs, "nodata": self.nodata}
-                ), self.__class__(
+                ), self.from_array(
                     {"data": output[1], "transform": self.transform, "crs": self.crs, "nodata": self.nodata}
                 )
 
         # If the universal function takes two inputs (Note: no ufunc exists that has three inputs or more)
         else:
             if ufunc.nout == 1:
-                return self.__class__(
+                return self.from_array(
                     {
                         "data": getattr(ufunc, method)(inputs[0].data, inputs[1].data, **kwargs),  # type: ignore
                         "transform": self.transform,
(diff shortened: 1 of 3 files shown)


geoutils/georaster/raster.py
```diff
@@ -559,12 +598,17 @@ def _overloading_check(
         # Case 2 - other is a numpy array
         elif isinstance(other, np.ndarray):
             # Check that both array have the same shape
-            if self.data.shape == other.shape:
+
+            if len(other.shape) == 2:
```
adehecq, line 602: There's a function for getting the dimension of the array: `np.ndim`. 
Of course here is does not matter, but it works also for lists/tuples.

geoutils/georaster/raster.py
```diff
@@ -958,12 +1002,101 @@ def copy(self: RasterType, new_array: np.ndarray | None = None) -> RasterType:
...
+        self.__class__. The cases separate the number of input nin and output nout, to properly feed .data and return
+        Raster objects.
+        See more details in NumPy doc, e.g., https://numpy.org/doc/stable/user/basics.dispatch.html#basics-dispatch.
+        """
+
+        # If the universal function takes only one input
+        if ufunc.nin == 1:
+            # If the universal function has only one output
+            if ufunc.nout == 1:
+                return self.__class__(
```
adehecq, line 1029: Should we not use `self.from_array` rather than the indirect method with class?

geoutils/georaster/raster.py
```diff
@@ -958,12 +1002,101 @@ def copy(self: RasterType, new_array: np.ndarray | None = None) -> RasterType:
...
+                    {
+                        "data": getattr(ufunc, method)(inputs[0].data, **kwargs),  # type: ignore
```
adehecq, line 1031: Can't you define the type of the output (I assume `np.ndarray`) rather than ignoring mypy complain?

geoutils/georaster/raster.py
```diff
@@ -958,12 +1002,101 @@ def copy(self: RasterType, new_array: np.ndarray | None = None) -> RasterType:
...
+            # If the universal function has two outputs (Note: no ufunc exists that has three outputs or more)
+            else:
+                output = getattr(ufunc, method)(inputs[0].data, inputs[1].data, **kwargs)  # type: ignore
+                return self.__class__(
+                    {"data": output[0], "transform": self.transform, "crs": self.crs, "nodata": self.nodata}
+                ), self.__class__(
+                    {"data": output[1], "transform": self.transform, "crs": self.crs, "nodata": self.nodata}
+                )
+
+    def __array_function__(self, func: Callable[[np.ndarray, Any], Any], types: type, args: Any, kwargs: Any) -> Any:
```
adehecq, line 1068: what's the role/meaning of `types` here?

geoutils/georaster/raster.py
```diff
@@ -958,12 +1002,101 @@ def copy(self: RasterType, new_array: np.ndarray | None = None) -> RasterType:
...
+
+        # We now choose the behaviour of array functions
+        # For median, np.median ignores masks of masked array, so we force np.ma.median
+        if func.__name__ in ["median", "nanmedian"]:
+            func = np.ma.median
+            first_arg = args[0].data
+
+        # For percentiles and quantiles, there exist no masked array version, so we compute on the valid data directly
+        elif func.__name__ in ["percentile", "nanpercentile"]:
+            first_arg = args[0].data.data[~args[0].data.mask]
```
adehecq, line 1090: You can use `self.data.compressed()` to get only the valid values (I know, the name is not very straightforward...).

tests/test_georaster.py
```diff
@@ -1913,3 +1926,279 @@ def test_power(self, power: float | int) -> None:
...
+    # --  Third, we define the test data --
+
+    # We create two random array of varying dtype
+    width = height = 5
+    min_val = np.iinfo("int32").min
+    max_val = np.iinfo("int32").max
+    transform = rio.transform.from_bounds(0, 0, 1, 1, width, height)
+    np.random.seed(42)
+    arr1 = np.random.randint(min_val, max_val, (height, width), dtype="int32") + np.random.normal(size=(height, width))
+    arr2 = np.random.randint(min_val, max_val, (height, width), dtype="int32") + np.random.normal(size=(height, width))
```
adehecq, line 2012: This makes me think. We must have written these few lines many times in the tests. If that's the case, maybe we could have a function to generate a random Raster of a given shape and dtype, to make sure we are consistent in all tests.

tests/test_georaster.py
```diff
@@ -1913,3 +1926,279 @@ def test_power(self, power: float | int) -> None:
...
+            # Check if our input dtype is possible on this ufunc, if yes check that outputs are identical
+            if com_dtype in (str(np.dtype(t[0])) for t in ufunc.types):
+
+                # For a single output
+                if ufunc.nout == 1:
+                    assert np.ma.allequal(ufunc(rst.data), ufunc(rst).data)
+
+                # For two outputs
+                elif ufunc.nout == 2:
+                    assert np.ma.allequal(ufunc(rst.data)[0], ufunc(rst)[0].data) and np.ma.allequal(
```
adehecq, line 2061: Note that you are running each function twice. You could run it twice and save the outputs instead.

adehecq (approved):
Apart from a few minor comments, it looks all good. Thanks for fixing this !!
Just of curiosity, did you check how this is done in pandas?

geoutils/georaster/raster.py
rhugonnet: OK. Here for consistency I copied the syntax that was already used in `data.setter`

geoutils/georaster/raster.py
rhugonnet: I don't think so because we don't have the arguments that a user can pass to `from_array` here (like a custom transform, an integer CRS).

So the only difference is the check for the transform at the beginning of `from_array`, which we don't need because the Raster object already exists and we simply copy the transform.

geoutils/georaster/raster.py
rhugonnet: The typing is already defined for `__array_ufunc__` and `__array_function__` in the function definition.

 Mypy still complains about something else specific to those lines, I think it's the indexing of `inputs` or something similar. I couldn't figure it out so I silenced it.

geoutils/georaster/raster.py
rhugonnet: It is passing the types of the input(s) on which the NumPy function was used, to ensure those apply to our class/subclass (potentially needs subclassing). More info in the NumPy doc.

geoutils/georaster/raster.py
rhugonnet: Good point, changed everywhere!

geoutils/georaster/raster.py
rhugonnet: (and I changed the typing as it should have been: `tuple[type]`)

tests/test_georaster.py
rhugonnet: Yes, it's not a CPU issue right now because we always define them outside the loops, so it's defined maybe ~3-4 times. But we could have it as a `pytest.fixture` to reuse it everywhere consistently, and ensuring it is not modified. Although it would make the code a bit harder to read...

rhugonnet: I leave as is and don't open an issue, and we have much more important things to do in the test suite as of now :sweat_smile:. 
A small future improvement for someone motivated.

tests/test_georaster.py
rhugonnet: Fixed at both occurences!

geoutils/georaster/raster.py
adehecq: What do you mean we don't have the arguments for `from_array`? You do pass a CRS and a transform, so I don't see the issue...
My only concern is that calling the Raster class with the dictionary is a trick to get `from_array` to work. There is no 100% guarantee that it will not change in the future, while the function `from_array` should always exist.  So it's safer if we use `from_array` everywhere, because then all we need to update is the `from_array` function.

adehecq: In brief, I think the call to `__class__` should never be used outside of the definition of `from_array`.

adehecq:
> I checked how this is recommended by NumPy (https://numpy.org/doc/stable/user/basics.dispatch.html#basics-dispatch), which is how it's typically done for dask arrays or cupy arrays (pandas does not encapsulate np.ndarray, so it's not exactly the same use cases).

Ok cool, it's nicely explained on the NumPy doc indeed !

geoutils/georaster/raster.py
rhugonnet: Agree, changing to `from_array`.

PR #303, opened 2022-09-07
Add instructions on how to update the conda-forge feedstock

erikmannerfelt:
Hello there!

I just updated the conda-forge version.
This process should be automatic but isn't at the moment, so I thought it best to document how an update is done.
Please let me know if I should change or add something.


adehecq (approved):
Looks great, thanks !!

Issue #306, opened 2022-09-13
Raster.intersection does not raise error if rasters in different projection


adehecq:
Currently, Raster.intersection is not implemented when the input rasters have a different projection and it does not raise an error when that occurs.
https://github.com/GlacioHack/geoutils/blob/96c432c4018a239b8ace1fc7442b0fa75a7b5130/geoutils/georaster/raster.py#L1669

See #272 for a quick fix.

adehecq:
Closed via #307

Issue #305, opened 2022-09-13
Raster.reproject fails if nodata not set and data not loaded


adehecq:
To reproduce the error:
```
import geoutils as gu
img1 = gu.Raster(gu.examples.get_path("everest_landsat_b4"), load_data=False)
img1.reproject(dst_res=10)
```
raises "ValueError: Data are not loaded"

This is because of the test added in PR #289 to check that the default nodata value does not exist in `self.data`. Need to add a special case for when data is not loaded.

adehecq:
Closed via #307

Issue #272, opened 2022-05-18
Include a function to load multiple rasters on overlapping extent


adehecq:
Let's say we want to compare a large raster and a small raster covering only a subsample of the larger raster. If the large raster is too large to fit in memory, or simply it takes long to load, it would be interesting to load the two rasters only over common area. Currently, doing this would require several lines of code. This is not really implemented if the two rasters have different CRS.

Same CRS:
```
r1 = gu.Raster(file1, load_data=False)
r2 = gu.Raster(file2, load_data=False)
intersection = r1.intersection(r2)
r1 = gu.Raster(file1, load_data=intersection)
r2 = gu.Raster(file2, load_data=intersection)
```

Different CRS:
```
r1 = gu.Raster(file1, load_data=False)
r2 = gu.Raster(file2, load_data=False)
intersection = gu.projtools.merge_bounds([r1.bounds, r2.get_bounds_projected(r1.crs)], "intersection")
r1 = gu.Raster(file1, load_data=intersection)
# Need to convert intersection into r2's CRS, which is not straightforward
r2 = gu.Raster(file2, load_data=intersection_r2)
```

Another option is to modify the behavior of Raster.reproject to not work on self.data, but on the dataset directly. This may become difficult after PR #265.

adehecq:
>  Need to convert intersection into r2's CRS, which is not straightforward

Actually, this can be done in one line as in Raster.get_bounds_projected:
`rio.warp.transform_bounds(src_crs, dst_crs, left, bottom, right, top, densify_pts)`

adehecq:
Closed via #307

PR #307, opened 2022-09-13
Fix several issues in Raster

adehecq:
- [x] Solves #305 
- [x] Solves #306 
- [x] Solves #272

In summary, three new functionalities have been implemented:
- `Raster.intersection` handles different CRS
- there is a whole new function `spatial_tools.load_mutliple_rasters` to efficiently load a list of files. The user has the option to crop all rasters to their common extent (default) and to reproject all to a reference grid (default to False). Very useful when wanting to load multiple raster to be compared.
- `projtools.bounds2poly` used to raise a NotImplemetedError when different CRS were used. This is now implemented. However that function has 0 tests so far...


geoutils/georaster/raster.py
```diff
@@ -1284,7 +1284,13 @@ def reproject(
...
+                if not self.is_loaded:
+                    warnings.warn(
+                        f"For reprojection, dst_nodata must be set. Setting default nodata to {dst_nodata}. \
+You may set a different nodata with `dst_nodata`."
+                    )
+
+                elif dst_nodata in self.data:
                     warnings.warn(
                         f"For reprojection, dst_nodata must be set. Default chosen value {dst_nodata} exist in \
 self.data. This may have unexpected consequences. Consider setting a different nodata with \
```
rhugonnet, line 1296: Typo from previous PR in this warning: "exists" and not "exist"

geoutils/georaster/raster.py
```diff
@@ -1663,17 +1669,18 @@ def get_bounds_projected(self, out_crs: CRS, densify_pts_max: int = 5000) -> rio
...
+    def intersection(self, rst: str | Raster, match_ref: bool = True) -> tuple[float, float, float, float]:
         """
         Returns the bounding box of intersection between this image and another.
 
         If the rasters have different projections, the intersection extent is given in self's projection system.
 
         :param rst : path to the second image (or another Raster instance)
-
+        :param match_ref: if set to True, returns the smallest intersection that is a multiple of self's georeferences
         :returns: extent of the intersection between the 2 images \
```
rhugonnet, line 1684: Cannot understand "that is a multiple of self's georeferences" by itself. Is the "multiple" linked to the Raster's resolution?

geoutils/projtools.py
```diff
@@ -97,6 +98,36 @@ def merge_bounds(
     return new_bounds
 
 
+def align_bounds(
+    ref_transform: rio.transform.Affine,
+    src_bounds: rio.coords.BoundingBox | tuple[float, float, float, float],
+) -> tuple[float, float, float, float]:
+    """
+    Aligns the bounds in src_bounds so that it matches the georeferences in ref_transform
+    i.e. the distance between the upper-left pixels or ref and src is a multiple of resolution and
```
rhugonnet, line 107: pixels "of" ?

geoutils/spatial_tools.py
```diff
@@ -88,6 +88,69 @@ def get_valid_extent(array: np.ndarray | np.ma.masked_array) -> tuple[int, ...]:
...
+        return output_rst
+
+    for rst in output_rst:
+        # Calculate bounds in rst's CRS
+        new_bounds = rio.warp.transform_bounds(
+            ref_crs, rst.crs, intersection[0], intersection[1], intersection[2], intersection[3], densify_pts=5000
+        )
+        # Ensure bounds align with the original ones, to avoid resampling at this stage
+        new_bounds = gu.projtools.align_bounds(rst.transform, new_bounds)
+        rst.crop(new_bounds, mode="match_pixel")
```
rhugonnet, line 142: So crop is memory efficient, doesn't load all the entire Raster?

geoutils/spatial_tools.py
```diff
@@ -88,6 +88,69 @@ def get_valid_extent(array: np.ndarray | np.ma.masked_array) -> tuple[int, ...]:
...
+    # Second get the intersection of all raster bounds and crop rasters accordingly
+    intersection = gu.projtools.merge_bounds(bounds, "intersection")
+    if intersection == ():
+        warnings.warn("No intersection found, returning unloaded rasters")
+        return output_rst
+
+    for rst in output_rst:
+        # Calculate bounds in rst's CRS
+        new_bounds = rio.warp.transform_bounds(
+            ref_crs, rst.crs, intersection[0], intersection[1], intersection[2], intersection[3], densify_pts=5000
```
rhugonnet, line 138: Why 5000 on `densify_pts`?

geoutils/spatial_tools.py
rhugonnet: If the cropping exactly matches the extent, won't it negatively affect the reprojection later? Like, the 1/2 pixels on the edge might end up as `nodata` because of the resampling, while they wouldn't have been if using the entire original Raster?

tests/test_georaster.py
```diff
@@ -942,20 +942,128 @@ def test_reproject(self, example: str) -> None:
...
+            r.transform.d,
+            r.transform.e,
+            r.transform.f,
+        )
+        r_nonoverlap = gu.Raster.from_array(r.data, crs=r.crs, transform=transform_shifted, nodata=r.nodata)
 
-        assert inters == (left, bottom, right, top)
+        with pytest.warns(UserWarning, match="Intersection is void"):
+            intersection = r.intersection(r_nonoverlap)
+            assert intersection == (0.0, 0.0, 0.0, 0.0)
```
rhugonnet, line 1066: Great series of tests! :slightly_smiling_face:

geoutils/georaster/raster.py
adehecq: Done

geoutils/projtools.py
adehecq: Done

geoutils/spatial_tools.py
adehecq: That's the default maximum used in `Raster.get_bounds_projected`. From early tests I found that the default was not enough, and the final bounds was not encompassing the entire image once warped.
I will try to clarify.

adehecq: It is true that rasterio's documentation says it worsen for large numbers. But I don't understand why...
Because a rectangle will not be a rectangle anymore in a different projection, it is not sufficient to reproject the position of the corners. Instead one needs to add points on the edge in-between the corners. Ideally, one would have one vertex for each edge pixel (which is more or less the default in `get_bounds_projected` with a max of 5000 points). But here I don't really know the width/height of the raster over that area.

geoutils/spatial_tools.py
adehecq: Yes, crop only loads the data within the corresponding window. That's why it's useful to crop the raster, especially if one is much larger.
Actually, reproject does not need to have the data loaded anymore, so if followed by reproject, one does not need to crop and load the data. I will try that instead. If the data is not loaded, it uses the raster's band instead, so rasterio handles all this for us.

geoutils/georaster/raster.py
adehecq: Done

geoutils/spatial_tools.py
adehecq: Solved in [df1bad1](https://github.com/GlacioHack/geoutils/pull/307/commits/df1bad1df509dee2f8f2cb4f31a7cf2af34d157b)

geoutils/spatial_tools.py
adehecq: I've clarified the choice of a larger value, but I kept the (arbitrary) value of 5000 for now.

tests/test_georaster.py
adehecq: 😄

rhugonnet (approved):
Looks good! :smile:

rhugonnet:
This would fit better in an `io.py` module than `spatial_tools.py`, but we need to rethink the structure of functions in `spatial_tools.py` anyway, which is a bit where we dump all the stuff we don't know where to put otherwise.

adehecq:
> This would fit better in an `io.py` module than `spatial_tools.py`, but we need to rethink the structure of functions in `spatial_tools.py` anyway, 

If there is no issue about it, I will open a new one 😅 

> which is a bit where we dump all the stuff we don't know where to put otherwise.

Well, there is a logic, which is indicated in the module's docstring: "Basic operations to be run on 2D image arrays". But the name could be improved, and we need to define a list of functionalities we want there.

PR #308, opened 2022-09-14
Add `get_nanarray` method

### LINKED ISSUE #277: Add a `get_nan_array` method to the Raster class.

rhugonnet:
As discussed many times, it is much more familiar for many people to work with classical arrays rather than masked arrays. Right now, we have to use the `gu.spatial_tools.get_array_and_mask` method to convert between the two, which is a bit unintuitive. Additionally, the `get_array_and_mask` method is made to work with any type of input to homogenize the behaviour of some functions. This is not needed when working with a `Raster` object that always has the same `.data`.

In this PR, we add the `Raster` class method `get_nanarray` to return a NaN array, with argument `return_mask` (defaults as `False`) to possibly the original mask of valid data as well. 

The behaviour is the following:

- The NaN array corresponds exactly to `Raster.data.filled(fill_value=np.nan).squeeze()`, i.e. the data with masked values replaced by NaNs, and squeezed (2D array if there was only a single band);
- The mask corresponds exactly to `np.ma.getmaskarray(Raster.data).squeeze()`, i.e. the squeezed mask if one existed for the masked array `.data`, or an array filled with `False` the same shape as `.data` otherwise.

Resolves #277

### DIFF
--- a/geoutils/georaster/raster.py
+++ b/geoutils/georaster/raster.py
@@ -1002,19 +1002,37 @@ Must be a Raster, np.ndarray or single number."
 
         return cp
 
+    @overload
+    def get_nanarray(self, return_mask: Literal[False]) -> np.ndarray:
+        ...
+
+    @overload
+    def get_nanarray(self, return_mask: Literal[True]) -> tuple[np.ndarray, np.ndarray]:
+        ...
+
+    @overload
+    def get_nanarray(self, return_mask: bool = False) -> np.ndarray | tuple[np.ndarray, np.ndarray]:
+        ...
+
     def get_nanarray(self, return_mask: bool = False) -> np.ndarray | tuple[np.ndarray, np.ndarray]:
         """
-        Method to return a squeezed array filled with NaN and the associated mask from the masked array in .data
+        Method to return the squeezed masked array filled with NaNs and associated squeezed mask, both as a copy.
 
         :param return_mask: Whether to return the mask of valid data
 
         :returns Array with masked data as NaNs, (Optional) Mask of valid data
         """
 
+        # Get the array with masked value fill with NaNs
         nanarray = self.data.filled(fill_value=np.nan).squeeze()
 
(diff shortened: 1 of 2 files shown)


geoutils/georaster/raster.py
```diff
@@ -1002,6 +1002,22 @@ def copy(self: RasterType, new_array: np.ndarray | None = None) -> RasterType:
 
         return cp
 
+    def get_nanarray(self, return_mask: bool = False) -> np.ndarray | tuple[np.ndarray, np.ndarray]:
```
adehecq, line 1005: SO that mypy does not annoy us in theh future, it would be good to write the overload now.

adehecq (approved):
Great !
Just one question, do you know if the returned array will be a coy every time or if in some instances it will be a view? This is important to know as one may or may not modify the original raster.

rhugonnet:
Good point, it seems this depends if there are unvalid values to mask in the input masked array: https://github.com/numpy/numpy/blob/v1.23.0/numpy/ma/core.py#L3776

I will add tests to check the behaviour!

adehecq:
> Good point, it seems this depends if there are unvalid values to mask in the input masked array: https://github.com/numpy/numpy/blob/v1.23.0/numpy/ma/core.py#L3776
> 
> I will add tests to check the behaviour!

In that case, we should just make sure that it always returns a copy (the safest option anyway). Adding a .copy() after .filled() should do it?

rhugonnet:
I added the `copy()` just for when the array has no invalid masked by checking with `np.ma.is_masked()`. This way the copying does not happen twice when invalid values exists (could be bad for performance).

PR #309, opened 2022-09-14
Modify `set_nodata` behaviour, improve related warnings and tests

### LINKED ISSUE #286: `set_ndv` does not update nodatas in array
During `set_ndv`, the array is not updated with previous nodata.

```python
r = gu.Raster(landsat_b4_path)
r.data.ravel()[0:100]=255
r.set_ndv(255)
np.max(r.data)
Out[22]: 254
np.max(r.data.data)
Out[23]: 255
r.set_ndv(254, update_array=True)
np.max(r.data.data)
Out[25]: 255
```
All the 255 no data values should have been updated to 254 by the `set_ndv(..., update_array=True)`, which is also the default parameter.

rhugonnet:
This PR homogenizes the getter and setter functionality related to `nodata` and adds more exhaustive tests.

In details, this PR:

- Renames `set_ndv()` into `set_nodata()`, and all other occurences of "ndv" to "nodata", in order to match the `nodata` naming of `rasterio` and avoid the confusion of having "ndv" and "nodata" standing for the same thing,
- Corrects the behaviour of `update_array` in `set_nodata()`,
- Adds the `update_mask` argument to `set_nodata()`,
- Adds a `nodata.setter` method that calls `set_nodata()` with default parameters,
- Improves or adds tests for `set_nodata` and `nodata.setter`,
- Moves the nodata-related functions/class methods together in the file (among others, mypy request the getter/setter method to follow each other).

The old behaviour was:

- The argument `update_array` was `False` by default,
- The argument `update_array` would modify `.data.mask` by unmasking old nodata and masking new nodatas (i.e. a rare case where the nodata value was wrongly defined and needed to be changed while ignoring old nodatas).

The new behaviour is:
- Both arguments `update_array` and `update_mask` are `True` by default,
- The argument `update_array` updates the array `.data.data` by replacing old nodata values into new nodata values.
- The argument `update_mask` updates the mask `.data.mask` by unmasking old nodata and masking new nodatas (including the ones possibly updated by `update_array`, i.e. essentially keeping the mask on old nodata that have been updated, which makes the big difference with the previous behaviour).
- A `UserWarning` is raised if the new nodata value already exist as a valid value of the array (as it will be automatically masked).

This behaviour is also explained in the function description, with typical use-cases.

Resolves #286

### DIFF
--- a/geoutils/georaster/raster.py
+++ b/geoutils/georaster/raster.py
@@ -567,8 +567,8 @@ np.ndarray or number and correct dtype, the compatible nodata value.
         # See https://docs.python.org/3/reference/datamodel.html#emulating-numeric-types
         if not isinstance(other, (Raster, np.ndarray, Number)):
             raise NotImplementedError(
-                f"Operation between an object of type {type(other)} and a Raster impossible. \
-Must be a Raster, np.ndarray or single number."
+                f"Operation between an object of type {type(other)} and a Raster impossible. Must be a Raster, "
+                f"np.ndarray or single number."
             )
 
         # Get self's dtype and nodata
@@ -910,21 +910,21 @@ Must be a Raster, np.ndarray or single number."
                     if np.count_nonzero(index_new_nodatas) > 0:
                         if update_array and update_mask:
                             warnings.warn(
-                                message="New nodata value found in the data array. Those will be masked, and the old nodata "
-                                        "cells will now take the same value. Use set_nodata() with update_array=False and/or "
-                                        "update_mask=False to change this behaviour",
+                                message="New nodata value found in the data array. Those will be masked, and the old "
+                                "nodata cells will now take the same value. Use set_nodata() with update_array=False "
+                                "and/or update_mask=False to change this behaviour",
                                 category=UserWarning,
                             )
                         elif update_array:
                             warnings.warn(
-                                "New nodata value found in the data array. The old nodata cells will now take the same value. "
-                                "Use set_nodata() with update_array=False to change this behaviour",
+                                "New nodata value found in the data array. The old nodata cells will now take the same "
+                                "value. Use set_nodata() with update_array=False to change this behaviour",
                                 category=UserWarning,
                             )
                         elif update_mask:
                             warnings.warn(
-                                "New nodata value found in the data array. Those will be masked. Use set_nodata() with "
-                                "update_mask=False to change this behaviour",
+                                "New nodata value found in the data array. Those will be masked. Use set_nodata() "
+                                "with update_mask=False to change this behaviour",
                                 category=UserWarning,
                             )
 
@@ -943,7 +943,7 @@ Must be a Raster, np.ndarray or single number."
 
                     # Only update mask with new nodata if it is defined
                     if nodata is not None:
-                        # Masking like this works from the masked array directly, whether a mask previously existed or not
+                        # Masking like this works from the masked array directly, whether a mask exists or not
                         imgdata[i, index_new_nodatas] = np.ma.masked
 
             # Update the data
@@ -1003,10 +1003,8 @@ Must be a Raster, np.ndarray or single number."
         # Check that new_data has the right type
         if str(new_data.dtype) != dtype:
             raise ValueError(
-                "New data must be of the same type as existing\
- data: {}. Use copy() to set a new array with different dtype, or astype() to change type.".format(
-                    dtype
-                )
+                "New data must be of the same type as existing data: {}. Use copy() to set a new array with "
+                "different dtype, or astype() to change type.".format(dtype)
             )
 
         if new_data.shape[1:] != orig_shape:
@@ -1443,15 +1441,15 @@ Must be a Raster, np.ndarray or single number."
                 # TODO: for uint8, if all values are used, apply rio.warp to mask to identify invalid values
                 if not self.is_loaded:
                     warnings.warn(
-                        f"For reprojection, dst_nodata must be set. Setting default nodata to {dst_nodata}. \
-You may set a different nodata with `dst_nodata`."
+                        f"For reprojection, dst_nodata must be set. Setting default nodata to {dst_nodata}. You may "
+                        f"set a different nodata with `dst_nodata`."
                     )
 
                 elif dst_nodata in self.data:
                     warnings.warn(
-                        f"For reprojection, dst_nodata must be set. Default chosen value {dst_nodata} exists in \
-self.data. This may have unexpected consequences. Consider setting a different nodata with \
-self.set_nodata()."
+                        f"For reprojection, dst_nodata must be set. Default chosen value {dst_nodata} exists in "
+                        f"self.data. This may have unexpected consequences. Consider setting a different nodata with "
+                        f"self.set_nodata()."
                     )
 
         from geoutils.misc import resampling_method_from_str
@@ -1722,10 +1720,7 @@ self.set_nodata()."
 
                 # Warning: this will overwrite the transform
                 if dst.transform != rio.transform.Affine(1, 0, 0, 0, 1, 0):
-                    warnings.warn(
-                        "A geotransform previously set is going \
-to be cleared due to the setting of GCPs."
-                    )
+                    warnings.warn("A geotransform previously set is going to be cleared due to the setting of GCPs.")
 
                 dst.gcps = (rio_gcps, gcps_crs)
 
--- a/geoutils/georaster/raster.py
+++ b/geoutils/georaster/raster.py
@@ -372,7 +372,7 @@ class Raster:
 
             # Set nodata
             if nodata is not None:
-                self.set_ndv(nodata)
+                self.set_nodata(nodata)
 
         # Provide a catch in case trying to load from data array
         elif isinstance(filename_or_dataset, np.ndarray):
@@ -1288,7 +1288,7 @@ Must be a Raster, np.ndarray or single number."
                     warnings.warn(
                         f"For reprojection, dst_nodata must be set. Default chosen value {dst_nodata} exist in \
 self.data. This may have unexpected consequences. Consider setting a different nodata with \
-self.set_ndv."
+self.set_nodata()."
                     )
 
         from geoutils.misc import resampling_method_from_str
@@ -1390,7 +1390,7 @@ self.set_ndv."
             elif dst_nodata is not None:
                 if not silent:
                     warnings.warn(
-                        "Only nodata is different, consider using the 'set_ndv()' method instead'\
+                        "Only nodata is different, consider using the 'set_nodata()' method instead'\
                     ' -> return self (not a copy!)"
                     )
                 return self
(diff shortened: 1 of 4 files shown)


geoutils/georaster/raster.py
```diff
@@ -817,14 +813,142 @@ def is_modified(self) -> bool:
...
-        Get nodata value
+        Get nodata value.
 
-        :returns: nodata value
+        :returns: Nodata value
         """
         return self._nodata
 
+    @nodata.setter
+    def nodata(self, new_nodata: int | float | list[int] | list[float] | None) -> None:
```
adehecq, line 825: Is it actually possible to see that docstring?? By calling `help(self.nodata)`? That would be very unusual.
Any way, my thought is that this message is going to be missed, so to avoid problems, I would just raise an error stating that `self.nodata` can only be updated with `self.update_nodata()`. This way, the user will really get the chance to understand what is being done.

geoutils/georaster/raster.py
```diff
@@ -817,14 +813,142 @@ def is_modified(self) -> bool:
...
+    def set_nodata(
+        self, nodata: int | float | list[int] | list[float] | None, update_array: bool = True, update_mask: bool = True
+    ) -> None:
+        """
+        Set a new nodata value for each band. This updates the old nodata into a new nodata value in the metadata,
+        replaces the nodata values in the data of the masked array, and updates the mask of the masked array.
+
+        Careful! If the new nodata value already exists in the array, the related grid cells will be masked by default.
+
+        If the nodata value was not defined in the raster, run this function with a new nodata value corresponding to
```
adehecq, line 850: At least this is well documented, so one knows what needs to be done. And actually, in many cases, all these options will be equivalent, if the previous and new nodata were not set to used values.

geoutils/georaster/raster.py
```diff
@@ -817,14 +813,142 @@ def is_modified(self) -> bool:
...
+        :param nodata: Nodata values
+        :param update_array: Update the old nodata values into new nodata values in the data array
+        :param update_mask: Update the old mask by unmasking old nodata and masking new nodata (if array is updated,
+            old nodata are changed to new nodata and thus stay masked)
+        """
+        if nodata is not None and not isinstance(nodata, (list, int, float, np.integer, np.floating)):
+            raise ValueError("Type of nodata not understood, must be list or float or int")
+
+        elif (isinstance(nodata, (int, float, np.integer, np.floating))) and self.count > 1:
+            print("Several raster band: using nodata value for all bands")
```
adehecq, line 872: I would add a verbose option (to True by default) for all the print statements, so one can ignore them if needed.

geoutils/georaster/raster.py
```diff
@@ -817,14 +813,142 @@ def is_modified(self) -> bool:
...
+            nodata = [nodata] * self.count
+
+        elif isinstance(nodata, list) and self.count == 1:
+            print("Only one raster band: using first nodata value provided")
+            nodata = list(nodata)[0]
+
+        elif nodata is None:
+            self._nodata = None
+            return
```
adehecq, line 881: Should it not continue? For example if one wants to unmask previously masked values?

geoutils/georaster/raster.py
```diff
@@ -817,14 +813,142 @@ def is_modified(self) -> bool:
...
+                index_old_nodatas = imgdata.data[i, :, :] == self.nodata
+
+                # Get the index of new nodatas
+                index_new_nodatas = imgdata.data[i, :, :] == new_nodata
+
+                if np.count_nonzero(index_new_nodatas) > 0:
+                    if update_array and update_mask:
+                        warnings.warn(
+                            message="New nodata value already found in the data array, the corresponding grid cells "
+                            "will be indistinguishable from that updated from the old nodata value, and will "
```
adehecq, line 914: I found it a bit difficult to read. I would make it more consistent with the last one and say "the corresponding grid cells will be masked and the values updated to match the new nodata."
Same for the second warning.

tests/test_georaster.py
```diff
@@ -907,10 +909,17 @@ def test_reproject(self, example: str) -> None:
...
 
         # If a nodata is set, make sure it is preserved
-        r_ndv = r.copy()
-        r_ndv.set_ndv(255)
-        r3 = r_ndv.reproject(r2)
-        assert r_ndv.nodata == r3.nodata
+        r_nodata = r.copy()
+
+        # Exception with the Landsat example
+        if np.count_nonzero(r_nodata.data.data == 255) > 0:
```
adehecq, line 915: Why is this addition needed?
I'm curious, is the 255 value important here or not? Now I realized that 0 is a more appropriate nodata value, when one needs to be set, as it is not used.

tests/test_georaster.py
```diff
@@ -1220,72 +1229,219 @@ def test_value_at_coords(self) -> None:
...
+        # The rest of the array and mask should be unchanged
+        assert np.array_equal(r.data.data[~index_old_nodata], r_copy.data.data[~index_old_nodata])
+        # We call np.ma.getmaskarray to always compare a full boolean array
+        assert np.array_equal(np.ma.getmaskarray(r.data), np.ma.getmaskarray(r_copy.data))
+
+        # Then, we repeat for a nodata that already existed, we artificially introduce the value on an unmasked pixel
+        r = r_copy.copy()
+        mask_pixel_artificially_set = np.zeros(np.shape(r.data), dtype=bool)
+        mask_pixel_artificially_set[0, 0, 0] = True
+        r.data.data[mask_pixel_artificially_set] = new_nodata
```
adehecq, line 1271: Can't you just do `r.data.data[0, 0, 0] = new_nodata` directly?

adehecq: Ok, I see you need that for later.

tests/test_georaster.py
```diff
@@ -1220,72 +1229,219 @@ def test_value_at_coords(self) -> None:
...
+        if old_nodata is not None:
+            index_old_nodata = r_copy.data.data == old_nodata
+            assert all(r.data.data[index_old_nodata] == new_nodata)
+        else:
+            index_old_nodata = np.zeros(np.shape(r.data.data), dtype=bool)
+
+        # The rest of the array is similarly unchanged
+        index_unchanged = np.logical_and(~index_old_nodata, ~mask_pixel_artificially_set)
+        assert np.array_equal(r.data.data[index_unchanged], r_copy.data.data[index_unchanged])
+        # But, this time, the mask is only unchanged outside of the pixel artificially has changed
```
adehecq, line 1302: you mean "artificially changed"?

tests/test_georaster.py
```diff
@@ -1220,72 +1229,219 @@ def test_value_at_coords(self) -> None:
...
+
+        # -- Fifth, let's check that errors are raised when they should --
+
+        # A ValueError if input nodata is neither a list, tuple, integer, floating
+        with pytest.raises(ValueError, match="Type of nodata not understood, must be list or float or int"):
+            r.set_nodata(nodata="this_should_not_work")  # type: ignore
+
+        # A ValueError if nodata value is incompatible with dtype
+        expected_message = r"nodata value .* incompatible with self.dtype .*"
+        if "int" in r.dtypes[0]:
```
adehecq, line 1402: This will only be tested in one case (uint8). Maybe you could use the min value of the float128 dtype (which is very unlikely to be used in our sample data!).

adehecq:
It's great to have everything a bit more consistent now ! Thanks ! 👏 
I guess this will call for some changes in xdem...

>     * Renames `set_ndv()` into `set_nodata()`, and all other occurences of "ndv" to "nodata", in order to match the `nodata` naming of `rasterio` and avoid the confusion of having "ndv" and "nodata" standing for the same thing,

A good way to increase your git blame score 😝 

>     * Adds a `nodata.setter` method that calls `set_nodata()` with default parameters,

See my comment, I would simply raise an error and tell the user to use set_nodata, as it will force the use to think about the consequences.

geoutils/georaster/raster.py
rhugonnet: I think this is already taken care of:

- The `nodata.setter` always performs default behaviour, which is what we need in 95% of cases. 
- For the other 5%, a UserWarning is raised by `set_nodata` who detects something is unusual, and the warning explicity states what will happen and to use `set_nodata` to avoid the default setting.

This is sufficient, no?

rhugonnet: I made the docstring here for us mostly, and API documentation :smile:

geoutils/georaster/raster.py
rhugonnet: I'll remove the print statement that are from the old function (we know `rasterio` does not support a list of nodata anyway? not sure it's an important feature)

geoutils/georaster/raster.py
rhugonnet: Actually that statement doesn't work: the values of these grid cell will not be updated. They simply will have the same value as the other cells updated from old nodata to new nodata.

I'll change the statement to this which maybe sounds easier to grasp:
"New nodata value found in the data array. Those will be masked, and the old nodata cells will now take the same value. Use set_nodata() with update_array=False and/or update_mask=False to change this behaviour`

tests/test_georaster.py
rhugonnet: Corrected

geoutils/georaster/raster.py
rhugonnet: I made it continue to do exactly that.

rhugonnet:
Actually, I already checked, and there's not a single function using this in xDEM! :partying_face: Except what I just added to remove the floating nodata in the `ddem` example.

rhugonnet:
But `data` and `nodata` are our two big properties of GeoUtils, and intertwined one with the other, so I'm happy if we finally get them consistent with most used behaviour, and with descriptions and options for other behaviours! :slightly_smiling_face:

tests/test_georaster.py
rhugonnet: Here I was just making the existing test work now that there is a `UserWarning` raised when the nodata value exists, I didn't pick `255`.
I'll change to `0` to avoid the warning, I agree it's better.

geoutils/georaster/raster.py
rhugonnet: And added specific test with `None`

geoutils/georaster/raster.py
adehecq: Alright, let's try like this.

geoutils/georaster/raster.py
adehecq: 👍

tests/test_georaster.py
```diff
@@ -1403,6 +1397,33 @@ def test_set_nodata(self, example: str) -> None:
             with pytest.raises(ValueError, match=expected_message):
                 # Feed a floating numeric to an integer type
                 r.set_nodata(0.5)
+        elif "float" in r.dtypes[0]:
```
adehecq, line 1400: Ahah ok, my intention was to use only a single test for both, as the second test also works for uint8 😉

adehecq (approved):
All good, except for a minor comment to one test

tests/test_georaster.py
rhugonnet: Ah yes. Well I guess the test is more precise the current way, as it checks that `integer` types fail with a `nodata` value that would work for `float` types, so I think we can leave as is :)

PR #313, opened 2022-09-16
Fix `__eq__`, make arithmetic operations fully consistent with masked arrays, deprecate `gu.misc.array_equal` and bug fixes

### LINKED ISSUE #290: Deprecate `gu.misc.array_equal` for `np.array_equal` and enforce Numpy>1.19
The `equal_nan` functionality of `np.array_equal` was introduced in NumPy 1.19.0, compatible with Python 3.6.
It was relevant to have this for backward compatibility a year ago when it was relatively new (see #227), but now we only use Python >3.7 (soon >3.8) anyway, so NumPy will always be >1.19.
We can deprecate `gu.misc.array_equal` and switch fully to `np.array_equal` for simplicity.

_Originally posted by @rhugonnet in https://github.com/GlacioHack/geoutils/pull/289#discussion_r958390156_

### LINKED ISSUE #299: Strange behaviour of `array_interface` on mask and nodata setting
Example below: the nodata has been assigned to 63?! 
```python
r.data
Out[8]: 
masked_array(
  data=[[[255, 255, 255, ..., 255, 255, 255],
         [255, 255, 255, ..., 255, 255, 255],
         [255, 255, 255, ..., 255, 255, 255],
         ...,
         [ 74,  76,  79, ..., 121, 119, 141],
         [ 75,  83,  70, ..., 112, 130, 150],
         [ 64,  86,  68, ..., 124, 131, 130]]],
  mask=False,
  fill_value=999999,
  dtype=uint8)
r.data += 5
r.data
Out[10]: 
masked_array(
  data=[[[  4,   4,   4, ...,   4,   4,   4],
         [  4,   4,   4, ...,   4,   4,   4],
         [  4,   4,   4, ...,   4,   4,   4],
         ...,
         [ 79,  81,  84, ..., 126, 124, 146],
         [ 80,  88,  75, ..., 117, 135, 155],
         [ 69,  91,  73, ..., 129, 136, 135]]],
  mask=False,
  fill_value=63,
  dtype=uint8)
np.count_nonzero(r.data==63)
Out[11]: 3441
```
Looking into it in more details...

### LINKED ISSUE #312: Raster.reproject fails if res is numpy number
To reproduce:
```
import geoutils as gu
import numpy as np
rst = gu.Raster(gu.examples.get_path("everest_landsat_b4"))
rst.reproject(dst_res=np.float64(20))
```
Fails with

> ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

The issue occurs at [this line](https://github.com/GlacioHack/geoutils/blob/8c21c3b44a513356f6ebb975418abccf6ec931ac/geoutils/georaster/raster.py#L1418).
The origin is that :
- `30 == (20, 20)` returns a bool (`False`)
- `np.float64(30) == (20, 20)` returns an array (`array([False, False])`)

An easy fix would probably be to use `np.all()` in the condition.

rhugonnet:
## Summary

This PR modifies the behaviour of `__eq__` to have a rigorous equality check between Raster objects.

As a result, a certain number of test failed linked to arithmetic operations, and thus a correction was made to `_overloading_check` and `data.setter` to avoid inconsistencies that were created in the masked arrays. Then, tests relying on `__eq__` were updated to be more robust. 

Additionally, the custom function `gu.misc.array_equal` initially introduced for back-compatibility and for which the behaviour was not clearly defined to all (it only checked valid data, but did not compare unmasked `data`, the `mask` themselves, or the `dtypes` as some of us expected) is now deprecated as similar functionalities exist in NumPy: `np.array_equal(equal_nan=True)`, or `np.ma.allequal`, or in our new `__eq__`.

Finally, a small reprojection bug was fixed, a couple tests that did not work after the `__eq__` update, and the type linting of `Raster.astype` that was missing a default parameter for one of its `Literal`.

_Note: as the `array_equal` function lives in the same module as `deprecate`, I could not use the `deprecate` decorator (could only made it work from another module), so the function is removed entirely. I think this is fine as it is a "developer" function we use exclusively for tests, and not a "user" function._

## Details

In details, this PR:

- [x] Makes `__eq__` compute an exact Raster equality, now considering masked values in `data.data`, the mask `data.mask`, the fill_value `data.fill_value` and the dtype `data.dtype` (called by to get the Raster dtype, so always matches);
- [x] Fixes inconsistent arithmetic operation behaviour due to a forcing of `dtype` or an encapsulation by `np.ma.masked_array` in `_overloading_check` and `data.setter` to ensure the masked array operations with `Raster` always behave exactly like they would for arithmetic operation of masked arrays (the issue #299 was triggered by passing a masked array as `data=`, i.e. first argument, into `np.ma.masked_array`. This happened in both `data.setter` and `_overloading_check` in some cases);
- [x] Fixes behaviour of `__array_ufunc__` (which takes over reverse arithmetic functions when array is passed first) because of exceptions in a few arithmetic functions (found a NumPy inconsistency!),
- [x] Improves tests throughout `test_georaster.py` and `test_satimg.py` when checking raster or array equality: use `__eq__` when relevant instead of array comparison, use double `np.array_equal` when both raw data and mask should be the same, or  use`np.ma.allequal` when valid data should be the same,
- [x] Removes usage of `gu.misc.array_equal` that is now equivalent to `np.ma.allequal` but no needed for backwards compatibility (see #290).
- [x] Fixes reprojection issue and add tests (see #312).

Resolves #312
Resolves #299
Resolves #290

### DIFF
--- a/geoutils/georaster/raster.py
+++ b/geoutils/georaster/raster.py
@@ -1193,10 +1193,11 @@ np.ndarray or number and correct dtype, the compatible nodata value.
         # In addition to running ufuncs, this function takes over arithmetic operations (__add__, __multiply__, etc...)
         # when the first input provided is a NumPy array and second input a Raster.
 
-        # The Raster ufuncs behave exactly as arithmetic operations (+, *, .) of NumPy that call np.ma  instead of np
-        # when available, which sometimes returns a full boolean mask even when there is no invalid value (true_divide
-        # and floor_divide). We find one exception, however, for modulo: np.ma.remainder is not called but np.remainder
-        # instead (an inconsistency in NumPy?!), so we mirror it below:
+        # The Raster ufuncs behave exactly as arithmetic operations (+, *, .) of NumPy masked array (call np.ma instead
+        # of np when available). There is an inconsistency when calling np.ma: operations return a full boolean mask
+        # even when there is no invalid value (true_divide and floor_divide).
+        # We find one exception, however, for modulo: np.ma.remainder is not called but np.remainder instead one the
+        # masked array is the second input (an inconsistency in NumPy!), so we mirror this exception below:
         if 'remainder' in ufunc.__name__:
             final_ufunc = getattr(ufunc, method)
         else:
--- a/geoutils/georaster/raster.py
+++ b/geoutils/georaster/raster.py
@@ -1538,7 +1538,7 @@ np.ndarray or number and correct dtype, the compatible nodata value.
                 (dst_transform == self.transform) or (dst_transform is None),
                 (dst_crs == self.crs) or (dst_crs is None),
                 (dst_size == self.shape[::-1]) or (dst_size is None),
-                (dst_res == self.res) or (dst_res == self.res[0] == self.res[1]) or (dst_res is None),
+                np.all(dst_res == self.res) or (dst_res == self.res[0] == self.res[1]) or (dst_res is None),
             ]
         ):
             if (dst_nodata == self.nodata) or (dst_nodata is None):
--- a/geoutils/georaster/raster.py
+++ b/geoutils/georaster/raster.py
@@ -761,7 +761,7 @@ np.ndarray or number and correct dtype, the compatible nodata value.
         return out_rst
 
     @overload
-    def astype(self, dtype: DTypeLike, inplace: Literal[False]) -> Raster:
+    def astype(self, dtype: DTypeLike, inplace: Literal[False] = False) -> Raster:
         ...
 
     @overload
@@ -1027,16 +1027,16 @@ np.ndarray or number and correct dtype, the compatible nodata value.
         # 1/ If the new data is not masked (either classic array or masked array with no mask, hence the use of
         # as array) and contains non-finite values such as NaNs, define a mask
         if not np.ma.is_masked(new_data) and np.count_nonzero(~np.isfinite(new_data)) > 0:
-            self._data = np.ma.masked_array(data=np.asarray(new_data),
-                                            mask=~np.isfinite(new_data.data),
-                                            fill_value=self.nodata)
+            self._data = np.ma.masked_array(
+                data=np.asarray(new_data), mask=~np.isfinite(new_data.data), fill_value=self.nodata
+            )
 
         # 2/ If the new data is masked but some non-finite values aren't masked, add them to the mask
         elif np.ma.is_masked(new_data) and np.count_nonzero(~np.isfinite(new_data.data[~new_data.mask])) > 0:
             self._data = np.ma.masked_array(
                 data=new_data.data,
                 mask=np.logical_or(~np.isfinite(new_data.data), new_data.mask),
-                fill_value=self.nodata
+                fill_value=self.nodata,
             )
 
         # 3/ If the new data is a Masked Array, we pass data.data and data.mask independently (passing directly the
@@ -1047,7 +1047,6 @@ np.ndarray or number and correct dtype, the compatible nodata value.
         # 4/ If the new data is classic ndarray
         else:
             self._data = np.ma.masked_array(data=new_data, fill_value=self.nodata)
-
 
     def set_mask(self, mask: np.ndarray) -> None:
         """
@@ -1198,7 +1197,7 @@ np.ndarray or number and correct dtype, the compatible nodata value.
         # even when there is no invalid value (true_divide and floor_divide).
         # We find one exception, however, for modulo: np.ma.remainder is not called but np.remainder instead one the
         # masked array is the second input (an inconsistency in NumPy!), so we mirror this exception below:
-        if 'remainder' in ufunc.__name__:
+        if "remainder" in ufunc.__name__:
             final_ufunc = getattr(ufunc, method)
         else:
             try:
--- a/geoutils/georaster/raster.py
+++ b/geoutils/georaster/raster.py
@@ -527,13 +527,15 @@ class Raster:
 
     def __eq__(self, other: object) -> bool:
         """Check if a Raster's data and georeferencing is equal to another."""
-        from geoutils.misc import array_equal
 
         if not isinstance(other, type(self)):  # TODO: Possibly add equals to SatelliteImage?
             return NotImplemented
         return all(
             [
-                array_equal(self.data, other.data, equal_nan=True),
+                np.array_equal(self.data.data, other.data.data, equal_nan=True),
+                np.array_equal(self.data.mask, other.data.mask),
+                self.data.fill_value == other.data.fill_value,
+                self.data.dtype == other.data.dtype,
                 self.transform == other.transform,
                 self.crs == other.crs,
                 self.nodata == other.nodata,
@@ -545,7 +547,7 @@ class Raster:
 
     def _overloading_check(
         self: RasterType, other: RasterType | np.ndarray | Number
-    ) -> tuple[np.ndarray, np.ndarray | Number, float | int | list[int] | list[float] | None]:
+    ) -> tuple[np.ma.masked_array, np.ma.masked_array | Number, float | int | list[int] | list[float] | None]:
         """
         Before any operation overloading, check input data type and return both self and other data as either \
 a np.ndarray or number, converted to the minimum compatible dtype between both datasets.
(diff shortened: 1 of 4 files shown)


rhugonnet:
Merging this to check if it solves the issue of xDEM, which is apparently not linked to nodata (still fails with -9999), but more likely dtypes during arithmetic operations (changed in this PR), and that I cannot reproduce locally...

geoutils/georaster/raster.py
```diff
@@ -527,13 +527,15 @@ def __str__(self) -> str:
...
     def __eq__(self, other: object) -> bool:
         """Check if a Raster's data and georeferencing is equal to another."""
-        from geoutils.misc import array_equal
 
         if not isinstance(other, type(self)):  # TODO: Possibly add equals to SatelliteImage?
             return NotImplemented
         return all(
             [
-                array_equal(self.data, other.data, equal_nan=True),
+                np.array_equal(self.data.data, other.data.data, equal_nan=True),
```
adehecq, line 535: Sorry if I'm debating things that we maybe already discussed...
Should we not just compare the valid values here? I agree that it's important to compare the masks, but I feel that comparing the values of `self.data.data` that are masked could lead to a lot of confusions for most users...

geoutils/georaster/raster.py
```diff
@@ -527,13 +527,15 @@ def __str__(self) -> str:
 
     def __eq__(self, other: object) -> bool:
         """Check if a Raster's data and georeferencing is equal to another."""
```
adehecq, line 529: We should improve the docstring to clearly state what is compared for us to remember.

geoutils/georaster/raster.py
```diff
@@ -624,11 +626,7 @@ def _overloading_check(
         if (nodata1 is not None) and (out_dtype == dtype1):
             out_nodata = nodata1
 
-        # Convert output data to correct dtype and masked_array
-        if isinstance(other_data, np.ndarray):
-            other_data = np.ma.asarray(other_data).astype(out_dtype, copy=False)
-
-        self_data = self.data.astype(out_dtype)
+        self_data = self.data
```
adehecq, line 629: Are you sure the call to `astype` is not needed anymore? I guess if the tests were well designed (hum hum) they would fail otherwise...

tests/test_georaster.py
```diff
@@ -557,24 +542,30 @@ def test_copy(self, example: str) -> None:
             assert r.__getattribute__(attr) == r2.__getattribute__(attr)
 
         # Check data array
-        assert geoutils.misc.array_equal(r.data, r2.data, equal_nan=True)
+        assert np.array_equal(r.data, r2.data, equal_nan=True)
```
adehecq, line 545: Why not use the __eq__ functionality here? It would be simpler and easier to understand.

tests/test_georaster.py
```diff
@@ -557,24 +542,30 @@ def test_copy(self, example: str) -> None:
...
         r2 = r.copy(new_array=r.data)
         assert r == r2
 
-        # Same when passing the new array as a NaN ndarray
+        # When passing the new array as a NaN ndarray, only the valid data is equal, because masked data is NaN in one
+        # case, and -9999 in the other
         r_arr = gu.spatial_tools.get_array_and_mask(r)[0]
         r2 = r.copy(new_array=r_arr)
+        assert np.ma.allequal(r.data, r2.data)
+        # If a nodata value exists, and we update the NaN pixels to be that nodata value, then the two Rasters should
```
adehecq, line 565: This test would not be needed if comparing only valid values and mask in __eq__

tests/test_georaster.py
```diff
@@ -1626,23 +1639,6 @@ def test_coords(self) -> None:
             assert yy.min() == pytest.approx(img.bounds.top + hy)
             assert yy.max() == pytest.approx(img.bounds.bottom - hy)
 
-    def test_eq(self) -> None:
```
adehecq, line 1629: Did you just delete this test? I don't see any replacement, but maybe I missed something...

adehecq: Ok, there was a duplicate in test_equal, right?

tests/test_georaster.py
```diff
@@ -2166,44 +2182,44 @@ def test_ops_2args_implicit(self) -> None:
         # Addition
         assert r1 + r2 == self.from_array(r1.data + r2.data, rst_ref=r1)
         assert r1_f32 + r2 == self.from_array(r1_f32.data + r2.data, rst_ref=r1)
-        assert array_3d + r2 == self.from_array(array_3d + r2.data, rst_ref=r1)
-        assert r2 + array_3d == self.from_array(r2.data + array_3d, rst_ref=r1)
-        assert array_2d + r2 == self.from_array(array_2d[np.newaxis, :, :] + r2.data, rst_ref=r1)
-        assert r2 + array_2d == self.from_array(r2.data + array_2d[np.newaxis, :, :], rst_ref=r1)
-        assert r1 + floatval == self.from_array(r1.data.astype("float32") + floatval, rst_ref=r1)
```
adehecq, line 2173: Ok, with reference to my comment on astype earlier, you changed the behavior.
This was by intention that, during an arithmetic operation, the dtype was changed to be compatible with the dtype of both elements. For example, if you add a float to a Raster of dtype uint8, the output was a float32. Now it will be a uint8, which means that wrapping will occur for values > 255, which can be surprising.

geoutils/georaster/raster.py
rhugonnet: Yes, NumPy arithmetic already takes care of it all, we just have to pass it.
The tests on arithmetic operations are actually pretty thorough! (especially with the new `__eq__`!)

tests/test_georaster.py
```diff
@@ -137,9 +122,9 @@ def test_loading(self) -> None:
         assert r.height == 655
         assert r.shape == (r.height, r.width)
         assert r.count == 1
-        assert geoutils.misc.array_equal(r.dtypes, ["uint8"])
+        assert np.array_equal(r.dtypes, ["uint8"])
```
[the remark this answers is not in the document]
rhugonnet: I guess this series of test was copy/pasted from similar ones above that use `array_equal` because there is more than 1 `dtype` item. Fine to leave it like this.

tests/test_georaster.py
rhugonnet: Yes this was the original version (day 1!), which has been duplicated since and is useless now

tests/test_georaster.py
rhugonnet: The idea is simply to be fully consistent with NumPy arithmetic of MaskedArray objects. This way users can refer to NumPy directly, and we don't have to explain a special behaviour!

Surprised to hear about the issue you mention, I didn't think NumPy would not believe like (it automatically fetches the minimal common `dtype`, which would be `float` in that case).
Actually, after some tests, I cannot reproduce it:
```python
r1.data
Out[4]: 
masked_array(
  data=[[[136, 209, 209, 133, 206],
         [  0, 238, 181,  14,  84],
         [115, 109,  89,  87,  99],
         [ 28,  41, 222, 129,   6],
         [123,  91, 120, 112, 204]]],
  mask=False,
  fill_value=999999,
  dtype=uint8)
(1.5 + r1).data
Out[5]: 
masked_array(
  data=[[[137.5, 210.5, 210.5, 134.5, 207.5],
         [  1.5, 239.5, 182.5,  15.5,  85.5],
         [116.5, 110.5,  90.5,  88.5, 100.5],
         [ 29.5,  42.5, 223.5, 130.5,   7.5],
         [124.5,  92.5, 121.5, 113.5, 205.5]]],
  mask=False,
  fill_value=1e+20)
(r1 + 1.5).data
Out[6]: 
masked_array(
  data=[[[137.5, 210.5, 210.5, 134.5, 207.5],
         [  1.5, 239.5, 182.5,  15.5,  85.5],
         [116.5, 110.5,  90.5,  88.5, 100.5],
         [ 29.5,  42.5, 223.5, 130.5,   7.5],
         [124.5,  92.5, 121.5, 113.5, 205.5]]],
  mask=False,
  fill_value=1e+20)
```

We could bring this up in the meeting early October to see what's the general feeling. I'm definitely in favour of being 100% consistent with NumPy! :smile:

tests/test_georaster.py
rhugonnet: But then the others tests would have to change. We need to perform both functionalities in any case, see answer to point above.

geoutils/georaster/raster.py
adehecq: Actually, I think the \_\_eq\_\_ feature will be the more intuitive for users, as they just have to run `rst1 == rst2`. So I would make sure this one does not have counter-intuitive tests, like checking masked values. We could have a function called `strictly_equal` for that, that we would mostly only use for tests.

tests/test_georaster.py
adehecq: But in brief, yes I agree let's stay consistent with NumPy's behavior for operations.

geoutils/georaster/raster.py
rhugonnet: Solved in #317

geoutils/georaster/raster.py
rhugonnet: Issue opened #316

PR #314, opened 2022-09-18
Add test for deprecator function

rhugonnet:
All in the title, mirrors the test in xDEM.

### DIFF
--- a/tests/test_misc.py
+++ b/tests/test_misc.py
@@ -4,11 +4,10 @@ import os
 import warnings
 
 import pytest
+import yaml  # type: ignore
 
 import geoutils
 import geoutils.misc
-
-import yaml  # type: ignore
 
 
 class TestMisc:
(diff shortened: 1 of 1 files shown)


tests/test_misc.py
```diff
@@ -36,3 +41,60 @@ def test_environment_files(self) -> None:
...
+            else None
+        )
+
+        # Define a function with no use that is marked as deprecated.
+        @geoutils.misc.deprecate(removal_version, details=details)  # type: ignore
+        def useless_func() -> int:
+            return 1
+
+        # If True, a warning is expected. If False, a ValueError is expected.
+        should_warn = removal_version is None or removal_version > current_version
```
adehecq, line 74: Comparing the strings directly will not work, e.g. "0.0.10" > "0.0.8" yields False.
You need to use https://packaging.pypa.io/en/latest/version.html.

rhugonnet: Actually, no correction needed, looks like this is supported by default by Python now:

```python
removal_version
Out[8]: '0.0.8'
current_version
Out[9]: '0.0.7'
removal_version > current_version
Out[10]: True
removal_version < current_version
Out[11]: False
'0.1' > '0.2'
Out[12]: False
'0.1' < '0.2'
Out[13]: True
'0.1' <= '0.1'
Out[14]: True
```

adehecq: No, you haven't tested my counter-example: "0.0.10" > "0.0.8" does not work. 
As stated [here](https://www.geeksforgeeks.org/string-comparison-in-python/), "The relational operators compare the Unicode values of the characters of the strings from the zeroth index till the end of the string. It then returns a boolean value according to the operator used." The unicode value has nothing to do with the integers.
I know that Erik used that comparison in the past, but we had this discussion already and since then I am aware of this issue. I guess what you copy/pasted was an older piece of code that had never been fixed...

adehecq: We can even try with a simpler example: `"10" > "8"` returns False.

PR #317, opened 2022-09-21
Elaborate in `__eq__` docstring

rhugonnet:
Account for @adehecq comment a posteriori: https://github.com/GlacioHack/geoutils/pull/313#discussion_r975422348

### DIFF
--- a/geoutils/georaster/raster.py
+++ b/geoutils/georaster/raster.py
@@ -526,7 +526,8 @@ class Raster:
         return self.info()
 
     def __eq__(self, other: object) -> bool:
-        """Check if a Raster's data and georeferencing is equal to another."""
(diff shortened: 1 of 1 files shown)


geoutils/georaster/raster.py
```diff
@@ -526,7 +526,8 @@ def __str__(self) -> str:
         return self.info()
 
     def __eq__(self, other: object) -> bool:
-        """Check if a Raster's data and georeferencing is equal to another."""
+        """Check if a Raster's data (including unmasked values), mask, nodata, and georeferencing are all equal to
```
adehecq, line 529: You mean "including masked values"?
I would be exhaustive, so you're missing fill_value and dtype.

PR #320, opened 2022-09-26
Fix `deprecate` version check

rhugonnet:
Fix version check in our `deprecate` decorator using `packaging.Version`

Didn't realize this what also an issue in `misc.py` when solving #315


rhugonnet:
Same as previous PR on this, so I merge

PR #321, opened 2022-09-26
Add caching to CI environment setup

### LINKED ISSUE #322: Python 3.8 and 3.9 are upgraded to 3.10 in CI
Since #301, by using `mamba env update` in the CI, we mistakenly update the Python 3.8 and 3.9 into 3.10. Same goes for xDEM. Unfortunately, the different `conda` options for freezing packages do not work well with update.

I will try to solve this in #321 somehow.

rhugonnet:
## Summary

This PR enables the caching of conda environment using the methods of `conda-incubator/setup-miniconda` and `github/actions/cache` to avoid long running CI times, especially on Mac and Windows.

This PR also fixes a new error during the sphinx build detected by `sphinx-autodoc-typehints` because `rioxarray` was not listed in `dev-environment.yml`.

## In short

**CI durations before:**

- **ubuntu:** 6-8min
- **mac:** 13-16min 
- **windows:** 13-16min

**CI durations after:**

All OSs take 3-5 min :partying_face:! (~1 min for uncaching environment, ~3-4min for testing). However, it looks like some runs are sometimes delayed when uncaching (maybe by the others doing the same thing), so it can get up to ~10min. Not sure how we could prevent that.
EDIT: this seems out of our hands, see issue here: https://github.com/actions/cache/issues/752#issuecomment-1119931553. Hopefully, it will be solved by GitHub soon :smile:!

## How does it work

The implementation is basically a copy of the tutorial in this subsection of `conda-incubator/setup-miniconda`: https://github.com/conda-incubator/setup-miniconda#caching-environments. It caches all conda-related packages (not the ones from pip; this could be doable but does not seem necessary as they are test-only and very fast install).

The cache is based on a key to differentiate cache versions and reset them if necessary. In our case, the key is `os`-`python_version`-`year_and_month`-`hash_of_yml_file`-`cache_nb`:

- The first two serve to differentiate the `os` and `python_version` runs in our CI. 
- The third `year_and_month` serves to automatically reset cache every month to account for updates of other packages,
- The fourth `hash_of_yml_file` serves to automatically change the key every time the `dev-environment.yml` file is modified, and ignore previous caches. 
- The last is set to `0` and can be increased (e.g. to `1`, `2`) to manually force the cache to reset if we want to.

The cache is stored in GitHub Actions of GeoUtils, with 10GB max (we use ~half of it with the 9 caches we have for 3 OSs and 3 Python versions), and is deleted automatically after 7 days if not used.

## The tricky part

However, in our case, the last step of the tutorial (https://github.com/conda-incubator/setup-miniconda#caching-environments) based on `mamba env update` does not work, for several reasons:

- This step updates the python version while we want it fixed for the tests e.g., `3.8`, `3.9` and `3.10` independently. This is actually an issue since #301, see #322.
- The `conda env` commands are the only ones that can take `.yml` files that we have at the base of the repository as arguments, but are also the ones without the options we would like to use such as `--freeze-installed` to update our environment in 2 steps without modifying python versions or other packages !

So this is problematic for us... As a reminder, our CI environment is created and updated the following way:

1. A **1st step** checks that `environment.yml` works to import package, 
2. A **2nd step** updates the dependencies listed in `dev-environment` and not in `environment.yml` to be able to run the tests, as started in #296.

I looked at some packages that try to deal with this. For instance, `dask/distributed` manually copies and updates the `dev-environment.yml` with 3 environment files for 3 python versions, which I found a bit redundant/hard to maintain... (https://github.com/dask/distributed/blob/main/.github/workflows/tests.yaml)

## Homemade solution

To remedy this, this PR:

- Re-adds function `diff_environment_yml` in `misc.py` for differencing our two `yml` files (base and dev), which was removed in #301, and 
- Adds a second function `get_env_yml_nopy` that is added as a python script in `.github/` requiring only `pyyaml` to run (as, for this step, geoutils is not yet installed) and which simply passes the content of the base `environment.yml` file as a list **without the Python version**.

Now, the environments are installed or updated using `mamba install/update` and the list of "package=version" dependencies captured as outputs of these two functions. This allows to specify the python version directly specified in the CI (i.e., we won't need to update anything when we move to different Python versions), and to use arguments such as `--freeze-installed` to update with the `dev-...` packages for testing, while still checking that the `environment.yml` file is sufficient to import the package.

Voilà !
I think it's not such a recurrent issue to have to deal with such long-solving dependencies, while keeping in mind that we might once in a while add new ones. It seems quite specific to georeferencing packages, so we really needed this! :slightly_smiling_face: 

Resolves #322


dev-environment.yml
```diff
@@ -26,3 +26,4 @@ dependencies:
   - sphinx-gallery
   - scikit-image
   - pyyaml
+  - rioxarray
```
adehecq, line 29: Why is it needed now?

adehecq (approved):
Wow, this looks like a lot of trial and error !
I hope this will work on the long term. Thanks for spending a lot of time on this, which should save us quite some time in the future !

dev-environment.yml
rhugonnet: For typing in `raster.py`: https://github.com/GlacioHack/geoutils/blob/main/geoutils/georaster/raster.py#L1756

This was caught by `sphinx-autodoc-typehints` that raised an error during the sphinx build (and thus made our tests fail). This error was new, and the cause seemed to have been the upgrade of `sphinx-autodoc-typehints`.

Adding this to the description.

PR #325, opened 2022-10-13
Upgrade `pre-commit`

rhugonnet:
Our current build is not passing because of `pre-commit` issues that arose since CI last run for #321 (sometime in the last two weeks). 
This PR solves those by upgrading packages, and also adding the black profile to `isort`.

After the upgrade, there is only two new mypy errors related to `__init__`  calls used in `geovector`. Those are not crucial (related to potential subclassing), so they are ignored for now.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
         - repo: https://github.com/pre-commit/pre-commit-hooks
-          rev: v4.0.1
+          rev: v4.3.0
           hooks:
                 - id: check-yaml
                 - id: end-of-file-fixer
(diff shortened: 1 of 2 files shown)


rhugonnet:
PR ready to merge

adehecq (approved):
I believe there is a way to automate this with a bot. In another repo that I copied from a template, there was such an action (which I found annoying there and hence deleted). It's here: https://github.com/adehecq/ragmac_xdem/commit/81c9ad6522258a9563a82f9c52edffe880d947a6.
But in the meantime, we're good with this one !

adehecq:
Merging this, as it may solve the CI error in PR #326.

PR #326, opened 2022-10-13
Fix small bug in spatial_tools

adehecq:
This is a one-line bug fix, so I'll just merge it.


adehecq:
Bypassing review as this is a minor fix.

PR #328, opened 2022-10-16
Fix `pip` dependency listing in CI

rhugonnet:
To match the issues found in https://github.com/GlacioHack/xdem/pull/320, in case GeoUtils relies more on `pip` in `environment.yml` and `dev-environment.yml` in the future.

The "pop" was not happening at the right line, and some functions were listing the dictionary key instead of values.


rhugonnet:
Ready to merge!

adehecq (approved):

PR #335, opened 2023-01-04
Subscript-crop `Raster` or `Vector` by bracket call `[]`, consistency and bug fixes

### LINKED ISSUE #213: Subscript a Raster with a Vector
The main principle of GeoUtils is to bridge Vectors and Rasters. Wouldn't it be cool if a Vector could be provided as a mask to a Raster by simply subscripting it?

```python
img = gu.Raster("path/to/img.tif")
area_of_interest = gu.Vector("path/to/aoi.shp")


masked_img = img[area_of_interest]
```
The underyling functionality would be something alike:
```python
mask = area_of_interest.create_mask(img)

masked_img = img.copy()
masked_img.data.mask = masked_img.data.mask | mask.reshape(masked_img.data.shape)
```

The `masked_img` in the top could be several different types:

- A `Raster` in the same grid as the input
- A cropped `Raster` in a new grid.
- A 2D array of the masked values.
- A 1D array of the masked values.

I think a new `Raster` instance with the same grid should be the output, but I am open to discussion regarding this.

### LINKED ISSUE #262: `Raster.crop` does not reproject the bounding box of a Raster object provided as `cropGeom`
It's all in the title! I'm not sure why this behaviour would be intended?

We should create a polygon of the bounding box and reproject it to the same crs than `Raster` before cropping.

### LINKED ISSUE #268: Shapely triggers DeprecationWarnings
**Describe the bug**
Shapely 2.0 will be quite breaking, so there are lots of `DeprecationWarning`s floating around. This should be fixed before they turn into errors!

```
tests/test_geovector.py::TestSynthetic::test_generate_voronoi
  ./geoutils/geovector.py:448: ShapelyDeprecationWarning: Iteration over multi-part geometries is deprecated and will be removed in Shapely 2.0. Use the `geoms` property to access the constituent parts of a multi-part geometry.
    exteriors = [p.exterior for p in geom]
```

**To Reproduce**
Run the test suite.

**Expected behavior**
No (shapely) warnings should be triggered.

### LINKED ISSUE #276: Remove warning when using `Vector.create_mask()` only with `xres` argument
When creating a mask from a `Vector` only passing the resolution `xres` which behaves according to the description as _"Only xres is mandatory, by default yres=xres and bounds/crs are set to self's."_, a UserWarning is raised which might not desirable (or need to be rephrased):
```python
        mask = outlines.create_mask(xres=res)
```
>  UserWarning: Bounds not a multiple of xres/yres, use rounded bounds
>   warnings.warn("Bounds not a multiple of xres/yres, use rounded bounds")

This will always appear, as the shapefile bounds will virtually always not be rounded to a multiple of `xres`. So I'm thinking we could hide this warning in that occurence? Alternatively, we could rephrase as the current warning lets us think something went wrong (when it didn't), saying for instance "_Bounds were not a multiple of xres/yres, so lower bounds were selected to define the start of the grid and upper bounds redefined._"

### LINKED ISSUE #333: Polygonize does not support all `Raster` dtypes
Error in `geopandas` using `GeoDataFrame.from_features`:
```
ValueError: image dtype must be one of: int16, int32, uint8, uint16, float32
```

Would be more practical to be able to go directly from mask -> shapefile, or even other integer/float dtypes.

rhugonnet:
## Summary

This PR updates the `get_geoprojected_bounds` function, moving its core to `projtools` to make it consistent and accessible for both `Raster` and `Vector` classes. 
Then, it adds the same options in `Vector.crop()` than for `Raster.crop()` (= passing another `Raster`, `Vector` or a list of bounds). 
Finally, it implements the `__getitem__` (= `[]`) method for `Raster` and `Vector`, which calls their underlying `crop` functionality (= subsetting), as suggested previously by @erikmannerfelt.

```python
rst = Raster(fn_rst)
vec = Vector(fn_vec)

# Get only Raster on extent of Vector
rst = rst[vec]
```

Other than that, several bug fixes and test additions.

## Discussion on the `__getitem__` implementation

Several things:

- We could add the option of provides `slice` objects, which would then be used on `self.data` for `Raster` (subsetting `np.MaskedArray` in 2D) and `self.ds` on `Vector` (subsetting `gpd.GeoDataFrame` in 1D).
- We could, by default, `reproject` instead of `crop` when it is `Raster[Raster]` (only one where it makes sense)? The resampling algorithm could be passed as second argument `Raster[Raster, "bilinear"]`.

## Related issues and more details

This PR:
- Resolves #276,
- Resolves #268,
- Resolves #262,
- Resolves #213,
- Resolves #333.

Also it:
- Removes the dependency of `projtools.py` on `raster.py` and `geovector.py`,
- Moves the core of the `get_geoprojected_bounds` function in `Raster` to `projtools.py`,
- Adds `get_geoprojected_bounds` as a method for the `Vector` class,
- Mirrors the crop functionality of the `Raster` class for `Vector` objects (can provide any type of geometry: bounds, `Raster`, `Vector`),
- Adds more test for `Raster.polygonize()`,
- Solves a new bug with checking void in `intersection` triggered by package updates.

### DIFF
--- a/geoutils/georaster/raster.py
+++ b/geoutils/georaster/raster.py
@@ -1830,11 +1830,8 @@ np.ndarray or number and correct dtype, the compatible nodata value.
         # Calculate intersection of bounding boxes
         intersection = projtools.merge_bounds([self.bounds, rst_bounds_sameproj], merging_algorithm="intersection")
 
-        print(intersection)
-        print(type(intersection[0]))
-        print(intersection == (np.nan, np.nan, np.nan, np.nan))
         # check that intersection is not void, otherwise return 0 everywhere
-        if intersection == () or intersection == (np.nan, np.nan, np.nan, np.nan):
+        if intersection == () or intersection == (float('nan'), float('nan'), float('nan'), float('nan')):
             warnings.warn("Intersection is void")
             return (0.0, 0.0, 0.0, 0.0)
 
--- a/geoutils/georaster/raster.py
+++ b/geoutils/georaster/raster.py
@@ -3,6 +3,7 @@ geoutils.georaster provides a toolset for working with raster data.
 """
 from __future__ import annotations
 
+import math
 import os
 import warnings
 from collections import abc
@@ -14,7 +15,6 @@ from typing import IO, Any, Callable, TypeVar, overload
 import geopandas as gpd
 import matplotlib
 import matplotlib.pyplot as plt
-import math
 import numpy as np
 import pyproj
 import rasterio as rio
--- a/geoutils/georaster/raster.py
+++ b/geoutils/georaster/raster.py
@@ -1830,6 +1830,7 @@ np.ndarray or number and correct dtype, the compatible nodata value.
         # Calculate intersection of bounding boxes
         intersection = projtools.merge_bounds([self.bounds, rst_bounds_sameproj], merging_algorithm="intersection")
 
+        print(intersection)
         # check that intersection is not void, otherwise return 0 everywhere
         if intersection == () or intersection == (np.nan, np.nan, np.nan, np.nan):
             warnings.warn("Intersection is void")
--- a/geoutils/georaster/raster.py
+++ b/geoutils/georaster/raster.py
@@ -1409,8 +1409,8 @@ np.ndarray or number and correct dtype, the compatible nodata value.
 
         :param dst_ref: a reference raster. If set will use the attributes of this
             raster for the output grid. Can be provided as Raster/rasterio data set or as path to the file.
-        :param dst_crs: Specify the Coordinate Reference System to reproject to. If dst_ref not set, defaults to
-        self.crs.
+        :param dst_crs: Specify the Coordinate Reference System or EPSG to reproject to. If dst_ref not set,
+            defaults to self.crs.
         :param dst_size: Raster size to write to (x, y). Do not use with dst_res.
         :param dst_bounds: a BoundingBox object or a dictionary containing left, bottom, right, top bounds in the
         source CRS.
--- a/geoutils/georaster/raster.py
+++ b/geoutils/georaster/raster.py
@@ -1831,6 +1831,7 @@ np.ndarray or number and correct dtype, the compatible nodata value.
         intersection = projtools.merge_bounds([self.bounds, rst_bounds_sameproj], merging_algorithm="intersection")
 
         print(intersection)
+        print(intersection == (np.nan, np.nan, np.nan, np.nan))
         # check that intersection is not void, otherwise return 0 everywhere
         if intersection == () or intersection == (np.nan, np.nan, np.nan, np.nan):
             warnings.warn("Intersection is void")
--- a/geoutils/georaster/raster.py
+++ b/geoutils/georaster/raster.py
@@ -1831,6 +1831,7 @@ np.ndarray or number and correct dtype, the compatible nodata value.
         intersection = projtools.merge_bounds([self.bounds, rst_bounds_sameproj], merging_algorithm="intersection")
 
         print(intersection)
+        print(type(intersection[0]))
         print(intersection == (np.nan, np.nan, np.nan, np.nan))
         # check that intersection is not void, otherwise return 0 everywhere
         if intersection == () or intersection == (np.nan, np.nan, np.nan, np.nan):
--- a/geoutils/georaster/raster.py
+++ b/geoutils/georaster/raster.py
@@ -2484,9 +2484,23 @@ np.ndarray or number and correct dtype, the compatible nodata value.
 
             raise ValueError("in_value must be a number, a tuple or a sequence")
 
+        # GeoPandas.from_features() only supports certain dtypes, we find the best common dtype to optimize memory usage
+        # TODO: this should be a function independent of polygonize, reused in several places
+        gpd_dtypes = ["uint8", "uint16", "int16", "int32", "float32"]
+        list_common_dtype_index = []
+        for gpd_type in gpd_dtypes:
+            polygonize_dtype = np.find_common_type([gpd_type, self.dtypes[0]], [])
+            if str(polygonize_dtype) in gpd_dtypes:
+                list_common_dtype_index.append(gpd_dtypes.index(gpd_type))
+        if len(list_common_dtype_index) == 0:
+            final_dtype = "float32"
+        else:
+            final_dtype_index = min(list_common_dtype_index)
+            final_dtype = gpd_dtypes[final_dtype_index]
+
         results = (
             {"properties": {"raster_value": v}, "geometry": s}
-            for i, (s, v) in enumerate(shapes(self.data, mask=bool_msk, transform=self.transform))
+            for i, (s, v) in enumerate(shapes(self.data.astype(final_dtype), mask=bool_msk, transform=self.transform))
         )
 
         gdf = gpd.GeoDataFrame.from_features(list(results))
--- a/geoutils/georaster/raster.py
+++ b/geoutils/georaster/raster.py
@@ -1315,7 +1315,9 @@ np.ndarray or number and correct dtype, the compatible nodata value.
         inplace: bool = True,
     ) -> RasterType | None:
         """
-        Crop the Raster to a given extent.
+        Crop the Raster to a given extent, or bounds of a raster or vector.
+
+        Reprojection is done on the fly if georeferenced objects have different projections.
 
         :param cropGeom: Geometry to crop raster to, as either a Raster object, a Vector object, or a list of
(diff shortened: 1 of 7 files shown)


tests/test_georaster.py
```diff
@@ -750,28 +771,42 @@ def test_crop(self, data: list[str]) -> None:
             cropGeom[2] - rand_float * r.res[0],
             cropGeom[3] - rand_float * abs(r.res[1]),
         ]
-        r_cropped = r.crop(cropGeom2, inplace=False, mode="match_extent")
+        with warnings.catch_warnings():
```
erikmannerfelt, line 774: Could you add a comment for why UserWarnings are expected here?

tests/test_georaster.py
```diff
@@ -704,6 +711,20 @@ def test_crop(self, data: list[str]) -> None:
         r_cropped2 = r.crop(r_cropped, inplace=False)
         assert r_cropped2 == r_cropped
 
+        # Check that bound reprojection is done automatically if the CRS differ
+        with warnings.catch_warnings():
```
erikmannerfelt, line 715: Same as below; why do we expect warnings here?

tests/test_georaster.py
erikmannerfelt: Also, I'm pretty sure that all our warnings are UserWarnings, so that means all of them get suppressed.

geoutils/spatial_tools.py
```diff
@@ -136,8 +137,8 @@ def load_multiple_rasters(
 
     # Optionally, crop the rasters
     if crop:
-        # Check that intersection is not void
-        if intersection == ():
+        # Check that intersection is not void (changed to NaN instead of empty tuple end 2022)
+        if intersection == () or all(math.isnan(i) for i in intersection):
```
erikmannerfelt, line 141: What's the advantage of  `math.isnan` to `~np.isfinite`?

tests/test_georaster.py
```diff
@@ -1798,22 +1836,42 @@ def test_resampling_str(self) -> None:
...
-        value = np.unique(img)[0]
+        img = gr.Raster(example)
 
-        pixel_area = np.sum(img == value) * img.res[0] * img.res[1]
+        # -- Test 1: basic functioning of polygonize --
 
-        polygonized = img.polygonize(value)
+        # Get unique value for image and the corresponding area
+        value = np.unique(img)[0]
+        pixel_area = np.sum(img == value) * img.res[0] * img.res[1]
```
erikmannerfelt, line 1849: Could you change this to `np.count_nonzero()`? I know they're synonymous in the case of booleans, but "count nonzero" makes things easier to understand in the long term in my opinion.

tests/test_georaster.py
```diff
@@ -1798,22 +1836,42 @@ def test_resampling_str(self) -> None:
...
         assert polygon_area == pytest.approx(pixel_area)
         assert isinstance(polygonized, gv.Vector)
         assert polygonized.crs == img.crs
 
+        # -- Test 2: data types --
+
+        # Check that polygonize works as expected for any input dtype (e.g. float64 being not supported by GeoPandas)
+        for dtype in ["uint8", "int8", "uint16", "int16", "uint32", "int32", "float32", "float64"]:
+            img_dtype = img.copy()
+            with warnings.catch_warnings():
```
erikmannerfelt, line 1865: What warnings are suppressed?

erikmannerfelt (commented):
Hey Romain!
These additions are awesome. You're really contributing to this being maintainable in the long term! On the subject of maintainability, I have some comments on details that I struggle to understand, but generally, it's great!

Regarding your two suggestions about the subscripting:
- On the first, do you mean to slice by indices like `arr.isel()` or `df.iloc[]` in xarray and pandas, respectively? If so, it would actually be pretty cool, but I struggle to see a use case. For iterating along blocks perhaps? Or subsetting for debug purposes? Those two are good arguments in my head, but for the amount of complexity they introduce, it might not be a great idea unless there are more uses.
- I personally think the subscripting syntax should only use crop. Reproject can be very cpu-intensive and should in my opinion always be explicit. Cropping, however, is really fast, so there the subscripting syntax makes a ton of sense to me.

geoutils/projtools.py
```diff
@@ -15,12 +15,9 @@
 from shapely.geometry.base import BaseGeometry
 from shapely.geometry.polygon import Polygon
 
-from geoutils.georaster import Raster
-from geoutils.geovector import Vector
-
 
 def bounds2poly(
-    boundsGeom: list[float] | rio.io.DatasetReader | Raster | Vector,
+    boundsGeom: list[float] | rio.io.DatasetReader,
```
adehecq, line 20: why did you delete it? It's actually meant to take a Raster or Vector instance.

geoutils/georaster/raster.py
```diff
@@ -1375,13 +1385,13 @@ def crop(
     def reproject(
         self: RasterType,
         dst_ref: RasterType | rio.io.Dataset | str | None = None,
-        dst_crs: CRS | str | None = None,
+        dst_crs: CRS | str | int | None = None,
```
adehecq, line 1388: why is int needed? For EPSG codes? If so, it should be clarified in the docstring what argument can be passed.

tests/test_georaster.py
```diff
@@ -750,28 +771,42 @@ def test_crop(self, data: list[str]) -> None:
...
+            r_cropped2 = r.crop(r_cropped, inplace=False, mode="match_extent")
         assert r_cropped2 == r_cropped
 
         # -- Test with CropGeom being a Vector -- #
         outlines = gu.Vector(outlines_path)
-        outlines.ds = outlines.ds.to_crs(r.crs)
-        r_cropped = r.crop(outlines, inplace=False)
+
+        # First, we reproject manually the outline
+        outlines_reproj = gu.Vector(outlines.ds.to_crs(r.crs))
```
adehecq, line 792: When reading your PR, I was thinking that in the same effort (making crop be similar between Raster and Vector), we should make reproject more consistent between the two, i.e. have a Vector.reproject function.

adehecq (approved):
Great idea to make crop more consistent and thanks for fixing all the small issues.
The brackets are a nice feature, but they would need to be documented at some point, otherwise, we'll just forget that it's possible !

geoutils/projtools.py
rhugonnet: Cannot use `Raster` or `Vector` in `projtools` due to circular import

tests/test_georaster.py
rhugonnet: Done, added a comment everywhere + the `message` argument with a regex pattern to only catch the desired warnings.

tests/test_georaster.py
rhugonnet: Done as above

geoutils/spatial_tools.py
rhugonnet: Hmmm, curious, for some reason when I tried with `np.isnan` the first time, it did not work, because the output of geopandas is a `float('nan')`, but now it works element-wise...
I changed to `np.isnan`.

tests/test_georaster.py
rhugonnet: Specified the match message.

tests/test_georaster.py
rhugonnet: Done

geoutils/georaster/raster.py
rhugonnet: Done

geoutils/geovector.py
```diff
@@ -120,7 +156,7 @@ def create_mask(
     ) -> np.ndarray:
         """
         Rasterize the vector features into a boolean raster which has the extent/dimensions of \
-the provided raster file.
+        the provided raster file.
```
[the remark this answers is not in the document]
rhugonnet: I don't know either... except that it makes the code hard to read, and does not respect Python indentation. I haven't seen that in any of the other packages I went through (NumPy, Pandas, etc).
I'll open an issue on this. For now, I know it displays fine in the API with this kind of tabulated text.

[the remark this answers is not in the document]
rhugonnet: Opened #338

tests/test_georaster.py
rhugonnet: I was thinking the same, but then reproject would only be about the CRS, right?

rhugonnet:
Thanks a lot! All updated, merging :smile:

tests/test_georaster.py
adehecq: Yes that's right, but still it would be more consistent between the two classes.

adehecq: Just added issue #340

Issue #311, opened 2022-09-15
Re-organize the structure of geoutils


adehecq:
We need to re-think the organization of some submodules of geoutils and write a clear plan for what functionalities we aim to include and where, write a release schedule. A non exhaustive list of questions and things to be addressed: 
- spatial_tools: probably rename and clarify the functionalities to be included in here (should filters be added here?). Move to the raster submodule?
- same for projtools
- move satimg: see #260 

Some related issues: 
- #255 
- #200 
- #51 
- #26.

PR #336, opened 2023-01-08
Add `proximity` and `buffer_metric` functionalities with `metric` option based on finding local UTM 

### LINKED ISSUE #145: Add equivalent for GDAL proximity functionality
This is not wrapped by rasterio: https://github.com/mapbox/rasterio/issues/319
Apparently could be done with `scipy.ndimage.morphology.distance_transform_edt`.

### LINKED ISSUE #202: Add method to check if two rasters have the same georeferencing
No

A method to check whether two DEMs have the same georeferencing/grid, or if reprojection is needed. For example, 
`raster1.is_same_grid(raster2)`

Currently, the best approach is to compare all attributes, e.g.
```
if (dem_1990.shape != dem_2009.shape) or (dem_1990.transform != dem_2009.transform):
    dem_1990 = dem_1990.reproject(dst_ref=dem_2009)
```

**Additional context**
None

### LINKED ISSUE #267: GeoPandas warns about incorrect buffer results due to geographic CRS
**Describe the bug**
In the test suite, an error is triggered about the use of a geographic CRS which may give wrong buffers (I don't know why, but that's what it says). 
```
tests/test_geovector.py::TestSynthetic::test_buffer_without_overlap
  ./geoutils/geovector.py:373: UserWarning: Geometry is in a geographic CRS. Results from 'buffer' are likely incorrect. Use 'GeoSeries.to_crs()' to re-project geometries to a projected CRS before this operation.
  
    merged_buffer = merged.buffer(buffer_size)
```

We should:

1. Silence/fix the warning in the test suite
2. Make sure that it doesn't point to a larger underlying problem (e.g. why did it only start occurring recently?)

**To Reproduce**
Run the test suite or see the CI logs.

**Expected behavior**
No warning should come up

### LINKED ISSUE #330: Add a function to find local UTM zone/projection for given bounds
Some functionalities only work on projected coordinates (like xdem's slope).
It would be great to have a tool to automatically find a local projected CRS for given bounds. There is for example [this function](https://gis.stackexchange.com/a/423614) to automatically find a suitable UTM zone.

rhugonnet:
## Summary

This PR adds a `proximity()` function to the `Vector` and `Raster` classes that mirrors GDAL's proximity not implemented in `rasterio` (https://github.com/rasterio/rasterio/issues/319), and a `buffer_metric()` method for the `Vector` class to avoid the trouble of reprojecting from geographic system, buffer, than reverse-projecting, based on finding a local UTM zone.
It also adds a `Raster` method `equal_georeferenced_grid` to check if two rasters have the same georeferencing.

### Proximity

The proximity is computed using `scipy.distance_transform_edt` (https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.distance_transform_edt.html).

The functioning is as follows:

- For `Raster` classes, the proximity can be either computed directly from target values of the array `self.data`, or from a `Vector` passed as argument that is automatically rasterized on `self`. For target values, the default behaviour is the same as GDAL: a list of any number of values can be passed. If none are passed, all non-zero values are targets. For the distance unit, I also implemented both "pixel" and "georeferenced", but put the default as "georeferenced" contrary to GDAL's "pixel".

- For `Vector` classes, the proximity can be computed from `self` by passing a `Raster` object to rasterize the vector on.

The resulting proximity values are thoroughly tested against GDAL Python's `gdal.ComputeProximity()` with artificial and real Raster data. All results are the same, except for very complex case where a few percentage of values (<1%) differ by ~1% (not even a floating point issue, GDAL's algorithm is simply slightly different for some cases).

Additionally, any `geometry_type` can be passed to rasterize the `Vector` object with (default: `boundary`, which is the most common). And an `in_or_out` option can be passed to include only the outside or inside proximity for a geometry that have an area (`Polygon`, `MultiPolygon`, etc). 

### Buffer metric

This function is to avoid the trouble of reprojecting a `Vector` in a local metric system for buffering without deformation, and then reverse-projecting. This involves finding the local metric system wanted, reprojecting to it, recreating a`geopandas.GeoDataFrame` object in the right CRS with the buffer `geopandas.GeoSeries`, and reverse-projecting.

For now this function only supports reprojecting in the local UTM which, for speed, is approximated by the latitude/longitude coordinates of the centroid of non-projected vector (centroid are not exactly the same depending on projections, but generally robust enough?).

Two new functions `latlon_to_utm` and `utm_to_epsg` in `projtools` help converting coordinates to UTM zone and to related EPSG (maybe there's some new stuff out there, but a couple years ago the `utm` Python package didn't allow UTM zones near poles, which was annoying, and I didn't know of any package that could interpret the UTM zone into an EPSG code).

All functions of `projtools` and the `buffer_metric` method in `Vector` are tested. I also added the `metric` argument in `buffer_without_overlap` function, which does the same procedure.

## Issues closed

This PR:

- Resolves #330,
- Resolves #267,
- Resolves #145,
- Resolves #202.

And adds tests for all.

### DIFF
--- a/geoutils/georaster/raster.py
+++ b/geoutils/georaster/raster.py
@@ -2493,68 +2493,104 @@ np.ndarray or number and correct dtype, the compatible nodata value.
         vector: Vector | None = None,
         target_values: list[float] | None = None,
         geometry_type: str = "boundary",
-        in_or_out: str = "both",
-        distance_unit: str = "georeferenced",
+        in_or_out: Literal["in"] | Literal["out"] | Literal["both"] = "both",
+        distance_unit: Literal["pixel"] | Literal["georeferenced"] = "georeferenced",
     ) -> Raster:
         """
-        Proximity to a vector's geometry or to self's target pixels computed for each cell of a raster grid.
+        Proximity to this Raster's target pixels, or to a Vector's geometry, computed for each cell of this Raster's grid.
+
+        When passing a Vector, by default, the boundary of the geometry will be used. The full geometry can be used by
+        passing "geometry", or any lower dimensional geometry attribute such as "centroid", "envelope" or "convex_hull".
+        See all geometry attributes in the Shapely documentation at https://shapely.readthedocs.io/.
 
         :param vector: Vector for which to compute the proximity to geometry,
-            if not provided computed on self's target pixels.
-        :param target_values: (Only with self) List of target values to use for the proximity,
+            if not provided computed on this Raster target pixels.
+        :param target_values: (Only with Raster) List of target values to use for the proximity,
             defaults to all non-zero values.
-        :param geometry_type: (Only with vector) Type of geometry to use for the proximity, defaults to 'boundary'.
-        :param in_or_out: (Only with vector) Compute proximity only 'in' or 'out'-side the geometry, or 'both'.
+        :param geometry_type: (Only with a Vector) Type of geometry to use for the proximity, defaults to 'boundary'.
+        :param in_or_out: (Only with a Vector) Compute proximity only 'in' or 'out'-side the geometry, or 'both'.
         :param distance_unit: Distance unit, either 'georeferenced' or 'pixel'.
 
         :return: Proximity raster.
         """
 
-        # 1/ First, if there is a vector input, we rasterize the geometry type
-        # (works with .boundary that is a LineString (.exterior exists, but is a LinearRing)
-        if vector is not None:
-            # We create a geodataframe with the geometry type
-            boundary_shp = gpd.GeoDataFrame(geometry=vector.ds.__getattr__(geometry_type), crs=vector.crs)
-            # We mask the pixels that make up the geometry type
-            mask_boundary = Vector(boundary_shp).create_mask(self).squeeze()
-
-        else:
-            # We mask target pixels
-            if target_values is not None:
-                mask_boundary = np.logical_or.reduce(
-                    [self.get_nanarray() == target_val for target_val in target_values]
-                )
-            # Otherwise, all non-zero values are considered targets
-            else:
-                mask_boundary = self.get_nanarray().astype(bool)
-
-        # 2/ Now, we compute the distance matrix relative to the masked geometry type
-        if distance_unit.lower() == "georeferenced":
-            sampling: int | tuple[float | int, float | int] = self.res
-        elif distance_unit.lower() == "pixel":
-            sampling = 1
-        else:
-            raise ValueError('Distance unit must be either "georeferenced" or "pixel".')
-
-        # If not all pixels are targets, then we compute the distance
-        non_targets = np.count_nonzero(mask_boundary)
-        if non_targets > 0:
-            proximity = distance_transform_edt(~mask_boundary, sampling=sampling)
-        # Otherwise, pass an array full of nodata
-        else:
-            proximity = np.ones(np.shape(mask_boundary)) * np.nan
-
-        # 3/ If there was a vector input, apply the in_and_out argument to optionally mask inside/outside
-        if vector is not None:
-            if in_or_out == "both":
-                pass
-            elif in_or_out in ["in", "out"]:
-                mask_polygon = Vector(vector.ds).create_mask(self).squeeze()
-                if in_or_out == "in":
-                    proximity[~mask_polygon] = 0
-                else:
-                    proximity[mask_polygon] = 0
-            else:
-                raise ValueError('The type of proximity must be one of "in", "out" or "both".')
+        proximity = proximity_from_vector_or_raster(raster=self, vector=vector, target_values=target_values, geometry_type=geometry_type,
+                                                    in_or_out=in_or_out, distance_unit=distance_unit)
 
         return self.copy(new_array=proximity)
+
+
+# -----------------------------------------
+# Additional stand-alone utility functions
+# -----------------------------------------
+
+
+def proximity_from_vector_or_raster(
+        raster: Raster | None = None,
+        vector: Vector | None = None,
+        target_values: list[float] | None = None,
+        geometry_type: str = "boundary",
+        in_or_out: Literal["in"] | Literal["out"] | Literal["both"] = "both",
+        distance_unit: Literal["pixel"] | Literal["georeferenced"] = "georeferenced") -> np.ndarray:
+    """
+    (This function is defined here as mostly raster-based, but used in a class method for both Raster and Vector)
+    Proximity to a Raster's target values if no Vector is provided, otherwise to a Vector's geometry type rasterized on the Raster.
+
+    :param raster: Raster to burn the proximity grid on.
+    :param vector: Vector for which to compute the proximity to geometry,
+        if not provided computed on the Raster target pixels.
+    :param target_values: (Only with a Raster) List of target values to use for the proximity,
+        defaults to all non-zero values.
+    :param geometry_type: (Only with a Vector) Type of geometry to use for the proximity, defaults to 'boundary'.
+    :param in_or_out: (Only with a Vector) Compute proximity only 'in' or 'out'-side the geometry, or 'both'.
+    :param distance_unit: Distance unit, either 'georeferenced' or 'pixel'.
+    """
+
+    # 1/ First, if there is a vector input, we rasterize the geometry type
+    # (works with .boundary that is a LineString (.exterior exists, but is a LinearRing)
+    if vector is not None:
+        # We create a geodataframe with the geometry type
+        boundary_shp = gpd.GeoDataFrame(geometry=vector.ds.__getattr__(geometry_type), crs=vector.crs)
+        # We mask the pixels that make up the geometry type
+        mask_boundary = Vector(boundary_shp).create_mask(raster).squeeze()
+
+    else:
+        # We mask target pixels
+        if target_values is not None:
+            mask_boundary = np.logical_or.reduce(
+                [raster.get_nanarray() == target_val for target_val in target_values]
+            )
+        # Otherwise, all non-zero values are considered targets
+        else:
+            mask_boundary = raster.get_nanarray().astype(bool)
+
+    # 2/ Now, we compute the distance matrix relative to the masked geometry type
+    if distance_unit.lower() == "georeferenced":
+        sampling: int | tuple[float | int, float | int] = raster.res
+    elif distance_unit.lower() == "pixel":
+        sampling = 1
+    else:
+        raise ValueError('Distance unit must be either "georeferenced" or "pixel".')
+
+    # If not all pixels are targets, then we compute the distance
+    non_targets = np.count_nonzero(mask_boundary)
+    if non_targets > 0:
+        proximity = distance_transform_edt(~mask_boundary, sampling=sampling)
+    # Otherwise, pass an array full of nodata
+    else:
+        proximity = np.ones(np.shape(mask_boundary)) * np.nan
+
+    # 3/ If there was a vector input, apply the in_and_out argument to optionally mask inside/outside
+    if vector is not None:
+        if in_or_out == "both":
+            pass
+        elif in_or_out in ["in", "out"]:
+            mask_polygon = Vector(vector.ds).create_mask(raster).squeeze()
+            if in_or_out == "in":
+                proximity[~mask_polygon] = 0
+            else:
+                proximity[mask_polygon] = 0
+        else:
+            raise ValueError('The type of proximity must be one of "in", "out" or "both".')
+
+    return proximity
--- a/geoutils/georaster/raster.py
+++ b/geoutils/georaster/raster.py
@@ -27,7 +27,7 @@ from rasterio.crs import CRS
 from rasterio.features import shapes
 from rasterio.plot import show as rshow
 from rasterio.warp import Resampling
-from scipy.ndimage import map_coordinates
+from scipy.ndimage import map_coordinates, distance_transform_edt
 
 import geoutils.geovector as gv
 from geoutils._typing import AnyNumber, ArrayLike, DTypeLike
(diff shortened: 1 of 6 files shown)


rhugonnet:
Tests all passing locally (= will be all good PR #335 is merged)

geoutils/projtools.py
```diff
@@ -19,6 +19,76 @@
 from geoutils.geovector import Vector
 
 
+def latlon_to_utm(lat: float, lon: float) -> str:
```
erikmannerfelt, line 22: This whole function could be done with one long pyproj line:
```python
>>> lat = 78
>>> lon = 15
>>> pyproj.database.query_utm_crs_info("WGS 84", area_of_interest=pyproj.aoi.AreaOfInterest(lon, lat, lon, lat))[0].name.split(" ")[-1]
"33N"
```
Might save a few headaches down the line since most logic is handled by pyproj instead. No need to reinvent the wheel!

tests/test_georaster.py
```diff
@@ -26,6 +26,46 @@
...
+    from osgeo import gdal, gdalconst
+
+    # Initiate empty GDAL raster for proximity output
+    drv = gdal.GetDriverByName("MEM")
+    proxy_ds = drv.Create("", input_raster.shape[1], input_raster.shape[0], 1, gdal.GetDataTypeByName("Float32"))
+    proxy_ds.GetRasterBand(1).SetNoDataValue(-9999)
+
+    # Save input in temporary file to read with GDAL
+    # (avoids the nightmare of setting nodata, transform, crs in GDAL format...)
+    temp_dir = tempfile.TemporaryDirectory()
```
erikmannerfelt, line 43: I've realized the hard way that this does not delete the directory if the program crashes before this goes out of scope. Please change this to:
```python
with tempfile.TemporaryDirectory() as temp_dir:
    temp_path = os.path.join(temp_dir, "input.tif")  # The temp_dir variable is now just a string of the path.
    ...
```

tests/test_georaster.py
```diff
@@ -26,6 +26,46 @@
...
+    # Save array
+    proxy_array = proxy_ds.GetRasterBand(1).ReadAsArray().astype("float32")
+    proxy_array[proxy_array == -9999] = np.nan
+
+    # Close GDAL datasets
+    proxy_ds = None
+    ds_raster_in = None
+
+    # Clean temporary directory
+    temp_dir.cleanup()
```
erikmannerfelt, line 64: This is not required if used with a context manager (my suggestion above)

tests/test_georaster.py
```diff
@@ -1814,6 +1854,116 @@ def test_polygonize(self) -> None:
...
+        and tests in rasterize and shapely.
+        #TODO: Maybe add one test with an artificial vector to check it works as intended
+        """
+
+        # -- Test 1: with self's Raster alone --
+        raster1 = gu.Raster(self.landsat_b4_path)
+        prox1 = raster1.proximity()
+
+        # The raster should have the same extent, resolution and CRS
+        assert raster1.crs == prox1.crs
```
erikmannerfelt, line 1945: Hmm, we should have a function that's like `r1.equal_grid(r2)` which is true if the CRS, transform and shape is the same. Would be pretty useful! I might add it as an issue later.

erikmannerfelt: Or `is_aligned()`, `aligned()`, `same_place()` or something. Suggestions are welcome!

tests/test_georaster.py
```diff
@@ -1814,6 +1854,116 @@ def test_polygonize(self) -> None:
...
+
+        # It should change with target values specified
+        prox2 = raster1.proximity(target_values=[255])
+        assert not np.array_equal(prox1.data, prox2.data)
+
+        # -- Test 2: with a vector provided --
+        vector = gu.Vector(self.everest_outlines_path)
+
+        # With default options (boundary geometry)
+        prox3 = raster1.proximity(vector=vector)  # noqa
```
erikmannerfelt, line 1956: What's with the `# noqa`?

tests/test_geovector.py
```diff
@@ -195,6 +196,11 @@ def test_create_mask(self) -> None:
             eroded_diff = binary_erosion(diff.squeeze(), np.ones((abs(buffer) + 1, abs(buffer) + 1)))
             assert np.count_nonzero(eroded_diff) == 0
 
+    def test_proximity(self) -> None:
```
erikmannerfelt, line 199: Is this added for this coverall thing?

erikmannerfelt (commented):
Awesome functionality!

tests/test_georaster.py
erikmannerfelt: Speak of the devil: #202 !

geoutils/georaster/raster.py
```diff
@@ -2487,3 +2487,74 @@ def polygonize(
         gdf.set_crs(self.crs, inplace=True)
 
         return gv.Vector(gdf)
+
+    def proximity(
+        self,
+        vector: Vector | None = None,
+        target_values: list[float] | None = None,
+        geometry_type: str = "boundary",
+        in_or_out: str = "both",
```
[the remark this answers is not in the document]
rhugonnet: Good point, done!

geoutils/georaster/raster.py
```diff
@@ -2487,3 +2487,74 @@ def polygonize(
...
+        distance_unit: str = "georeferenced",
+    ) -> Raster:
+        """
+        Proximity to a vector's geometry or to self's target pixels computed for each cell of a raster grid.
+
+        :param vector: Vector for which to compute the proximity to geometry,
+            if not provided computed on self's target pixels.
+        :param target_values: (Only with self) List of target values to use for the proximity,
+            defaults to all non-zero values.
+        :param geometry_type: (Only with vector) Type of geometry to use for the proximity, defaults to 'boundary'.
```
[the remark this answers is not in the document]
rhugonnet: I've improved the description. 

I realize that the main reason I chose "boundary" as the default is that it is annoying to write `Vector(geopandas.GeoDataFrame(geometry=vector.ds.boundary, crs=vector.crs)`.

But to remedy this more directly, we should rather add functionalities to the `Vector` class to apply any attribute matching that of `Shapely` to the `GeoSeries`, turn it into a GeoDataFrame (the main question being: what metadata should be kept from the original one?) and put that in a new `Vector` object.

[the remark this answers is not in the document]
rhugonnet: Otherwise it is really long to use any geometry manipulation with the `Vector` class...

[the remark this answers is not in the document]
rhugonnet: Opened #337

geoutils/georaster/raster.py
```diff
@@ -2487,3 +2487,74 @@ def polygonize(
...
+        target_values: list[float] | None = None,
+        geometry_type: str = "boundary",
+        in_or_out: str = "both",
+        distance_unit: str = "georeferenced",
+    ) -> Raster:
+        """
+        Proximity to a vector's geometry or to self's target pixels computed for each cell of a raster grid.
+
+        :param vector: Vector for which to compute the proximity to geometry,
+            if not provided computed on self's target pixels.
```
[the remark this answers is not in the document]
rhugonnet: Rephrased as "Proximity to this Raster's target pixels, or to a Vector's geometry, computed for each cell of this Raster's grid."
Using "this Raster" or "this Vector" everywhere now.

geoutils/georaster/raster.py
```diff
@@ -2487,3 +2487,74 @@ def polygonize(
...
 
         return gv.Vector(gdf)
+
+    def proximity(
+        self,
+        vector: Vector | None = None,
+        target_values: list[float] | None = None,
+        geometry_type: str = "boundary",
+        in_or_out: str = "both",
+        distance_unit: str = "georeferenced",
```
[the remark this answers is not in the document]
rhugonnet: Done!

geoutils/geovector.py
```diff
@@ -338,15 +339,120 @@ def query(self: VectorType, expression: str, inplace: bool = False) -> VectorTyp
         new_vector.__init__(self.ds.query(expression))  # type: ignore
         return new_vector  # type: ignore
 
-    def buffer_without_overlap(self, buffer_size: int | float, plot: bool = False) -> np.ndarray:
+    def proximity(
+        self,
+        raster: gu.Raster = None,
```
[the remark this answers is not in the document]
rhugonnet: Added as it's still checked here (we've removed this `mpy` check in xDEM however, as the implicit version of it is supported now: https://github.com/GlacioHack/xdem/pull/306#:~:text=mypy%20now%20authorizes,definition%20and%20overload.)

geoutils/geovector.py
```diff
@@ -338,15 +339,120 @@ def query(self: VectorType, expression: str, inplace: bool = False) -> VectorTyp
         new_vector.__init__(self.ds.query(expression))  # type: ignore
         return new_vector  # type: ignore
 
-    def buffer_without_overlap(self, buffer_size: int | float, plot: bool = False) -> np.ndarray:
+    def proximity(
```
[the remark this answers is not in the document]
rhugonnet: Done! Took a bit of effort, especially avoiding circular imports :wink:, but it's more consistent this way, I like it!

geoutils/projtools.py
```diff
@@ -19,6 +19,76 @@
...
+        utm_zone = str(utm_nb).zfill(2) + "N"
+    elif -90 <= lat < 0:
+        utm_zone = str(utm_nb).zfill(2) + "S"
+    else:
+        raise ValueError("Latitude value is out of range [-90, 90[.")
+
+    return utm_zone
+
+
+def utm_to_epsg(utm: str) -> int:
```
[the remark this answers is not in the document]
rhugonnet: Yes, reinventing the wheel with bits of old code... I've replaced it.

At least writing the tests was useful, `pyproj` output is not the same format as some other software (single digit for UTM zones from 1-9).

geoutils/projtools.py
rhugonnet: Updated with this, thanks!!

tests/test_georaster.py
rhugonnet: Done!

tests/test_georaster.py
rhugonnet: Removed

geoutils/georaster/raster.py
```diff
@@ -2487,3 +2487,110 @@ def polygonize(
...
+        """
+
+        proximity = proximity_from_vector_or_raster(raster=self, vector=vector, target_values=target_values, geometry_type=geometry_type,
+                                                    in_or_out=in_or_out, distance_unit=distance_unit)
+
+        return self.copy(new_array=proximity)
+
+
+# -----------------------------------------
+# Additional stand-alone utility functions
```
adehecq, line 2524: At some point, we re-organized raster.py because the idea was to have the Raster class in a separate file than the other functions, see [the issue](https://github.com/GlacioHack/geoutils/issues/215).
Unfortunately, we never went all the way, because there is no other file in the georaster folder. spatial_tools.py most likely belongs there, but maybe it should be renamed and the functions double-checked. This is related to [this opened issue](https://github.com/GlacioHack/geoutils/issues/311).

geoutils/geovector.py
```diff
@@ -338,15 +339,99 @@ def query(self: VectorType, expression: str, inplace: bool = False) -> VectorTyp
...
+
+        :return: Buffered shapefile
+        """
+
+        from geoutils.projtools import latlon_to_utm, utm_to_epsg
+
+        # Get a rough centroid in geographic coordinates (ignore the warning that it is not the most precise):
+        with warnings.catch_warnings():
+            warnings.simplefilter(action="ignore", category=UserWarning)
+            shp_wgs84 = self.ds.to_crs(epsg=4326)
```
adehecq, line 406: since we're duplicating the dataset again below, should we close that one to reduce memory usage?

geoutils/geovector.py
```diff
@@ -338,15 +339,99 @@ def query(self: VectorType, expression: str, inplace: bool = False) -> VectorTyp
...
+            lat, lon = shp_wgs84.centroid.y.values[0], shp_wgs84.centroid.x.values[0]
+
+        # Get the EPSG code of the local UTM
+        utm = latlon_to_utm(lat, lon)
+        epsg = utm_to_epsg(utm)
+
+        # Reproject the shapefile in the local UTM
+        ds_utm = self.ds.to_crs(epsg=epsg)
```
adehecq, line 414: same here, maybe use the syntax `with` to close it after the following line?

tests/test_geovector.py
adehecq: Yes, is it intended to be empty?

tests/test_georaster.py
rhugonnet: Done with the name `equal_georeferenced_grid`, added tests and replaced by this where relevant!

tests/test_georaster.py
rhugonnet: Because of `flake8` warning on `prox3`, `prox4`, etc... not being used. I just removed them and simply ran the functions.

tests/test_geovector.py
rhugonnet: Forgot it, done now.

geoutils/geovector.py
```diff
@@ -338,15 +339,99 @@ def query(self: VectorType, expression: str, inplace: bool = False) -> VectorTyp
...
+        :param in_or_out: Compute proximity only 'in' or 'out'-side the polygon, or 'both'.
+        :param distance_unit: Distance unit, either 'georeferenced' or 'pixel'.
+
+        :return: Proximity raster.
+        """
+
+        # TODO: (specific to Vector proximity) raster size could passed as argument when
+        #  nothing else is passed (here and in create_mask/rasterize)
+
+        # 0/ If no Raster is passed, create one on the Vector bounds of size 1000 x 1000
```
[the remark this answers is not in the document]
rhugonnet: Added `grid_size` as an argument.

geoutils/geovector.py
rhugonnet: Done with contextual manager `with`

geoutils/geovector.py
rhugonnet: Done too

geoutils/georaster/raster.py
rhugonnet: Ah yes, that was why... Good point.
This will have to be done separately during re-structuration, especially when we address `spatial_tools.py` so I leave it out of this PR for the moment.

geoutils/geovector.py
rhugonnet: Actually, context managers are not supported for all `geopandas` objects/function, so I simply clear objects with `del` along the code.

rhugonnet:
Merging here as well!

PR #339, opened 2023-01-21
Fix bugs and write tests for `Raster.value_at_coords()`

### LINKED ISSUE #245: `Raster.value_at_coords()` cannot take a list of coordinates (contradicting its function signature)
**Describe the bug**
`Raster.value_at_coords()` has a function signature that suggests it can accept a list of coordinates. The following minimal example shows that this is not true.

**To Reproduce**
```python
>>> import geoutils as gu
>>> img = gu.Raster(gu.datasets.get_path("landsat_B4"))
>>> img.bounds
BoundingBox(left=478000.0, bottom=3088490.0, right=502000.0, top=3108140.0)
>>> img.value_at_coords([478020, 478040], [3088500, 3088500])
  File "<stdin>", line 1, in <module>
  File "/home/erik/Projects/ETH/GeoUtils/geoutils/georaster.py", line 1513, in value_at_coords
    col = int(col)
TypeError: int() argument must be a string, a bytes-like object or a number, not 'list'
```

**Expected behavior**
It should either work with lists (and arrays) or it should not state that it will work with lists! I would certainly prefer the former.

### LINKED ISSUE #324: Raster.value_at_coords fails if window=float
**Describe the bug**
When Raster.value_at_coords() is called with a float value as the input to `window`, it fails with a non-obvious error message.

**To Reproduce**

```python
----> 8     v = raster.value_at_coords(X, Y, latlon=True,  window=float(3))

File geoutils/georaster/raster.py:2057, in Raster.value_at_coords(self, x, y, latlon, band, masked, window, return_window, boundless, reducer_function)
   2054 window = rio.windows.Window(col, row, width, height)
   2056 if self.is_loaded:
-> 2057     data = self.data[slice(None) if band is None else band + 1, row : row + height, col : col + width]
   2058     value = format_value(data)
   2059     win: np.ndarray | dict[int, np.ndarray] = data

File ~/opt/miniconda3/envs/gee/lib/python3.8/site-packages/numpy/ma/core.py:3222, in MaskedArray.__getitem__(self, indx)
   3212 """
   3213 x.__getitem__(y) <==> x[y]
   3214 
   3215 Return the item described by i, as a masked array.
   3216 
   3217 """
   3218 # We could directly use ndarray.__getitem__ on self.
   3219 # But then we would have to modify __array_finalize__ to prevent the
   3220 # mask of being reshaped if it hasn't been set up properly yet
   3221 # So it's easier to stick to the current version
-> 3222 dout = self.data[indx]
   3223 _mask = self._mask
   3225 def _is_scalar(m):

TypeError: slice indices must be integers or None or have an __index__ method
```

**Expected behavior**

This function should only accept ints. It should fail if an int is not provided.
It may also be useful to detect if the number is round, e.g. 3.0, and automatically convert to int only in this case.

rhugonnet:
This PR adds tests for the `value_at_coords` function, adds the functionality of passing an array-like object of coordinates, and fixes minors bugs.

It looks like  more changes than there actually is in the "Files changed". Owing to shifting blocks of code in `value_at_coords` to be either in or out of the loop (added for array-like).

There's also a setup bug with `isort` that they recommend to fix by upgrading to `5.12.0` (will need the same in xDEM): moved to #343.

Resolves #324
Resolves #245

### DIFF
--- a/geoutils/georaster/raster.py
+++ b/geoutils/georaster/raster.py
@@ -2026,12 +2026,18 @@ np.ndarray or number and correct dtype, the compatible nodata value.
 
         """
         # Check for array-like inputs
-        if not isinstance(x, (float, np.floating, int, np.integer)) and isinstance(y, (float, np.floating, int, np.integer))\
-            or isinstance(x, (float, np.floating, int, np.integer)) and not isinstance(y, (float, np.floating, int, np.integer)):
-                raise TypeError("Coordinates must be of the same type.")
+        if (
+            not isinstance(x, (float, np.floating, int, np.integer))
+            and isinstance(y, (float, np.floating, int, np.integer))
+            or isinstance(x, (float, np.floating, int, np.integer))
+            and not isinstance(y, (float, np.floating, int, np.integer))
+        ):
+            raise TypeError("Coordinates must be both numbers or both array-like.")
 
         # If for a single value, wrap in a list
-        if isinstance(x, (float, np.floating, int, np.integer)) and isinstance(y, (float, np.floating, int, np.integer)):
+        if isinstance(x, (float, np.floating, int, np.integer)) and isinstance(
+            y, (float, np.floating, int, np.integer)
+        ):
             x = [x]
             y = [y]
             # For the end of the function
@@ -2039,7 +2045,7 @@ np.ndarray or number and correct dtype, the compatible nodata value.
         else:
             unwrap = False
             # Check that array-like objects are the same length
-            if len(x) != len(y):
+            if len(x) != len(y):  # type: ignore
                 raise ValueError("Coordinates must be of the same length.")
 
         # Check window parameter
@@ -2068,17 +2074,17 @@ np.ndarray or number and correct dtype, the compatible nodata value.
             list_windows = []
 
         # Loop over all coordinates passed
-        for k in range(len(x)):
+        for k in range(len(x)):  # type: ignore
 
             value: float | dict[int, float] | tuple[float | dict[int, float] | tuple[list[float], np.ndarray] | Any]
 
             if latlon:
                 from geoutils import projtools
 
-                x0, y0 = projtools.reproject_from_latlon((y[k], x[k]), self.crs)
+                x0, y0 = projtools.reproject_from_latlon((y[k], x[k]), self.crs)  # type: ignore
             else:
-                x0 = x[k]
-                y0 = y[k]
+                x0 = x[k]  # type: ignore
+                y0 = y[k]  # type: ignore
 
             # Convert coordinates to pixel space
             row, col = rio.transform.rowcol(self.transform, x0, y0, op=round)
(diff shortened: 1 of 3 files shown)


geoutils/georaster/raster.py
```diff
@@ -2027,11 +2025,38 @@ def value_at_coords(
...
         """
-        value: float | dict[int, float] | tuple[float | dict[int, float] | tuple[list[float], np.ndarray] | Any]
+        # Check for array-like inputs
+        if (
+            not isinstance(x, (float, np.floating, int, np.integer))
+            and isinstance(y, (float, np.floating, int, np.integer))
+            or isinstance(x, (float, np.floating, int, np.integer))
+            and not isinstance(y, (float, np.floating, int, np.integer))
+        ):
+            raise TypeError("Coordinates must be both numbers or both array-like.")
```
adehecq, line 2035: We should also check that they have the same number of elements if they are an array.

geoutils/georaster/raster.py
```diff
@@ -2027,11 +2025,38 @@ def value_at_coords(
...
+
+        # If for a single value, wrap in a list
+        if isinstance(x, (float, np.floating, int, np.integer)) and isinstance(
```
adehecq, line 2038: The second condition is redundant after the first check...

geoutils/georaster/raster.py
adehecq: Oh I see it's done later, perfect !

geoutils/georaster/raster.py
```diff
@@ -2043,63 +2068,93 @@ def format_value(value: Any) -> Any:
...
-            # Start reading at col,row and read 1px each way
-            width = 1
-            height = 1
+            if latlon:
+                from geoutils import projtools
 
-        # Make sure coordinates are int
-        col = int(col)
-        row = int(row)
+                x0, y0 = projtools.reproject_from_latlon((y[k], x[k]), self.crs)  # type: ignore
```
adehecq, line 2084: Ideally, this would be done outside of the loop, as it is probably more efficient with the array (as the looping is done by pyproj itself).

geoutils/georaster/raster.py
```diff
@@ -2043,63 +2068,93 @@ def format_value(value: Any) -> Any:
...
+            else:
+                x0 = x[k]  # type: ignore
+                y0 = y[k]  # type: ignore
+
+            # Convert coordinates to pixel space
+            row, col = rio.transform.rowcol(self.transform, x0, y0, op=round)
```
adehecq, line 2090: Same, I think this can work on a list directly, rather than looping...

tests/test_georaster.py
```diff
@@ -1291,26 +1291,105 @@ def test_interp(self) -> None:
...
+        """
+        Test that value at coords works as intended
+        """
 
-        r = gr.Raster(self.landsat_b4_path)
-        r2 = gr.Raster(self.landsat_b4_crop_path)
-        r.crop(r2)
+        # -- Tests 1: check based on indexed values --
+
+        # Open two rasters and crop the first one
```
adehecq, line 1300: Looks like the comment is not appropriate anymore?

adehecq (approved):
Thanks for updating this !
Just see my comments about avoiding some operations within the loop.

geoutils/georaster/raster.py
rhugonnet: Ah yes good catch, updated!

geoutils/georaster/raster.py
rhugonnet: Done

geoutils/georaster/raster.py
rhugonnet: Done

tests/test_georaster.py
rhugonnet: Updated

PR #342, opened 2023-02-01
Add support for `pathlib.Path` object as `filename` of Raster and Vector

### LINKED ISSUE #239: `Raster`/`Vector` instantiation does not accept `pathlib.Path` filenames
**Describe the bug**
`pathlib` makes handling of paths a little easier. This package is well established (and part of the stdlib), but it is not supported as a `filename` to a new `Raster`.

**To Reproduce**
```python
import geoutils as gu
import pathlib

path = pathlib.Path(gu.datasets.get_path("landsat_B4"))

img = gu.Raster(path)
```
This fails with the error: 
```
  File "/home/erik/temp/hello.py", line 6, in <module>
    img = gu.Raster(path)
  File "/home/erik/Projects/ETH/GeoUtils/geoutils/georaster.py", line 251, in __init__
    raise ValueError("filename argument not recognised.")
ValueError: filename argument not recognised.
```

**Expected behavior**
All different `pathlib` path objects should be supported (they are by rasterio, so this should be an easy fix).

 - OS: Linux
 - Environment : latest main

rhugonnet:
Resolves #239 and adds tests.

Additionally, updates the list of resampling algorithms of `rasterio` from `rio.warp.Resampling` to `rio.enums.Resampling`, and gets the names using `.name` instead of `str()`, which fails in Python `3.11` (opened issue: https://github.com/rasterio/rasterio/issues/2750).

### DIFF
--- a/geoutils/georaster/raster.py
+++ b/geoutils/georaster/raster.py
@@ -5,13 +5,13 @@ from __future__ import annotations
 
 import math
 import os
+import pathlib
 import warnings
 from collections import abc
 from collections.abc import Iterable
 from contextlib import ExitStack
 from numbers import Number
 from typing import IO, Any, Callable, TypeVar, overload
-import pathlib
 
 import geopandas as gpd
 import matplotlib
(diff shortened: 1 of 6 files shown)


rhugonnet:
Merging to clean the diffs of the 6 PRs branched on this one (was necessary to run tests). Can account for comments in another PR.

adehecq (commented):
All good ! 😄

PR #344, opened 2023-02-02
Fix for `Raster.reproject()` and `stack_rasters` on multiple bands

### LINKED ISSUE #242: `Raster.reproject()` fails on 4-band rasters
**Describe the bug**
When `reproject()` is run on a 4-band raster (in my case RGB + mask), it fails. This is an easy fix, however.

**To Reproduce**
```python
import geoutils as gu
import rasterio as rio
import numpy as np

img1 = gu.Raster.from_array(
        np.ones((4, 500, 500), dtype="uint8"),
        transform=rio.transform.from_origin(0, 500, 1, 1),
        crs=4326
)

img2 = gu.Raster.from_array(
        np.ones((4, 500, 500), dtype="uint8"),
        transform=rio.transform.from_origin(50, 500, 1, 1),
        crs=4326
)

img2.reproject(img1)


```
This raises the error:
```
  File "/home/erik/temp/hello.py", line 19, in <module>
    img2.reproject(img1)
  File "/home/erik/Projects/ETH/GeoUtils/geoutils/georaster.py", line 1022, in reproject
    dst_data, dst_transformed = rio.warp.reproject(self.data, **reproj_kwargs)
  File "/home/erik/.local/share/conda/geoutils/lib/python3.9/site-packages/rasterio/env.py", line 387, in wrapper
    return f(*args, **kwds)
  File "/home/erik/.local/share/conda/geoutils/lib/python3.9/site-packages/rasterio/env.py", line 610, in wrapper
    return f(*args, **kwds)
  File "/home/erik/.local/share/conda/geoutils/lib/python3.9/site-packages/rasterio/warp.py", line 368, in reproject
    _reproject(
  File "rasterio/_warp.pyx", line 473, in rasterio._warp._reproject
  File "rasterio/_warp.pyx", line 429, in rasterio._warp._reproject
ValueError: Invalid destination shape
```

**Expected behavior**
This should not fail. It's a simple and easy to fix bug.

**Additional context**
The line that messes things up is:
https://github.com/GlacioHack/GeoUtils/blob/314764147b84e1680868d512f1195feda46e094c/geoutils/georaster.py#L984

If this is changed to:
```python
dst_data = np.ones((self.nbands, dst_size[1], dst_size[0]), dtype=dtype) 
```
... everything works as expected!

Makes me wonder, however, why that array is even necessary?? It's given as the `"destination"` argument to `rio.warp.reproject()`, but it's not used after that (the returned array from the call is used). Are these the same objects, only with different names? Or is the entire array duplicated??

We should validate that `destination` and `dst_data` is actually the same array:
https://github.com/GlacioHack/GeoUtils/blob/314764147b84e1680868d512f1195feda46e094c/geoutils/georaster.py#L1022

rhugonnet:
Resolves #242

For `merge_rasters` and `stack_rasters`: I forced the function to only work on the first band. Right now `stack_rasters` returns a `Raster` object, so it is impossible to stack/merge 3D Rasters (because `Raster` does not support 4D). We'd have to change the output to be a `list`, or `xarray.Dataset` (I'm not sure that's priority).

### DIFF
--- a/tests/test_spatial_tools.py
+++ b/tests/test_spatial_tools.py
@@ -182,16 +182,26 @@ def images_3d():  # type: ignore
 
 
 @pytest.mark.parametrize(
-    "rasters", [pytest.lazy_fixture("images_1d"), pytest.lazy_fixture("sat_images")]
+    "rasters", [pytest.lazy_fixture("images_1d"), pytest.lazy_fixture("sat_images"), pytest.lazy_fixture('images_3d')]
 )  # type: ignore
-#    pytest.lazy_fixture('images_3d')]) ## Requires Raster.reproject() fix.
 def test_stack_rasters(rasters) -> None:  # type: ignore
     """Test stack_rasters"""
 
-    warnings.filterwarnings("ignore", category=UserWarning)
+    # Silence the reprojection warning for default nodata value
+    warnings.filterwarnings("ignore", category=UserWarning, message="New nodata value found in the data array.*")
+    warnings.filterwarnings("ignore", category=UserWarning, message="For reprojection, dst_nodata must be set.*")
 
     # Merge the two overlapping DEMs and check that output bounds and shape is correct
-    stacked_img = gu.spatial_tools.stack_rasters([rasters.img1, rasters.img2])
+    if rasters.img1.count > 1:
+        # Check warning is raised once
+        with pytest.warns(expected_warning=UserWarning, match="Some input Rasters have multiple bands, only their first band will be used."):
+            stacked_img = gu.spatial_tools.stack_rasters([rasters.img1, rasters.img2])
+        # Then ignore the other ones
+        warnings.filterwarnings("ignore", category=UserWarning, message="Some input Rasters have multiple bands, only their first band will be used.")
+
+    else:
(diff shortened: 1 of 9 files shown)


tests/test_spatial_tools.py
```diff
@@ -227,24 +243,34 @@ def test_stack_rasters(rasters) -> None:  # type: ignore
...
 
-    assert rasters.img.data.shape == merged_img.data.shape
+    assert rasters.img.shape == merged_img.shape
     assert rasters.img.bounds == merged_img.bounds
     assert np.count_nonzero(np.isnan(merged_img.data)) == 0  # Check no NaNs introduced
 
     diff = rasters.img.data - merged_img.data
 
-    assert np.abs(np.nanmean(diff)) < 0.3
+    assert np.abs(np.nanmean(diff)) < 1
```
adehecq, line 273: Why was this not working with the previous threshold? Should it not be exactly equal?

rhugonnet: The 3 band RGB raster has different values that were not tested until now, and its residuals with this calculation are a tiny bit larger than 0.3.

PR #346, opened 2023-02-03
Use consistently `count` and `indexes` for band listing and counting (removing `nbands` and  `bands`)

### LINKED ISSUE #280: Fix the mess in attributes nbands/count/bands/indexes
See the discussion in PR #265 about the craziness in attributes count vs nbands, bands vs indexes. 
Here is my suggestion:

I think we should stay consistent with rasterio's naming convention and hence have `count` and `indexes` match the definition of rasterio:
- `count`: band count on disk
- `indexes`: The 1-based indexes of each band in the dataset

We can have an equivalent of each for bands loaded, i.e. `count_loaded` and `indexes_loaded` (not sure the latter is really needed...).
What do you think @erikmannerfelt, @atedstone @rhugonnet ? 
Others should be deprecated.

_Originally posted by @adehecq in https://github.com/GlacioHack/GeoUtils/pull/265#discussion_r953810255_

rhugonnet:
To be fully consistent with `rasterio`.

Resolves #280

### DIFF
--- a/geoutils/georaster/raster.py
+++ b/geoutils/georaster/raster.py
@@ -209,10 +209,10 @@ class Raster:
             The path/filename of the loaded, file, only set if a disk-based file is read in.
         data : np.array
             Loaded image. Dimensions correspond to (count, height, width).
-        count : int
+        count_loaded : int
             Number of bands loaded into .data
-        indexes : tuple
-            The indexes of the opened dataset which correspond to the bands loaded into data.
+        indexes_loaded : tuple
+            The indexes of the bands of the opened dataset loaded into .data.
         is_loaded : bool
             True if the image data have been loaded into this Raster.
 
@@ -284,6 +284,7 @@ class Raster:
         self._data: np.ma.masked_array | None = None
         self._nodata: int | float | list[int] | list[float] | None = nodata
         self._indexes = indexes
+        self._indexes_loaded = None
         self._masked = masked
         self._disk_hash: int | None = None
         self._is_modified = True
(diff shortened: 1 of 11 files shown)


geoutils/georaster/raster.py
```diff
@@ -2331,19 +2318,19 @@ def interp_points(
         #         rpts.append(zint)
         # rpts = np.array(rpts)
 
-    def split_bands(self: RasterType, copy: bool = False, subset: list[int] | int | None = None) -> list[Raster]:
+    def split_indexes(self: RasterType, copy: bool = False, subset: list[int] | int | None = None) -> list[Raster]:
```
adehecq, line 2321: I think in this case, it makes more sense to keep "band" as it is much clearer what this will do. To me, indexes refer rather to the position of the bands, so it does not make sense to split them.

adehecq (commented):
Hopefully it will be much clearer now! 
But what I don't get is that basically here you just removed nbands into count and bands into indexes. But I thought we had a different use of count and nbands, depending on what was loaded and what was on disk, no? 
In the issue, I suggested using a `count_loaded` and `indexes_loaded` but it seems like it was not needed, why?

geoutils/georaster/raster.py
rhugonnet: Agree, changed

rhugonnet:
My bad, I forgot that point! I added them back and added tests for them.

The names `count_loaded` and `indexes_loaded` sounded good. But then `count` and `index` are "on-disk", and can take a value of `None`, which is not so practical when manipulating in-memory-only `Raster`. So instead I went for `count_on_disk` and `indexes_on_disk`, while `count` and `indexes` are in-memory. Clarified everything in the descriptions.

adehecq:
> My bad, I forgot that point! I added them back and added tests for them.
> 
> The names `count_loaded` and `indexes_loaded` sounded good. But then `count` and `index` are "on-disk", and can take a value of `None`, which is not so practical when manipulating in-memory-only `Raster`. So instead I went for `count_on_disk` and `indexes_on_disk`, while `count` and `indexes` are in-memory. Clarified everything in the descriptions.

Sounds good. Everything seems perfect now, thanks !

PR #348, opened 2023-02-03
Change `xy2ij` behaviour to consider `AREA_OR_POINT` attribute as interpretation

### LINKED ISSUE #341: A suspicious bug on raster.xy2ij
**Describe the bug**

I have a short test on the raster's function of xy2ij, and ij2xy below:

```
x_1,y_1 = raster.ij2xy(i_1, j_1)
i_2, j_2 = raster.xy2ij(x_1, y_1)
```

The results show i_1, j_1 != i_2, j_2, which means **the two functions are not reversible**.

**To Reproduce**

Here is a minimal test function:

```
import numpy as np
import dem
dem = xdem.DEM('test_dem.tif')

# i,j = 0,0 on both cases

# case 1

x_0,y_0 = dtm.ij2xy(0, 0, offset='center')
dem.xy2ij(x_0,y_0,op=np.float32, area_or_point='Point')
-> (array([1.], dtype=float32), array([1.], dtype=float32))

# case 2

x_0,y_0 = dtm.ij2xy(0, 0, offset='ul')
dem.xy2ij(x_0,y_0,op=np.float32,area_or_point='Area')
-> (array([0], dtype=float32), array([0], dtype=float32)
```
**Expected behavior**

I expected the two functions to be reversible on both case 1 and 2

However, the 'xy2ij' has a problem of turning xy back to ij **when offset='center'.**

When offset='ul' and area_or_point='Area', it works as expected.

 - OS: Window 10
 - Environment: geoutils 0.0.9, xdem 0.0.7

**Additional context**

The main issue is how to turn the index of rows and columns into ij when area_or_point='Point'.

It seems like lines 2202/2203 at https://github.com/GlacioHack/geoutils/blob/main/geoutils/georaster/raster.py should be:

        i -= 0.5
        j -= 0.5

rhugonnet:
This PR changes `xy2ij` so that it never shifts the coordinates of pixels based on `self.tags['AREA_OR_POINT']` by default. Now, `ij2xy` and `xy2ij` are reversible.

Shifting with the interpretation of `self.tags['AREA_OR_POINT']` remains a possibility however (as it could be needed during interpolation of points, for instance), and is performed by setting the `shift_area_or_pixel` as `True` (which then reads the attribute from `self.tags`).

Some tests added also.

Resolves #341

### DIFF
--- a/tests/test_georaster.py
+++ b/tests/test_georaster.py
@@ -1664,14 +1664,14 @@ self.set_nodata()."
         temp_dir = tempfile.TemporaryDirectory()
 
         # Save file to temporary file, with defaults opts
-        with NamedTemporaryFile(mode="w", delete=False, dir=temp_dir.name) as temp_file:
-            img.save(temp_file.name)
-            saved = gr.Raster(temp_file.name)
-            assert img == saved
+        temp_file = NamedTemporaryFile(mode="w", delete=False, dir=temp_dir.name)
+        img.save(temp_file.name)
+        saved = gr.Raster(temp_file.name)
+        assert img == saved
 
-            # Try to save with a pathlib path
-            path = pathlib.Path(temp_file.name)
-            img.save(path)
+        # Try to save with a pathlib path
+        path = pathlib.Path(temp_file.name)
+        img.save(path)
 
         # Test additional options
         co_opts = {"TILED": "YES", "COMPRESS": "LZW"}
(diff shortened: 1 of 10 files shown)


tests/test_georaster.py
```diff
@@ -1247,7 +1271,7 @@ def test_interp(self) -> None:
             list_z_ind.append(z_ind)
 
         # First order interpolation
-        rpts = r.interp_points(pts, order=1, area_or_point="Area")
+        rpts = r.interp_points(pts, order=1, shift_area_or_point=True)
```
adehecq, line 1274: Why is `shift_area_or_point` set to True here? If it was "Area" in the previous implementation, it should not be shifted, right?

adehecq (changes requested):
Can you also add a test to check that xy2ij and ij2xy are reversible, like in the issue raised?

tests/test_georaster.py
rhugonnet: Yes, not needed! Removed.

rhugonnet:
Done!

PR #349, opened 2023-02-03
Add `Vector.reproject()`

### LINKED ISSUE #337: Wrap `geopandas` (and specifically `shapely`) functionalities in the `Vector` class
To avoid having to write something like:
```python
boundary_vec = Vector(geopandas.GeoDataFrame(geometry=vector.ds.boundary, crs=vector.crs)
```
for every manipulation.

We should add functionalities to the `Vector` class to apply any attribute matching that of `Shapely` to the `GeoSeries`, turn it into a GeoDataFrame (the main question being: what metadata should be kept from the original one?) and put that in a new `Vector` object.

_Originally posted by @rhugonnet in https://github.com/GlacioHack/geoutils/pull/336#discussion_r1070747879_

### LINKED ISSUE #340: Add a Vector.reproject method
To make the Raster and Vector classes more consistent, we should have a Vector.reproject method that simply reproject from one CRS to another.
It comes down to basically one line with GeoPandas but it would be more consistent.

rhugonnet:
Resolves #340

### DIFF
--- a/geoutils/geovector.py
+++ b/geoutils/geovector.py
@@ -3,11 +3,11 @@ geoutils.vectortools provides a toolset for working with vector data.
 """
 from __future__ import annotations
 
+import pathlib
 import warnings
 from collections import abc
 from numbers import Number
 from typing import Literal, TypeVar, overload
-import pathlib
 
 import geopandas as gpd
 import matplotlib.pyplot as plt
--- a/geoutils/geovector.py
+++ b/geoutils/geovector.py
@@ -7,6 +7,7 @@ import warnings
 from collections import abc
 from numbers import Number
 from typing import Literal, TypeVar, overload
+import pathlib
 
 import geopandas as gpd
 import matplotlib.pyplot as plt
@@ -30,7 +31,7 @@ class Vector:
     Create a Vector object from a fiona-supported vector dataset.
     """
 
-    def __init__(self, filename: str | gpd.GeoDataFrame):
+    def __init__(self, filename: str | pathlib.Path | gpd.GeoDataFrame):
         """
         Load a fiona-supported dataset, given a filename.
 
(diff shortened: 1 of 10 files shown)


geoutils/geovector.py
```diff
@@ -146,6 +155,62 @@ def crop(
             new_vector.ds = new_vector.ds.cx[xmin:xmax, ymin:ymax]
             return new_vector
 
+    def reproject(
+        self: Vector,
+        dst_ref: gu.Raster | rio.io.DatasetReader | VectorType | gpd.GeoDataFrame | str | None = None,
+        dst_crs: CRS | str | int | None = None,
+    ) -> Vector:
+        """
+        Reproject vector to a specified CRS and, optionally, cropped to certain bounds (no cropping by default).
```
adehecq, line 164: I don't understand, there is no cropping involved in this function... All that's done is to get the CRS from that reference.
I think we should go for something simple for now and only have the `dst_crs` argument. It's easy to provide it with a Raster or Vector, as the CRS is an attribute.
Or did you have something else in mind?

rhugonnet: Ah yes, removed the cropping in the description!
I simply wanted to mirror `Raster.reproject()`: you can provide a `dst_ref` or `dst_crs`. To have as many features consistent between `Vector` and `Raster` as possible, and keep the general policy of all our other functions of "accepting any raster or vector as inputs", so as to avoid manipulating the georeference metadata.

rhugonnet: It's a detail (removing 2 lines or not), so I'll merge to advance PR #350 without merging issues.
We can discuss it further at a monthly meeting if the above was not convincing :wink: .

adehecq: It's not 2 lines, but almost 30 lines of code (and a few more of docstring) !
I'm in favor of simplicity here, especially for maintenance, tests etc. Especially since we don't know how long these tools will be useful...

PR #350, opened 2023-02-07
Add `Mask` subclass

### LINKED ISSUE #316: Decide of the naming of strict or relaxed equality (`__eq__` or `raster_equal`?)
See https://github.com/GlacioHack/geoutils/pull/313#discussion_r975421530
Right now `__eq__` checks that `data.data` and `data.mask` (and other attributes) are equal, and `np.ma.allequal` can be used to compare `.data` with masked values.
Several solutions we could implement:

- We could add `np.array_equal` to `__array_function__` to be castable on a `Raster`;
- We could add a `gu.raster_equal` function that accounts potentially for nodata with an argument;
- For equality with `__eq__`, what do we do? Should we mirror behaviour from NumPy or not for this? Actually NumPy broadcasts to element-wise equality with `__eq__`.

Example copied from @adehecq on Slack:
```python
import numpy as np
data = np.ma.masked_array(np.arange(9).reshape((3,3)))
data[0, 0] = np.ma.masked
data2 = data.copy()
data2.data[0, 0] = 10
data == data2. # returns True for all unmasked values, masked otherwise
np.all(data == data2) # True
np.ma.allequal(data, data2) # True
```

rhugonnet:
This PR:

- Adds the `Mask` class,
- Casts any boolean array passed to `Raster.from_array()` into a `Mask`,
- Makes `__array_function__` interface compatible with logical functions, automatically cast into a `Mask`
- Changes output of `Vector.create_mask()` and `rasterize()` with certain parameters to `Mask`,
- Changes `__eq__` into `raster_equal()`.

Write new tests for all.

Resolves #316

### DIFF
--- a/geoutils/__init__.py
+++ b/geoutils/__init__.py
@@ -4,7 +4,7 @@ GeoUtils is a python package of raster and vector tools.
 
 from geoutils import spatial_tools  # noqa
 from geoutils import examples, georaster, geovector, projtools, satimg  # noqa
-from geoutils.georaster import Raster, Mask  # noqa
+from geoutils.georaster import Mask, Raster  # noqa
 from geoutils.geovector import Vector  # noqa
 from geoutils.satimg import SatelliteImage  # noqa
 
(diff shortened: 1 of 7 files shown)


rhugonnet:
Ready for review, just need to make `mypy` happy and it's all done.

rhugonnet:
Closing this, as it had to be developed directly in #351. See there for the final `Mask` features and tests.

PR #353, opened 2023-02-18
Fix cbar alignment in `Raster.show`

friedrichknuth:
This PR fixes the color bar alignment with the main plot when using `Raster.show()` and tests that the cbar and main plot heights are approximately equal. 

Additional changes:
- Add alpha parameter for plot and cbar to `Raster.show()`
- Rename `cb` to `cbar` to match matplotlib conventions.


rhugonnet (approved):
Amazing timing for the doc :stuck_out_tongue_winking_eye:

PR #363, opened 2023-04-16
Minor fixes in new documentation

### LINKED ISSUE #362: Revert `default_mode` of `html_context` to `auto` once Pydata Sphinx Theme is updated
After losing a couple hours on this, the logo problem actually didn't come from our file (SVG is the standard vector format for Sphinx, and ours was fine). 
It was an issue with the Pydata Sphinx Theme when light/dark mode was defined as "auto": https://github.com/pydata/pydata-sphinx-theme/issues/1180

_Originally posted by @rhugonnet in https://github.com/GlacioHack/geoutils/issues/351#issuecomment-1507568256_

rhugonnet:
This PR:
- Fixes the logo display on Firefox by using a PNG (couldn't find where the error originated using an SVG, despite digging quite deep...),
- Fixes the `_repr_html_` function in case `self._out_shape` is `None`

Resolves #362


rhugonnet:
Merging as it's only a couple lines + change from SVG to PNG, and our logo did not display otherwise in Firefox :cry:

PR #364, opened 2023-04-19
Fix GitHub Actions mamba caching

erikmannerfelt:
See GlacioHack/xdem#354 and GlacioHack/xdem#353 for the problem and the solution.


rhugonnet (approved):
Perfect!

PR #367, opened 2023-04-25
Fix `coverage` failure

rhugonnet:
We have a failing of coverage push to Coveralls on the main branch (dropped to 0%). While it wasn't the case on other PRs. Can't seem to be able to fix it by rerunning GitHub Actions of the main branch.

Let's try here.


rhugonnet:
Looks like it works from the fork branch, I'll merge to main to see if it solves the issue or if it is something else.

PR #368, opened 2023-04-26
Enable building the doc in offline formats #363 

adehecq:
All in the title. Same as xdem PR [363](https://github.com/GlacioHack/xdem/pull/363).


rhugonnet:
You need to correct `hmtlzip` into `htmlzip` (I did it in xDEM in a separate PR, tests cannot cover RTD so it can only be checked manually on the RTD doc)

rhugonnet:
Doing that to rerelease, see https://github.com/GlacioHack/xdem/issues/361

rhugonnet (approved):

adehecq:
> You need to correct `hmtlzip` into `htmlzip` (I did it in xDEM in a separate PR, tests cannot cover RTD so it can only be checked manually on the RTD build)

Haha good catch ! I also just fixed the issue with my fork that diverged due to merges. So we could have actually closed this PR and I could have rewritten a new one.

PR #380, opened 2023-08-07
Fix save with "uint8" `dtype` for `Mask` class

### LINKED ISSUE #379: When saving mask, fails due to dtype = 'bool'
**Describe the bug**
Saving a mask fails. According to the documentation (https://geoutils.readthedocs.io/en/stable/mask_class.html), I think it should automatically convert the bool dtype to numpy.uint8.

**To Reproduce**
```
outlines_nlcd = gu.Vector(path_to_nlcd)
inlier_mask_nlcd = outlines_nlcd.create_mask(reference_demr)
inlier_mask_all = nlier_mask_nlcd
```

**Expected behavior**
The mask saves as numpy.uint8 dtype, and can be viewed in black/white.

 - OS: macos Ventura
- geoutils v0.0.12

image

rhugonnet:
Trying to properly force this behaviour for the `Mask` class.

Resolves #379

### DIFF
--- a/geoutils/raster/raster.py
+++ b/geoutils/raster/raster.py
@@ -2189,7 +2189,6 @@ np.ndarray or number and correct dtype, the compatible nodata value.
 
         :returns: None.
         """
-        dtype = self.data.dtype if dtype is None else dtype
 
         if co_opts is None:
             co_opts = {}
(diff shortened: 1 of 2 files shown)


geoutils/raster/raster.py
```diff
@@ -2189,7 +2189,6 @@ def save(
 
         :returns: None.
         """
-        dtype = self.data.dtype if dtype is None else dtype
```
adehecq, line 2192: Why did you need to delete this line?

rhugonnet: Because `dtype` was never used later in the function (`mypy` strangely didn't pick that up but some of my other IDE linters did)

adehecq: Haha ok... In that case, it should be removed from the argument list. Or we could use the argument to force the output type to a preferred type. I will raise an issue.

PR #382, opened 2023-08-08
Update with NumPy 1.25

### LINKED ISSUE #381: Fix issues arising with NumPy 1.25
Issues that I found for now:
- A `TypeError` is not raised anymore during arithmetic tests of `ufuncs` in `test_raster.py` with types not matching: https://github.com/GlacioHack/geoutils/blob/18e0d11635eb2c6ab214ce5bcb4d3bffa9687620/tests/test_raster.py#L3561
- The function `np.find_common_types` is now deprecated.

rhugonnet:
In short:
- Replacing `np.find_common_dtypes` that is deprecated: to get the same behaviour, now using `np.promote_dtypes` in a `try`/`except` to catch promotion errors with datetimes and return an object dtype (see https://numpy.org/devdocs/release/1.25.0-notes.html#deprecations),
- Ignoring `int64` and `uint64` casting rules in `ufunc` tests, which were recently added.

Resolves #381

### DIFF
--- a/dev-environment.yml
+++ b/dev-environment.yml
@@ -7,7 +7,7 @@ dependencies:
   - matplotlib
   - pyproj
   - rasterio>=1.3
-  - numpy=1.24
+  - numpy
   - scipy
   - tqdm
   - xarray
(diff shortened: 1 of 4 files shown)


adehecq (commented):
👍

PR #385, opened 2023-08-09
Refine typing with NumPy plugin of `mypy`

### LINKED ISSUE #291: Deprecate `_typing` and enforce NumPy>1.20
`ArrayLike` and `DTypeLike` were added in `numpy.typing` in NumPy 1.20, that was released already a year and a half ago and is supported on Python 3.7 to 3.9.
As we are using Python>3.7 (soon >3.8), NumPy version will always be >1.20, so we can deprecate the `_typing` module soon (implemented in #227).

### LINKED ISSUE #319: Add NumPy plugin to `mypy` to specify array dtypes
To mirror the changes made in xDEM in https://github.com/GlacioHack/xdem/pull/306

rhugonnet:
This PR adds the NumPy plugin for `mypy` type checking.

NumPy arrays now have to be statically typed with certain dtypes and shapes (https://numpy.org/devdocs/reference/typing.html).
To keep it simple, I ignored the shape and only defined types (like was done in xDEM, still raised more than 100 errors on `mypy`) and introduced the following types for GeoUtils:
- `NDArrayNum` is a numerical ND array, i.e. with floating or integer type,
- `NDArrayBool` is a boolean ND array,
- `MArrayNum` is a masked numerical ND array,
- `MArrayBool` is a masked boolean ND array.

We can change those names if they don't feel like the best, it's easy as it's just a refactor. Eventually, we could even add the shape to have even more robust type-check in all functions (both here and in xDEM).

**Note: using an IDE, you don't see those names in function definition but only the corresponding detailed NumPy types, for examples `np.ndarray[Any, np.floating[Any]]`; hopefully it'll be the same in the documentation, so the names don't matter that much except for developers!**

Everything is defined in a way to be backwards-compatible during package import. I also redefined the `Number` type (Python `float` or `int` or NumPy `floating` or `integer`) to facilitate the typing in functions.

Resolves #291
Resolves #319

Some `mypy` issues were unsolvable without a major class change between `Raster` and `Mask`: temporarily ignored them and opened #387.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -48,7 +48,9 @@ repos:
           hooks:
                 - id: mypy
                   args: [
+                        --config-file=mypy.ini,
                         --strict,
+                        --implicit-optional,
                         --ignore-missing-imports,  # Don't warn about stubs since pre-commit runs in a limited env
                         --allow-untyped-calls,  # Dynamic function/method calls are okay. Untyped function definitions are not okay.
                         --show-error-codes,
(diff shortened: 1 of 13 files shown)


adehecq:
I'll try to remember next time I write code 😉

Issue #373, opened 2023-04-28
Force `geopandas > 0.12.0` to support `normalize`


rhugonnet:
See https://github.com/GlacioHack/xdem/issues/360#issuecomment-1527735925

erikmannerfelt:
This is still not fixed with pip installs (setup.py) and in the geoutils conda-forge feedstock. Reopening.

PR #388, opened 2023-08-14
Update setup.py requirements to geopandas >= 0.12.0

erikmannerfelt:
The minimum geopandas version was recently updated to 0.12.0. As mentioned in my comment in  #373, however, this was not written in the setup.py (meaning pip installs didn't work properly) nor in the conda-forge feedstock. This fixes the `setup.py` part, and I'll open a PR in the feedstock.


adehecq (approved):
Good point !

PR #393, opened 2023-08-16
Add `np.abs` and `np.absolute` to supported NumPy array functions in `Raster`

rhugonnet:
Helps on https://github.com/GlacioHack/xdem/issues/404

### DIFF
--- a/geoutils/raster/raster.py
+++ b/geoutils/raster/raster.py
@@ -99,6 +99,8 @@ _HANDLED_FUNCTIONS_1NIN = (
         "cumsum",
         "cumprod",
         "quantile",
+        "abs",
+        "absolute"
     ]
     + ["sort", "count_nonzero", "unique"]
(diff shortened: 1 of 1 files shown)


erikmannerfelt (approved):
Extravagant!

erikmannerfelt:
Oh by the way, why are there multiple lists like that? Not for this PR, but it could be simplified to just one list, and one comprehension perhaps that adds "nan" suffixes to some of them.

rhugonnet:
It's because they are from different NumPy submodules (logical: https://numpy.org/doc/stable/reference/routines.logic.html, linear algebra: https://numpy.org/doc/stable/reference/routines.linalg.html, mathematical functions: https://numpy.org/doc/stable/reference/routines.math.html, etc...) and only a few of them have a "NaN" version.

I could add a comment with the link to NumPy each time to clarify?

rhugonnet:
Comments added to make the list organization clear!

PR #392, opened 2023-08-16
Backward-compatiblity with Shapely < 2.0

rhugonnet:
Before the major refactoring of Shapely a year ago, `shapely.Geometry` did not exist. But we've been using for (static and dynamic) type checking in functions during the past year. It all passed in CI, as older versions of Shapely were never used.

As it's just about editing 4/5 types, I think we modify them all to the old `shapely.geometry.base.BaseGeometry` syntax so that we can be backwards-compatible for a bit longer! :slightly_smiling_face: 

Resolves #391

PR #405, opened 2023-09-28
Improve backward compatibility for shapely < 2.0

adehecq:
Improve on #392 after finding new errors with shapely=1.8.5.


rhugonnet (approved):
Nice! :)

Issue #409, opened 2023-10-20
Reproject fails with error when self.res contains numpy floats


adehecq:
**Describe the bug**
When running `Raster.reproject` on an object for which the items in `self.res` are numpy.float64 (as opposed to float), it fails at [this line](https://github.com/GlacioHack/geoutils/blob/main/geoutils/raster/raster.py#L2103) with "ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()"
This is because the test `dst_res == self.res[0] == self.res[1]` behaves differently. With float, it returns a single boolean, for np.float64, it returns a tuple of booleans.

**To Reproduce**
As `self.res` cannot be set manually, the best way to reproduce (and the way I came across the bug) is to load with downsampling:
```
import geoutils as gu
rst = gu.Raster(gu.examples.get_path("everest_landsat_b4"))
print(type(rst.res[0]))
# -> float

rst = gu.Raster(gu.examples.get_path("everest_landsat_b4"), downsample=2)
print(type(rst.res[0]))
# -> numpy.float64

rst.reproject(dst_res=(20, 20))
# -> ValueError
```

**Expected behavior**
The function should proceed without error.

**System (please complete the following information):**
 Irrelevant

Issue #407, opened 2023-10-11
Raster.show() does not always function as expected


atedstone:
**Describe the bug**
Sometimes Raster.show() fails to identify an appropriate axis/figure on which to plot:

```python
File ~/opt/miniconda3/envs/ee/lib/python3.11/site-packages/geoutils/raster/raster.py:2479, in Raster.show(self, index, cmap, vmin, vmax, alpha, cbar_title, add_cbar, ax, return_axes, **kwargs)
   2476         ax0 = plt.gca()
   2477     # Otherwise, get first axis
   2478     else:
-> 2479         ax0 = plt.gcf().axes[0]
   2480 elif isinstance(ax, str) and ax.lower() == "new":
   2481     _, ax0 = plt.subplots()

IndexError: list index out of range
```

**To Reproduce**

1. In a JupyterLab notebook, use `%matplotlib widget`
2. `myraster.show()`
3. `plt.figure()`
4. `myraster.show()` - fails

**Expected behavior**
The second `.show()` operation should plot the image in the new figure.

**System (please complete the following information):**
 - OS: OSX
 - Environment: geoutils 0.0.11, Python 3.11

adehecq:
Here is a MWE:
```
import geoutils as gu
import matplotlib.pyplot as plt
rst = gu.Raster(gu.examples.get_path("everest_landsat_b4"))
plt.figure()
rst.show()
```
-> fails with "IndexError: list index out of range"

adehecq:
Andrew, a quick fix to that is to run `fig, ax = plt.subplots()` instead of `plt.figure()`.

PR #410, opened 2023-10-20
Fix for issue #409 and #407

adehecq:
- solves #407 
- solves #409

### DIFF
--- a/geoutils/raster/raster.py
+++ b/geoutils/raster/raster.py
@@ -2514,7 +2514,12 @@ np.ndarray or number and correct dtype, the compatible nodata value.
 
         # Create axes
         if ax is None:
-            ax0 = plt.gca()
+            # If no figure exists, get a new axis
+            if len(plt.get_fignums()) == 0:
+                ax0 = plt.gca()
+            # Otherwise, get first axis
+            else:
+                ax0 = plt.gcf().axes[0]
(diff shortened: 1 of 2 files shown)


rhugonnet:
A bit afraid that the plotting changes will break examples in the documentation, as this is why this code was modified in the first place. Before merging, we should check how the documentation renders on your fork @adehecq (you need to go to readthedocs and build the branch on your fork, can give more details if you haven't done that before)

adehecq:
Right now, I fail building the documentation on readthedocs for this branch... But this does not seem to be related to the change to plots are undoing it did not solve the issue.

rhugonnet:
Strange, because the documentation of the `main` branch is still passing (restarted a build today to try): https://readthedocs.org/projects/geoutils/builds/22932914/

adehecq:
Ok, now the doc compiled (I don't know why it was not working before but now it does...).
Here is the link: https://geoutils-adehecq.readthedocs.io/en/amaury_fixes/index.html.
@rhugonnet, where did you expect the changes to break the doc? I looked at a few figures and they appeared as expected.

rhugonnet:
Indeed, I couldn't find any figure that looked different! :partying_face:
But let's remember to check every page of the doc before we release a `0.1` after the hackathon!

rhugonnet (approved):
Great, thanks for looking into this!

PR #351, opened 2023-02-10
Write documentation and streamline general behaviour

rhugonnet:
## Summary

This PR writes the documentation for GeoUtils! :partying_face: 

It also:
- Streamlines several features and behaviours to better fit the objective of the package,
- Re-organizes the structure of several modules,
- Homogenizes the syntax of the docstrings,
- Interfaces GeoPandas/Shapely functionalities in `Vector`,
- Adds many tests, solving several bugs and bringing test coverage to more than 90% (was 75% previously).

This PR is branched from #350 on the `Mask` addition (now closed and everything is in this PR, as some `Mask` fixes started to depend on new behaviours only available here).
This PR also casts single-band raster to 2D following #360 (now closed also).

## Documentation

**The hidden branch with the rendered documentation is here:** https://geoutils-rhugonnet.readthedocs.io/en/docs/index.html

Regarding theme and rendering, this PR:
- Implements the [sphinx-book-theme](https://sphinx-book-theme.readthedocs.io/en/stable/) (same as OGGM and many other packages, it's a sub-theme of the PyData theme, used by NumPy, Pandas, SciPy, etc). Good news also: the dark theme PR has been merged in their repo, we'll have it available at their next version release (EDIT: It's released! I just need to update the logo files to support the Dark theme :D; SECOND EDIT: Done!),
- Switches to [MyST-parser](https://myst-parser.readthedocs.io/en/latest/) which is an extension of Markdown: partially to avoid re-learning another language with rST, but mostly because it is getting popular thanks to its many functionalities. See next point.
- Switches to [MyST-NB](https://myst-nb.readthedocs.io/en/latest/) that allows to add Jupyter code-cells to Markdown to render code in the documentation, and easily switch to interactive mode with a Binder. It allows to write Python code in Markdown, or to write Markdown in Jupyter notebook. Both will be rendered the same! Older packages (and even more recent ones like Xarray) had to rely on the IPython sphinx extension, which was really a hassle (multi-line is hard, few rendering options, etc, for example here:  https://docs.xarray.dev/en/stable/user-guide/indexing.html#positional-indexing).
- Removes the `/doc/code` folder and the `literalinclude:: code/coregistration.py :lines: 5-21` we were using in xDEM, which was a hassle to maintain, thanks to the previous point. Proper errors should be raised by Sphinx or can be tested in the CI directly (we can run the `.md` files as a Jupyter notebook).
[...]

PR #436, opened 2024-01-17
Attempt to fix issue with doc building on Mac/Windows

adehecq:
Attempts to fix an issue only referenced on xdem: https://github.com/GlacioHack/xdem/issues/316.

- Fix an issue of shapely backward compatibility in one of the examples 
- Re-enable `test_doc.py` for Mac -> The test now passes ! 🥳 (I'm not sure if it's related to my change or it was fixed earlier). 
- Remove the fixed version for pydata-sphinx-theme which was causing a bug with the logo (see #351)
- Doc was also building on Windows, until the new CLI doc section (#418)

Raised issue #442 for the Windows problem.

### DIFF
--- a/dev-environment.yml
+++ b/dev-environment.yml
@@ -29,7 +29,7 @@ dependencies:
 
   # Doc dependencies
   - sphinx
-  - pydata-sphinx-theme==0.13.3
+  - pydata-sphinx-theme
   - sphinx-book-theme>=1.0
   - sphinx-gallery
   - sphinx-design
--- a/examples/analysis/point_extraction/reduction.py
+++ b/examples/analysis/point_extraction/reduction.py
@@ -52,7 +52,7 @@ np.nanmean(vals - vals_reduced)
 coords = rast.coords(grid=True)
 x_closest = rast.copy(new_array=coords[0]).value_at_coords(x=x_coords, y=y_coords).squeeze()
 y_closest = rast.copy(new_array=coords[1]).value_at_coords(x=x_coords, y=y_coords).squeeze()
-from shapely import box
+from shapely.geometry import box
 
 geometry = [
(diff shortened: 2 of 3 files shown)


adehecq:
Should I disable tests for Windows then? Or we should have someone test it locally on a Windows to understand the issue...

rhugonnet:
Talk about a useless message, `Undefined error`... :sweat: 

It also says "Theme error" right next to it.
We could try unfixing the `pydata-sphinx-theme` version that had to be fixed to build correctly with `sphinx-book-theme`: https://github.com/GlacioHack/geoutils/blob/main/dev-environment.yml#L32. See issue: https://github.com/GlacioHack/geoutils/pull/413.

We can unfix it now because Sphinx book theme just released a new version that works with the new Pydata theme: https://github.com/executablebooks/sphinx-book-theme/releases, and apparently fixes a lot of bugs.

If the rendering is good, it would also solve issue https://github.com/GlacioHack/xdem/issues/452 once mirrored in xDEM! :slightly_smiling_face:

rhugonnet:
And otherwise yes, we should probably not lose time on this mysterious error and just skip Windows

adehecq:
@rhugonnet, the test does not fail on Windows anymore after unfixing the `sphinx-book-theme` !! 🥳 
You can check the new doc here: https://geoutils-adehecq.readthedocs.io
I didn't see any issue. The theme is slightly different than the previous one, but it looks fine.

rhugonnet (approved):
Amazing!! We're solving a lot of old problems :)

adehecq:
Pfff, after merging with the main branch, the doc build fails again on Windows. It seems related to the new CLI section. The path to the script `geoviewer.py` is not found on windows. This may be a sign that actually, `geoviewer.py` is not working on Windows?

Issue #432, opened 2024-01-16
Homogenize arguments: `inplace` default of ops (`reproject`, `crop`) and georeferenced args (`ref`, `dst_ref`, etc...)


rhugonnet:
Two problems linked to the default arguments of our functions:

1. It's a bit un-intuitive that some operations would be inplace (`crop`) and others not (`reproject`).

We should probably do the same done by NumPy, Xarray and Pandas: never do inplace, and leave to the user to do it themselves by doing:
```python
raster = raster.crop(mycropgeom)
```
If they want to.

2. Arguments that do the same thing have different names in different functions.

For instance the "match-reference" argument: `dst_ref`, `ref`, etc... Same for many other arguments of functions of `Raster` and `Vector` (`dst_shape`, `size`, `rst_shape`, etc).

The sooner we homogenize the API, the better, otherwise it will be a hard change for users or would have to stay inconsistent for a long time! And doing it before mirroring the API in the Xarray accessor #383 would be great! :smiley: 
So, I think is a top priority we should take care off soon! Was already listed but also a bit lost in the long list here: https://github.com/GlacioHack/geoutils/issues/361.

**Maybe we should start a Google Docs to converge towards the arguments names that are the most intuitive to use everywhere**? 
What do you think @adehecq @atedstone @erikmannerfelt?

adehecq:
Overall I agree with both suggestions.
We can put this in a Google Doc, but ultimately we should maybe summarize this in a Wiki page, so we remember what was decided?

atedstone:
I also agree with both suggestions. I would prefer a wiki page over a Google Doc, but we can also simply add these information to CONTRIBUTING.md so that everything important about development is in one place?

rhugonnet:
Closed by #440, #445 and #449

PR #440, opened 2024-01-18
Homogenize API arguments

atedstone:
**This PR causes breaking changes to API**

Resolves the rest of #432. 

- [x] Remove `dst_`
- [x] Replace `rst` with raster
- [x] `grid_size` to `size`
- [x] The various forms of sample (see below)
- [x]  Always use `points`, never `pts`

## Grammar of sampling

This PR proposes to regularise the grammar around the word sample and its derivatives.
- When the intention is to return a (random) sample, the term `sample` is used. Previously, `subset` or `subsample` were both used to denote this purpose.
- `downsample` or `upsample` defines a change in a Raster's resolution.  The more generic term is `resample`, but the use of a specific direction is preferred. Try not to use `resample` as this has connotations of towards `reproject`.
- The terms `subset` and `subsample` refer to an contiguous portion of data (e.g. an `x*y` window of a `Raster`, specific bands of a `Raster`, or `row n...row m` of a `Vector`), without a change in resolution. I am regularising to `subset`.

### DIFF
--- a/doc/source/raster_class.md
+++ b/doc/source/raster_class.md
@@ -222,8 +222,8 @@ print(rast.bounds)
 ```{code-cell} ipython3
 # Reproject to smaller bounds and higher resolution
 rast_reproj = rast.reproject(
-    dst_res=0.1,
-    dst_bounds={"left": 0, "bottom": 0, "right": 0.75, "top": 0.75},
+    res=0.1,
+    bounds={"left": 0, "bottom": 0, "right": 0.75, "top": 0.75},
     resampling="cubic")
 rast_reproj
 ```
--- a/examples/analysis/point_extraction/interpolation.py
+++ b/examples/analysis/point_extraction/interpolation.py
@@ -28,7 +28,7 @@ np.random.seed(42)
 x_coords = np.random.uniform(rast.bounds.left + 50, rast.bounds.right - 50, 50)
 y_coords = np.random.uniform(rast.bounds.bottom + 50, rast.bounds.top - 50, 50)
 
-vals = rast.interp_points(pts=list(zip(x_coords, y_coords)))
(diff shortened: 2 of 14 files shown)


rhugonnet (approved):
Thanks @atedstone, this is great!! I didn't find anything to comment on in the code changes.

I like the homogenization to `sample` and differentiation with `subset`! :smile:  
It looks like `subset` is used only in `split_bands`, we could even rename into something specific to that function like `bands` to avoid any kind of confusion?

adehecq (approved):
Nice work !
As discussed on slack, I find that removing the `dst_` prefix everywhere makes sense from a user/API point of view, but it may be a bit less easy to read the code. We'll see!
I find that `size` is not clear enough. I would favor `grid_size` to clarify that we're talking about the dimensions of the array. But we can revert in a future PR if we all agree.

adehecq:
> It looks like `subset` is used only in `split_bands`, we could even rename into something specific to that function like `bands` to avoid any kind of confusion?

For consistency, we could use the same terminology as in `load` which is `indexes`. Although I'm not a huge fan of that term, I always forget the name...

atedstone:
I also don't find `indexes` that intuitive. From the xarray side of life, index can be along any dimension. How about the keyword `band/bands` instead, for both `load()` and `split_bands()`?

adehecq:
> I also don't find `indexes` that intuitive. From the xarray side of life, index can be along any dimension. How about the keyword `band/bands` instead, for both `load()` and `split_bands()`?

I would be fine with `bands`. But this would just be undoing what was done in a previous PR: https://github.com/GlacioHack/geoutils/pull/346. Any thoughts @rhugonnet ?

rhugonnet:
+1 for `grid_size` and for `bands`!

Issue #357, opened 2023-03-03
Floating precision of `rio.transform` in `Raster.reproject()` when passing both `dst_bounds` and `dst_res` (or `dst_size`)


rhugonnet:
See #356
This specific case is not fixed after #354.

adehecq:
To reproduce the error, with the dataset provided in #354 at [this link](https://drive.google.com/file/d/10usaitjrvmhAgywB_3wPOGDGWJZ8iVnT/view?usp=share_link):
```
import geoutils as gu
rst = gu.Raster("N62E010_FABDEM_V1-0.tif")
tmp = rst.reproject(dst_crs="EPSG:32633")
test = tmp.reproject(dst_res=(20, 20), dst_bounds=tmp.bounds)
print(test.res)
```
-> output resolution is (19.99999999999999, 20.0) instead of (20, 20).

Issue #212, opened 2021-06-22
Simplify Raster.reproject


adehecq:
We could use the functionality in rasterio.warp.calculate_default_transform to make the calculation of the transform easier in Raster.reproject (I should have seen that earlier...):
https://rasterio.readthedocs.io/en/latest/api/rasterio.warp.html#rasterio.warp.calculate_default_transform

SOme additional details from the rasterio [discussion group](https://rasterio.groups.io/g/main/message/786): "The intent is that rasterio.warp.reproject, given the values output from calculate_default_transform, will produce a result close to that of GDAL's gdalwarp utility *without* its "-tap" option. And that if you wrap the output of calculate_default_transform with aligned_target, adding a resolution, you will get a result close to that of gdalwarp *with* its "-tap" option."

adehecq:
This is actually a bit more complex than anticipated, as the functions `calculate_default_transform` and `aligned_target` only take into account the input transform, width and height, but not the output bounds.
Obtaining a result similar to what is done in reproject for now (and similar to gdalwarp), but using rasterio's functions seem non trivial for now.

adehecq:
GDAL's logic for default warping options is coded around here https://github.com/OSGeo/gdal/blob/d01976cd4fc11144662f44ec4e658ce1871f1713/gdal/apps/gdalwarp_lib.cpp#L3105

rhugonnet:
I might add here that #354 is still an issue with our current method when passing `dst_bounds` and `dst_res` (or `dst_size`) individually. See #356

PR #443, opened 2024-01-19
Updates in Raster.reproject: fix bug, simplify code and improve tests

adehecq:
Related to #357 and #212.
Re-organized the way `Raster.reproject` works :
- the part that handles the logic of finding the correct output goereferences has been moved to a separate (hidden) function `_get_reproject_params`. It relies now on rasterio functions (`rio.warp.calculate_default_transform` and `rio.transform.from_bounds`) only instead of making our own maths. 
- `reproject` now takes care only of the actual reprojection and handling the data
- re-organized the logic in reproject a bit and improved comments
- improve a test in TestRaster::test_reproject regarding the floating point issue
- updated a test in TestMask::test_reproject that was not actually requiring reprojection to run
- overall, improved logic in test_reproject (testing of match reference should come last instead of first), added some missing tests (coverage increased by 0.2% 🥳) and made minor edits.
- raised an issue in rasterio https://github.com/rasterio/rasterio/issues/3010

### DIFF
--- a/tests/test_raster.py
+++ b/tests/test_raster.py
@@ -2779,15 +2779,16 @@ class TestMask:
     def test_reproject(self, mask: gu.Mask) -> None:
         # Test 1: with a classic resampling (bilinear)
 
-        # Reproject mask
-        mask_reproj = mask.reproject()
+        # Reproject mask - resample to 100 x 100 grid
+        # dst_res = tuple(np.array(mask.res)*2)
+        mask_reproj = mask.reproject(size=(100, 100), src_nodata=2)
 
         # Check instance is respected
         assert isinstance(mask_reproj, gu.Mask)
 
         # This should be equivalent to converting the array to uint8, reprojecting, converting back
         mask_uint8 = mask.astype("uint8")
-        mask_uint8_reproj = mask_uint8.reproject()
+        mask_uint8_reproj = mask_uint8.reproject(size=(100, 100), src_nodata=2)
         mask_uint8_reproj.data = mask_uint8_reproj.data.astype("bool")
 
         assert mask_reproj.raster_equal(mask_uint8_reproj)
--- a/tests/test_raster.py
+++ b/tests/test_raster.py
@@ -1182,17 +1182,149 @@ class TestRaster:
                 f"consequences. Consider setting a different nodata with self.set_nodata()."
             ),
         ):
-            _ = r_nodata.reproject(res=r_nodata.res[0] / 2, src_nodata=default_nodata)
+            r_test = r_nodata.reproject(res=r_nodata.res[0] / 2, src_nodata=default_nodata)
+        assert r_test.nodata == default_nodata
 
         # 3 - if default nodata does not conflict, should not raise a warning
         r_nodata.data[r_nodata.data == default_nodata] = 3
-        _ = r_nodata.reproject(res=r_nodata.res[0] / 2, src_nodata=default_nodata)
+        r_test = r_nodata.reproject(res=r_nodata.res[0] / 2, src_nodata=default_nodata)
+        assert r_test.nodata == default_nodata
 
-        # -- Additional tests -- #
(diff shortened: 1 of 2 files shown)


adehecq:
There is an error in tests/test_raster.py::TestMask::test_reproject[mask1].
This occurs because in that case, nothing is done during reprojection: `mask_reproj = mask.reproject()` ([here](https://github.com/GlacioHack/geoutils/blob/5dcf6efdceab23d09bb95f65f61a15553cf13ec1/tests/test_raster.py#L2789)).

Three things:
- the error is caused by a bug in rasterio, see below
- to circumvent this bug, one can add a test that if dst res, size and bounds are all none, the transform and size do not change
- the test does not make sense here. It should test that reproject works as expected, but in that specific case, reproject return a copy of self, so an actual reprojection should be run

Error in rasterio:
```
import rasterio as rio
crs = rio.CRS.from_epsg(32718)
width, height, left, right, top, bottom = (539, 618, 627175.0, 643345.0, 4852085.0, 4833545.0)
rio.warp.calculate_default_transform(crs, crs, width, height, left=left, right=right, top=top, bottom=bottom)
rio.transform.from_bounds(left, bottom, right, top, width, height)
```
Because in `calculate_default_transform`, none of the crs, res or size are changed, the outcome should be the same as in `from_bounds` but it differs by a very small number...
-> need to raise an issue on rasterio and fix the bug in reproject next week!

tests/test_raster.py
```diff
@@ -1381,24 +1381,18 @@ def test_reproject(self, example: str) -> None:
...
+        # (we take an example that used to fail, see issue #354 and #357)
+        data = np.ones((4759, 2453))
+        transform = rio.transform.Affine(
+            24.12423878332849, 0.0, 238286.29553975424, 0.0, -24.12423878332849, 6995453.456051373
+        )
         crs = rio.CRS.from_epsg(32633)
         nodata = -9999.0
         rst = gu.Raster.from_array(data=data, transform=transform, crs=crs, nodata=nodata)
 
-        # This large grid geotransform is taken from https://cdn.proj.org/us_nga_egm08_25.tif
```
rhugonnet, line 1391: Why not leave this previous example as well?

rhugonnet (approved):
The functionality is so much cleaner (and robust), this is great!! :grin:  

I looked at the Coveralls output (https://coveralls.io/jobs/134992379/source_files/20142500455#L331), it looks like there are 2 cases not covered in the tests:
- Bounds + Size,
- Bounds + Res as a single value.

The rest looks like checks that are duplicated between `reproject` and `_get_reproject_params`. Should we remove them from one or the other?

tests/test_raster.py
adehecq: In your example, you created two rasters (`rst` and `rst2`), one to be reprojected on the same grid as the other, to test the option with match-reference. I find these examples quite specific because the transforms are provided. I found it easier to create only one and then reproject to given bounds and res. If this one case works, by definition the match-reference case works because now this is exactly what is used for the match-reference functionality. Is that ok?

adehecq:
> I looked at the Coveralls output (https://coveralls.io/jobs/134992379/source_files/20142500455#L331), it looks like there are 2 cases not covered in the tests:
> 
>     * Bounds + Size,
> 
>     * Bounds + Res as a single value.
> 

Coverall is really great at showing where tests are still weak ! And I wondered when writing this PR how to see the results of coveralls. 
I improved the tests accordingly. Still waiting for the tests to run on GitHub, but hopefully, tests coverage should increase now 😄 (the most annoying thing is to always have to wait ~10 min to make sure everything runs fine...)

> 
> The rest looks like checks that are duplicated between `reproject` and `_get_reproject_params`. Should we remove them from one or the other?

Good point... I deleted the one in reproject, although it means it will be caught a bit later in the code.

tests/test_raster.py
rhugonnet: OK makes sense!

PR #446, opened 2024-01-24
Add Xarray accessor mirroring `Raster` API

rhugonnet:
Ongoing

 - [ ] Resolves #xxx,
 - [ ] Tests added, otherwise issue #xxx opened,
 - [ ] New optional dependencies added to both `dev-environment.yml` and `setup.cfg`,
 - [ ] Fully documented, including `api/*.md` for new API.


rhugonnet:
@adehecq @atedstone @erikmannerfelt This PR is also ready for your first review! It is not finalized, but at a good stage to hear your feedback, questions, recommendations, and then move forward to finalize it.
I have described the general concepts in the description, and the tests comparing both types of outputs should give you a good idea of how this works in practice in the code.

Once you are done with this one, you should look at the one for the `dem` accessor in xDEM (which depends on this PR).
Much less changes are needed, but it has a slightly more complex class structure: https://github.com/GlacioHack/xdem/pull/656

environment.yml
```diff
@@ -12,5 +12,6 @@ dependencies:
   - scipy=1.*
   - tqdm
   - xarray
-  - dask
   - rioxarray=0.*
+  - geocube
+  - dask
```
adehecq, line 17: should we make it optional? Does it add a lot of inherited dependencies?

geoutils/raster/xr_accessor.py
```diff
@@ -0,0 +1,105 @@
...
+
+        data = data.squeeze()
+
+        # Get netCDF coordinates from transform and shape
+        coords = affine_to_coords(affine=transform, width=data.shape[0], height=data.shape[1])
+
+        # Build a data array
+        out_ds = xr.DataArray(
+            data=data,
+            coords={"x": coords["x"], "y": coords["y"]},  # Somehow need to re-order the coords...
```
adehecq, line 87: I don't understand the comment. Where is the re-ordering taking place?

geoutils/raster/base.py
adehecq, line 1: Is there any change here apart from changing the file name and class name? Here all lines are displayed as new so it is difficult to see any change...

geoutils/raster/raster.py
adehecq, line 1: Also here I assume that there is little change apart from the parts that have been moved to the RasterBase class?

adehecq:
This is a giant step forward!! 🎉 
Thanks for the effort you put into this over the past months and not only in this PR, which is just the final push! 🙌 🙌 🙌 

I don't have many comments, as I'm not so familiar with the xarray accessor, so I just have to trust you that it works. I'm looking forward to see it explained in the documentation and test it myself, but at the moment it is still abstract :wink:
My main comment is regarding the new Dask dependency. Is it really needed or could it be optional? Does it add a lot of inherited dependencies?

I can also only vaguely imagine the limitations you mention in the description about not accessing the hidden functions. I guess we'll figure out the issues by using and practicing. But the first step is to get it out!

geoutils/raster/xr_accessor.py
```diff
@@ -0,0 +1,105 @@
...
+        out_ds.rio.write_transform(transform)
+        out_ds.rio.set_crs(crs)
+        out_ds.rio.set_nodata(nodata)
+
+        return out_ds
+
+    def to_raster(self) -> RasterBase:
+        """
+        Convert to Raster object.
+        # TODO: Rioxarray uses "to_raster" to write to file... Change naming?
```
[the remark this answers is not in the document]
atedstone: Left-field suggestion and probably doesn't work, but any way to leverage `.astype(Raster)`?

geoutils/raster/xr_accessor.py
```diff
@@ -0,0 +1,105 @@
...
+        # Build a data array
+        out_ds = xr.DataArray(
+            data=data,
+            coords={"x": coords["x"], "y": coords["y"]},  # Somehow need to re-order the coords...
+            attrs=tags,
+        )
+
+        # Set other attributes
+        out_ds.rio.write_transform(transform)
+        out_ds.rio.set_crs(crs)
```
atedstone, line 93: This yields a warning, we could update already to `.write_crs()`.

```
FutureWarning: It is recommended to use 'rio.write_crs()' instead. 'rio.set_crs()' will likely be removed in a future release.
  out_ds.rio.set_crs(crs)
```

atedstone:
This also looks like a huge step forward in combining all the previous efforts and refactoring. I just had a quick go and was really happy with how simple a reproject() call, for example, can be. Similarly to Amaury, a lot of the functionality is a bit abstract for now - I'm especially thinking about the delayed computation/dask work - so the documentation effort will be super important I feel. 

Unclear to me is exactly what from the outstanding To Do list is needed before we can release the basics of the xarray accessor into the main channel? Do we need to complete everything before a release?

dev-environment.yml
```diff
@@ -11,9 +11,10 @@ dependencies:
   - numpy>=1,<3
   - scipy=1.*
   - tqdm
-  - xarray>2023,<=2025.04
-  - dask
+  - xarray
   - rioxarray=0.*
+  - geocube
+  - dask
```
guillaumeeb: dask is twice in this file, I guess this makes sense to comment it as optional.

environment.yml
guillaumeeb: Not sure it adds a lot of dependencies you don't have, but it should probably be optional.

geoutils/raster/base.py
```diff
@@ -0,0 +1,1909 @@
...
+    """
+    This class is non-public and made to be subclassed.
+
+    It gathers all the functions shared by the Raster class and the 'rst' Xarray accessor.
+    """
+    def __init__(self):
+        """Initialize all raster metadata as None, for it to be overridden in sublasses."""
+
+        # Attribute for Xarray accessor: will stay None in Raster class
+        self._obj: None | xr.DataArray = None
```
guillaumeeb, line 79: Why name this DataArray `_obj`?

geoutils/raster/base.py
```diff
@@ -0,0 +1,1909 @@
...
+        if new_tags is None:
+            new_tags = {}
+        self._tags = new_tags
+
+    @property
+    def is_loaded(self) -> bool:
+        """Whether the raster array is loaded."""
+        if self._is_xr:
+            # TODO: Activating this requires to have _disk_shape defined for RasterAccessor
+            return True
```
guillaumeeb: Why true? Xarray without Dask has already some lazy behavior, and much more when using dask.

geoutils/raster/base.py
```diff
@@ -0,0 +1,1909 @@
...
+
+        :returns: A new raster (or None if inplace).
+        """
+
+        crop_img, tfm = _crop(source_raster=self, bbox=bbox, mode=mode)
+
+        if inplace:
+            self._data = crop_img
+            self.transform = tfm
+            return None
```
guillaumeeb: Why don't your return self?

geoutils/raster/base.py
```diff
@@ -0,0 +1,1909 @@
...
+                output_win = list_windows  # type: ignore
+
+        # Return array or pointcloud
+        if not as_array:
+            output_val = PointCloud.from_xyz(x=points[0], y=points[1], z=output_val, crs=self.crs)
+
+        if return_window:
+            return (output_val, output_win)
+        else:
+            return output_val
```
guillaumeeb: Missing newline

geoutils/raster/base.py
```diff
@@ -0,0 +1,1909 @@
...
+
+    def set_crs(self, new_crs: CRS) -> None:
+        """
+        Set the coordinate reference system of the raster.
+        """
+
+        if new_crs is not None:
+            new_crs = CRS.from_user_input(value=new_crs)
+
+        if self._is_xr:
```
guillaumeeb: There are many places in this base class where this is tested. Why don't you define abstract methods to overload instead?

geoutils/raster/xr_accessor.py
```diff
@@ -0,0 +1,105 @@
...
+
+    # Open with Rioxarray
+    ds = rioxr.open_rasterio(filename, masked=False, **kwargs)
+
+    # Change all nodata to NaNs
+    ds = ds.astype(dtype=np.float32)
+    ds.data[ds.data == ds.rio.nodata] = np.nan
+
+    # Remove the band dimension if there is only one
+    if ds.rio.count == 1:
```
guillaumeeb: Couldn't we use the `xarray.squeeze` method directly?

geoutils/raster/xr_accessor.py
```diff
@@ -0,0 +1,105 @@
...
+        ds = ds.isel(band=0, drop=True)  # Delete band coordinate (only one dimension)
+
+    # Store disk attributes? Didn't follow the logic in Rioxarray on this, need to get more into code details
+    ds.rst._count_on_disk = 1
```
guillaumeeb: I'm not sure what is a disk attribute for you?

guillaumeeb (commented):
I'm really new to geo-utils, so please take my comment lightly :). 

I guess this is alright for a first implementation, but I have two points of attention for the future:

- Will this accessor be accessible on Xarray Datasets on top of Dataarrays, if the Dataset is composed of these DataArrays?
- The implementation relies on rioxarray, which is understandable, but it might be a good thing to replace it in the future.

dev-environment.yml
rhugonnet: Yes I agree, optional is definitely the way, it's even the case for Xarray: https://docs.xarray.dev/en/v2025.04.0/getting-started-guide/installing.html
We actually recently mirrored their optional dependency test structure for a more robust CI, so this is ready to go! (#738)

geoutils/raster/base.py
rhugonnet: As far as I'm aware, it's the traditional naming for writing an accessor class (`_obj` is not accessed by users, it's only within the accessor to be able to access `.data`, that otherwise doesn't exist with this type of accessor structure, as there's no inheritance/composition): https://docs.xarray.dev/en/stable/internals/extending-xarray.html#writing-custom-accessors

guillaumeeb: Ok sorry, I should have read the Xarray accessor documentation :). 

So I guess my main point here would be to be sure wether `_obj` and `_is_rx` are needed in this base class or not: could/should we move `_obj` in the accessor class? Do we really need the `is_xr` method?

rhugonnet: Good point, I'll see if we really need to have a `None` in the Base class :thinking:. It might purely be a typing issue.
For `is_xr`: I think yes probably, to know when to return a certain output type, because the Raster class works on masked arrays not supported by Xarray, especially for an in-place operation. Maybe that'll get better once every method (e.g. `_reproject()`) returns the same NumPy array type as input (masked or unmasked), outside of the classes.

dev-environment.yml
```diff
@@ -11,9 +11,10 @@ dependencies:
   - numpy>=1,<3
   - scipy=1.*
   - tqdm
-  - xarray>2023,<=2025.04
-  - dask
+  - xarray
   - rioxarray=0.*
+  - geocube
```
remi-braun: For me, classic dependencies in geospatial world are rioxarray and geopandas. Geocube is less well known. Should it really be a dependency of geoutils? If you ever discard rioxarray, won't it make it harder for you to depend on geocube?
(Moreover, I haven't seen any use of geocube in this PR, but i may have missed it)

geoutils/raster/base.py
```diff
@@ -0,0 +1,1909 @@
...
+
+    @property
+    def transform(self) -> Affine:
+        """
+        Geotransform of the raster.
+
+        :returns: Affine matrix geotransform.
+        """
+        if self._is_xr:
+            return self._obj.rio.transform(recalc=True)
```
remi-braun: I understood that the `recalc` argument in rioxarray is here to guarantee laziness (I may have misunderstood it though 😅 ). Are you sure setting it to true won't break this behaviour?

geoutils/raster/base.py
```diff
@@ -0,0 +1,1909 @@
...
+    def nodata(self) -> int | float | None:
+        """
+        Nodata value of the raster.
+
+        When setting with self.nodata = new_nodata, uses the default arguments of self.set_nodata().
+
+        :returns: Nodata value.
+        """
+        if self._is_xr:
+            return self._obj.rio.nodata
```
remi-braun: Do you manage [all the complexity of rioxarray nodata](https://corteva.github.io/rioxarray/stable/getting_started/nodata_management.html) ?
- If the array is not masked, then the `nodata` value is equal to the GTiff nodata and the `encoded_nodata` value is equal to None. In this case the raster is kept in original dtype.
- If the array is masked, then the `nodata` value is equal to nan and the `encoded_nodata` value is equal to the real nodata value given by the GTiff. In this case the raster is converted into float.

geoutils/raster/base.py
remi-braun: I agree, there should be a look to the chunk attribute of the raster, if None it's loaded (and this is True), but if the chunks are supplied, this should be False.

geoutils/raster/base.py
```diff
@@ -0,0 +1,1909 @@
...
+    def count_on_disk(self) -> None | int:
+        """Count of bands on disk if it exists."""
+        if self._disk_shape is not None:
+            return self._disk_shape[0]
+        return None
+
+    @property
+    def count(self) -> int:
+        """Count of bands loaded in memory if they are, otherwise the one on disk."""
+        if self.is_loaded:
```
remi-braun: NB: This also exist in rioxarray via [`rio.count`](https://corteva.github.io/rioxarray/stable/rioxarray.html#rioxarray.rioxarray.XRasterBase.count). Same for height, width, shape.

geoutils/raster/base.py
```diff
@@ -0,0 +1,1909 @@
...
+        data = self.data[band - 1, :, :] if self.count > 1 else self.data
+
+        # Derive inlier mask
+        if inlier_mask is not None:
+            valid_points = np.count_nonzero(np.logical_and(np.isfinite(data), ~data.mask))
+            if isinstance(inlier_mask, RasterMask):
+                inlier_points = np.count_nonzero(inlier_mask.data)
+            else:
+                inlier_points = np.count_nonzero(inlier_mask)
+            dem_masked = self.copy()
```
remi-braun: I think this dates from xdem, but no DEM are used here, rather a raster 😉

geoutils/raster/geotransformations.py
```diff
@@ -121,15 +122,24 @@ def _reproject(
...
-        # Set mask
-        dst_arr = np.ma.masked_array(dst_arr.astype(dtype), fill_value=nodata)
-        if nodata is not None:
-            dst_arr.mask = dst_arr == nodata
-        return False, dst_arr, reproj_kwargs["dst_transform"], reproj_kwargs["dst_crs"], reproj_kwargs["dst_nodata"]
 
+        # If input is Xarray, use NaNs for nodata
+        if source_raster._is_xr:
+            src_arr = source_raster.data
+            src_arr[np.isnan(src_arr)] = src_nodata
```
remi-braun: Are you sure this always works with dask arrays? Dask has some [slicing limitations](https://docs.dask.org/en/stable/array-slicing.html#slicing) wrt numpy.

geoutils/raster/geotransformations.py
```diff
@@ -121,15 +122,24 @@ def _reproject(
...
+            dst_arr, transformed = _rio_reproject(src_arr=src_arr, reproj_kwargs=reproj_kwargs)
+            dst_arr[dst_arr == nodata] = np.nan
```
remi-braun: Same here

geoutils/raster/xr_accessor.py
```diff
@@ -0,0 +1,105 @@
...
+
+import geoutils as gu
+from geoutils.raster.base import RasterBase
+
+from geoutils._typing import NDArrayNum, NDArrayBool, MArrayNum
+
+def open_raster(filename: str, **kwargs):
+
+    # Open with Rioxarray
+    ds = rioxr.open_rasterio(filename, masked=False, **kwargs)
```
remi-braun: Why masked = False? To be sure nodata stores the nodata values and not nan?

geoutils/raster/xr_accessor.py
```diff
@@ -0,0 +1,105 @@
...
+
+    # Change all nodata to NaNs
+    ds = ds.astype(dtype=np.float32)
+    ds.data[ds.data == ds.rio.nodata] = np.nan
```
remi-braun: Again here, I'm not sure this is bullet proof with dask

geoutils/raster/xr_accessor.py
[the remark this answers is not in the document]
remi-braun: As rioxarray will be a base dependency, I would agree with the comment : the name should be changed 👍

rhugonnet:
FYI everyone: I'll be on this during the winter holidays (only taking a few days off). Have a nice break!

environment.yml
rhugonnet: Solved a couple weeks ago by #810

geoutils/raster/base.py
rhugonnet: Almost none except the `if self.is_xr` bits.

geoutils/raster/raster.py
rhugonnet: Almost none also, yes.

dev-environment.yml
rhugonnet: Now optional since #810

rhugonnet:
@guillaumeeb @remi-braun @belletva @adehecq I'm refining the behaviour of the accessor, and I'd like your take to decide on nodata management. :thinking: 

**Some quick background:**
In the `Raster` class, every data type (e.g. `uint8` or `int32` or `float64`) is supported with nodata-resilient operations for users (in the package, e.g. `rast.reproject()`, but also with other packages that understand the array interface e.g. `np.mean(rast)`) natively because we use masked arrays.
For the Xarray accessor, we can't have masked arrays in the DataArray, only a classic NumPy array... So operations can only be nodata-resilient using NaNs for `np.floating` types. For `uint8` types, no way of defining a NaN, it can only be the original nodata value, which we know (it is stored in `.nodata`).
In Rioxarray, the data can be opened as **"raw arrays" i.e. arrays containing the nodata value directly** (e.g., -9999 everywhere) or masking it (which means replacing by NaN in this context, only works for `floating` types).

Thus, we could manage to support **raw arrays** in our methods (i.e. we'd use the `nodata` value to create a mask and update nodatas before/after an operation called by the accessor), but it would still limit users in other code (using NumPy, SciPy or simple Python arithmetic would fail because it would wrongly propagate the nodata, such as -9999, through operations).

So there's two choices:

1. What should we support?

Right now, our underlying functions (`_reproject()`, `_proximity`, `_filter`, etc...) only support masked array inputs (and can easily support NaN arrays).
Should we also allow users to input a "raw array" (with nodata values still in) for this type of use through the Xarray accessor?
Technically, it's fairly easy, we just add a wrapper everywhere before/after each operations (essentially doing `mask = input_array == nodata_value` before, then `output_array[output_mask] = nodata` after. But we'd also need a new attribute to know if we need to do this at all (defined on file opening, or at user request), which is a bit of management.

**Or, if we don't add this support for raw arrays, we would have to force NaN arrays on opening (through a `gu.open_raster`), converting any `integer` data to `float32`.**
In my opinion, it isn't such a big limitation (although it increases memory usage, this can be reduced using Dask), but it's a big design choice!

2/ What default behaviour?

If we support both NaN and raw arrays (as Rioxarray does), what should we choose as default? 
Here again, I would be in favour of **forcing the data to be a NaN array as default for users.** (so that it integrates naturally with the array interface of Xarra when calling NumPy/SciPy).

rhugonnet:
My choice would be to only support NaN arrays for now (which would then be the default, forcing all arrays to be floating on opening). 
And we could always add support for "raw arrays" in the future, if we feel it could be useful.

geoutils/raster/base.py
rhugonnet: Indeed, had to keep "None" for typing reasons, and `_is_xr` (non-public) for casting.

geoutils/raster/base.py
rhugonnet: In-place operations traditionally return None (some used to return self, but then it gets inadvertently printed in the console).
I checked against other package behaviour (Pandas/Xarray), and we're consistent :slightly_smiling_face: :

```
import pandas as pd
df = pd.DataFrame(data={"a": [1]})
test = df.reset_index(inplace=True)
print(test)
```
```
None
```

geoutils/raster/base.py
rhugonnet: You're right, I shouldn't branch directly in the base class methods!
I've replaced by abstract methods overloaded in the subclasses :slightly_smiling_face:

dev-environment.yml
rhugonnet: You're right, it was a leftover of the initial draft a while ago. Geocube removed.

geoutils/raster/base.py
rhugonnet: I haven't gotten there in the tests yet, but removed it! Leaving this open to think about adding tests on laziness

geoutils/raster/base.py
rhugonnet: As we just discussed below a few days ago, we are only keeping the `encoded_nodata`!

geoutils/raster/xr_accessor.py
rhugonnet: Done!

geoutils/raster/xr_accessor.py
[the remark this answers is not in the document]
rhugonnet: I have changed into `to_geoutils`, which makes sense given that the other function (`Raster` to `xr.DataArray` is called `to_xarray`).

geoutils/raster/xr_accessor.py
rhugonnet: Clarified.

geoutils/raster/xr_accessor.py
rhugonnet: Removed.

geoutils/raster/base.py
rhugonnet: In more details: Uses `encoded_nodata`, except if user has re-set another value since.

geoutils/raster/xr_accessor.py
rhugonnet: Updated to `masked=True` and using `encoded_data`.

geoutils/raster/xr_accessor.py
rhugonnet: Removed.

geoutils/raster/geotransformations.py
rhugonnet: Removed

geoutils/raster/geotransformations.py
rhugonnet: Also removed (this now happens directly in the `_rio_reproject` function that runs for each chunk, so works with Dask).

geoutils/raster/base.py
rhugonnet: Indeed. Actually this is a recent addition from either @marinebcht or @vschaffn, we need to be more careful! :stuck_out_tongue_winking_eye: 
Fixed it.

atedstone:
@rhugonnet this looks immense, I am looking forward to trying out this big new functionality properly. Given the decent test suite coverage and my lack of deep technical expertise on accessors, I would be content to see this merged into the main dev branch and then to get a bit more involved once we have the skeleton of documentation available as you suggest.

rhugonnet:
I've now finalized the tests that check that Dask out-of-memory reprojection gives 100% the same result as in-memory (only possible since we added the `tolerance` argument in Rasterio, which officially came out in 1.5, and set it at "0" by default).
Merging!

adehecq:
Well done @rhugonnet !! 🙌

Issue #447, opened 2024-01-24
Sampling interval is irregular when loading Raster with downsampling option


adehecq:
**Describe the bug**
When loading a raster with `geoutils.Raster(filename, downsample=6)`, one can expect that the pixels are sampled every 6 pixels, without interpolation. This is not the case.

**To Reproduce**

The best way to test this, is to create a Raster with data that increases linearly along lines and/or columns. Here I create a raster that has a gradient of 1 along rows and 100 along lines. I save this raster and open it with downsampling option. The gradient of the output should be `1*downsample` along rows and `100*downsample` along columns, but when the shape is not a multiple of downsample, it does not work.
For example:
```
import geoutils as gu
import numpy as np
import matplotlib.pyplot as plt
import rasterio as rio
from tempfile import NamedTemporaryFile

# Create a fake Raster with 
# row1: 0, 1, 2, 3... 99
# row2: 100, 101, ... 199 etc
data = np.arange(5000).reshape((50, 100))
transform = rio.transform.from_bounds(0, 0, 100, 50, 100, 50)
rst = gu.Raster.from_array(data, crs="EPSG:4326", transform=transform)

# Save to file (needed to open with downsample option
tmpfile = NamedTemporaryFile(suffix=".tif")
rst.save(tmpfile.name)

# Load with a downsampling that is not multiple of 50 or 100
test = gu.Raster(tmpfile.name, downsample=6)

# Plot gradients
fig, axes = plt.subplots(1, 2, figsize=(16, 4))
im1 = axes[0].imshow(np.gradient(test.data.data, axis=0))
plt.colorbar(im1)

im2 = axes[1].imshow(np.gradient(test.data.data, axis=1))
plt.colorbar(im2)
plt.tight_layout()
plt.show()
```

The output figure is
![Figure_1](https://github.com/GlacioHack/geoutils/assets/3285905/21775bd6-2e16-4dd6-8719-c1f6a49ed54d)

**Expected behavior**

The gradient of the downsampled image should be constant.

**System (please complete the following information):**
 Irrelevant

adehecq:
The issue is caused by the fact that when downsampling, we are rounding up the shape ([here](https://github.com/GlacioHack/geoutils/blob/bb88ed665098cc131bf66bcec692a8701c65744e/geoutils/raster/raster.py#L487)).
[...]

Issue #361, opened 2023-04-13
Features to finalize before `v0.1`


rhugonnet:
Bugs:
- [ ] Nodata are plotted for GeoDataFrame (see regular grid interpolation example),
- [ ] Colorbar is wrong for `Vector.show()`
- [ ] Polygonizing "all" values on the Landsat_B4 example creates some LineString inside polygon?

Improvements:
- [ ] Homogenize array shape input/output of `interp_points` and `value_at_coords`, and variable names. Make output shape consistent for `interp_points` and others: if single-coordinate `float` is passed, return the same. Otherwise, if `ArrayLike` is passed, return `ArrayLike`. Could create a generic function for dealing with points: accept a `Vector` of type point, accept any shape of input list or array, and use it everywhere?
- [ ] Overload `Mask.show` to only a discontinuous, 2-color colorbar,
- [ ] Consistently allow having only `shape` or `res` as argument for Vector rasterizing methods (`rasterize`, `create_mask`, `proximity`).

Features:
- [ ] A `Vector` random point generation in an outline, can pass a `Raster` object as match-reference to get a random point in its extent.
- [ ] Add mask options for `window` of `value_at_coords` (disk, etc),
- [ ]

PR #448, opened 2024-01-24
Solves #408 (incompatibility between crop and downsampling)

### LINKED ISSUE #408: Raster downsampling and cropping operations seem to be incompatible
**Describe the bug**
When opening a Raster with the downsample option, running Raster.crop() immediately afterwards results in a blank Raster.

**To Reproduce**
```python
im = gu.Raster(im_uri, downsample=5)
im2 = im.crop(v, inplace=False)
im2.show()
```

In contrast, if loading `im` is forced with
```python
im = gu.Raster(im_uri, downsample=5, load_data=True)
```
then the crop functions as expected.

**Expected behavior**
The cropping operation should take account of the downsampling during the on-the-fly data loading, returning a cropped raster at the specified downsample factor.

JupyterLab, geoutils=0.0.16.dev2+gaf2fc8c (also geoutils=0.0.11)

**Additional context**

adehecq:
Resolves #408.

Phew, this was a tough one! It turns out that `Raster.crop()` was not very thoroughly implemented in the case when data is not loaded. A few things to pay attention to:
- one needs to create a Window using the on-disk transform, not the in-memory one, otherwise we're just reading the wrong bits!
- calling `rio.DatasetReader.read` is actually more complex than it seems when either of cropping or downsampling need to be executed. One needs to make sure that Window and out_shape really match. The window shape can actually be misleading because with option `boundless=False` (default), the window will be cropped if it extends beyond the raster extent. It makes sense after all...

Doing so, I noticed that our downsampling functionality is actually slightly wrong, see issue #447. Basically, the sampling interval is not exactly constant.
Because of that, one does not get exactly the same result if cropping a downsampled raster with or without data loaded (see commented test). This should be fixed together with #447.

Decided to also modify `Raster.info()` to print to screen by default, as mentioned in #361.

### DIFF
--- a/geoutils/raster/raster.py
+++ b/geoutils/raster/raster.py
@@ -1654,16 +1654,23 @@ np.ndarray or number and correct dtype, the compatible nodata value.
         else:
             self.data[mask_arr > 0] = np.ma.masked
 
-    def info(self, stats: bool = False) -> str:
+    @overload
+    def info(self, stats: bool = False, *, verbose: Literal[True] = ...) -> None:
+        ...
+
+    @overload
+    def info(self, stats: bool = False, *, verbose: Literal[False]) -> str:
+        ...
+
+    def info(self, stats: bool = False, verbose: bool = True) -> None | str:
         """
-        Summarize information about the raster.
+        Print summary information about the raster.
 
         :param stats: Add statistics for each band of the dataset (max, min, median, mean, std. dev.). Default is to
             not calculate statistics.
+        :param verbose: If set to True (default) will directly print to screen and return None
 
-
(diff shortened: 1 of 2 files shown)


rhugonnet (approved):
Amazing! No comment on the code!

My only question: While you are still in this, what do you think is the best way to fix the downsampling issue consistently? #447

adehecq:
> My only question: While you are still in this, what do you think is the best way to fix the downsampling issue consistently? #447

I made some suggestions in the issue directly. I just don't want to have a look at this now, because I'm sure it will take a few hours to get everything sorted and thoroughly tested...

PR #450, opened 2024-01-24
Additional homogenization and fixes of argument names

### LINKED ISSUE #441: Remove `src_nodata` from `Raster.reproject()`
The proposal is to remove the ability to force a source `nodata` value in `Raster.reproject()`. 

Instead, we would instruct users to in the docstring that, to use a different nodata, use `raster.set_nodata` before reproject. 

See discussion in Jan 2024 hack GDoc for more info.

### LINKED ISSUE #445: `Raster.crop()` returns self instead of None
This should still return `None` if it is inplace and `crop_img` was already written to `.data`, no? :thinking:

_Originally posted by @rhugonnet in https://github.com/GlacioHack/geoutils/pull/439#discussion_r1457321485_

### LINKED ISSUE #449: Replace names `size` and `indexes`
+1 for `grid_size` and for `bands`!

_Originally posted by @rhugonnet in https://github.com/GlacioHack/geoutils/issues/440#issuecomment-1904244023_
            
Following #440, it was decided to rename `size` (in reproject and proximity) by `grid_size` and `indexes` (in load and split_bands) by `bands`, therefore completely reverting back what was done in previous PRs. 🤦‍♂️

rhugonnet:
Ongoing

Resolves #441 
Resolves #449 
Resolves #445

Also re-orders some arguments of `reproject` (most used to least used), to make access easier. 

Left to-do:

- [x] Decide if we keep `indexes` and `indexes_on_disk` in the properties, or change those as well?
- [x] Simplify the "checking `nodata`" mess in `reproject`.

### DIFF
--- a/bin/geoviewer.py
+++ b/bin/geoviewer.py
@@ -135,7 +135,7 @@ def main(test_args: Sequence[str] = None) -> None:
         dfact = 1
 
     # Read image
-    img = Raster(args.filename, downsample=dfact, indexes=args.band)
+    img = Raster(args.filename, downsample=dfact, bands=args.band)
 
     # Set no data value
     if args.nodata == "default":
--- a/examples/handling/georeferencing/reproj_raster.py
+++ b/examples/handling/georeferencing/reproj_raster.py
@@ -58,5 +58,5 @@ rast2.show(ax="new", cmap="Blues")
 # Ensure the right nodata value is set
 rast2.set_nodata(0)
 # Pass the desired georeferencing parameters
-rast2_warped = rast2.reproject(size=(100, 100), crs=32645)
+rast2_warped = rast2.reproject(grid_size=(100, 100), crs=32645)
 print(rast2_warped.info())
--- a/examples/io/open_save/read_raster.py
+++ b/examples/io/open_save/read_raster.py
@@ -33,7 +33,7 @@ rast.show(cmap="Greys_r")
 # %%
 # Opening can be performed with several parameters, for instance choosing a single band with ``index`` and re-sampling with ``downsample``, to subset a 3-band
 # raster to its second band, and using 1 pixel out of 4.
-rast = gu.Raster(gu.examples.get_path("everest_landsat_rgb"), indexes=2, downsample=4)
+rast = gu.Raster(gu.examples.get_path("everest_landsat_rgb"), bands=2, downsample=4)
 rast
 
 # %%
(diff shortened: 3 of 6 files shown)


rhugonnet:
@adehecq @atedstone This PR can be finished fairly quickly, we just need to decide:
Do we also rename the public properties `Raster.indexes` and `Raster.indexes_on_disk` to `Raster.bands` and `Raster.bands_on_disk`?

There were a lot more `index` argument to change to `bands` than anticipated in the end (we forgot about `interp_points`, `value_at_coords`, `__init__` and `show`).

rhugonnet:
The problem would be that `Raster._indexes` and `Raster.bands` are the same thing while named differently, which is not super consistent... But it is only an internal thing to deal with really, aimed to make it easier for us (devs) to interface with Rasterio by using `._indexes` and easier for users to understand the variable with `.bands` (there are so many other Python "indexes" than Raster bands, it's quite unintuitive)

adehecq:
> @adehecq @atedstone This PR can be finished fairly quickly, we just need to decide: Do we also rename the public properties `Raster.indexes` and `Raster.indexes_on_disk` to `Raster.bands` and `Raster.bands_on_disk`?
> 
> There were a lot more `index` argument to change to `bands` than anticipated in the end (we forgot about `interp_points`, `value_at_coords`, `__init__` and `show`).

The issue with removing the attribute `indexes` is that we lose consistency with rasterio. So far, we ensured that all main attributes of a rasterio dataset exist for Raster... 
A (not so pleasant) option would be to have attribute `bands` and then `indexes` being an alias to `bands`? But it should be our only exception to not using bands. For example `indexes_on_disk` has not equivalent in rasterio, so I would rename it `bands_on_disk`.

adehecq:
> The problem would be that `Raster._indexes` and `Raster.bands` are the same thing while named differently, which is not super consistent... But it is only an internal thing to deal with really, aimed to make it easier for us (devs) to interface with Rasterio by using `._indexes` and easier for users to understand the variable with `.bands` (there are so many other Python "indexes" than Raster bands, it's quite unintuitive)

And hence, I would also rename `_indexes` with `_bands`. We should make sure that we *never* use the term `indexes` apart when calling rasterio functions, and the one attribute `Raster.indexes` (which should in no way be editable).

rhugonnet:
I like the solution of having both `indexes` and `bands` for inter-operability, I'll do that!

rhugonnet:
Ready to merge!

adehecq (approved):
Looking good to me ! 
Some more changes to be updated in xdem now 😄

PR #452, opened 2024-01-25
Fix #451 and improve tests for downsampling

### LINKED ISSUE #451: Raster.bounds is incorrect with downsampling
**Describe the bug**
The bounds output by `Raster.bounds` is incorrect for a downsampled raster.

**To Reproduce**
```
import geoutils as gu
fn = gu.examples.get_path("everest_landsat_b4")
rst_orig = gu.Raster(fn)
rst_dw = gu.Raster(fn, downsample=5)
print(rst_orig.bounds)
print(rst_dw.bounds)
```
-> BoundingBox(left=478000.0, bottom=3088490.0, right=502000.0, top=3108140.0)
-> BoundingBox(left=478000.0, bottom=3009890.0, right=598000.0, top=3108140.0)

The bottom and right bounds are moved by 5 times the array size. This is because the values of `Raster.width` and `Raster.height` used to calculate the bounds are the ones from the on-disk array and not the downsampled one.

Additionally, width and heights are different whether or not the data is loaded:
```
print(rst_dw.width)
rst_dw.load()
print(rst_dw.width)
```
yields 800 and 160 respectively.

I believe the properties width and height should be calculated differently, but I will check it does not break any test.

adehecq:
- Resolves #451,
 - Tests for downsampling improved

### DIFF
--- a/geoutils/raster/raster.py
+++ b/geoutils/raster/raster.py
@@ -549,7 +549,10 @@ class Raster:
     def height(self) -> int:
         """Height of the raster in pixels."""
         if not self.is_loaded:
-            return self._disk_shape[1]  # type: ignore
+            if self._out_shape is not None:
+                return self._out_shape[0]
+            else:
+                return self._disk_shape[1]  # type: ignore
         else:
             # If the raster is single-band
             if len(self.data.shape) == 2:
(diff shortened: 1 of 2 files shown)


rhugonnet (approved):
Amazing! :star_struck: 
Short fix in the end and most of the work was to make the (infinitely) better tests (it's nice to see all raster examples used too), thanks!!

PR #454, opened 2024-01-26
Improve error messages and make `Raster` indexing/assigment NumPy compatible

### LINKED ISSUE #431: Add indexing with slices to `Raster.__getitem__`
It would be nice to be able to do `Raster[0:100, 0:100]` and to have a cropped Raster with the right georeferencing! But this doesn't work right now, the `__getitem__` function only works with a boolean array or a Mask, the way it's coded.

### LINKED ISSUE #433: Raise more intuitive errors for common operations
For instance, for an arithmetic operation such as a subtraction:
```python
raster_diff = raster1 - raster2
```
If the raster are not aligned, right now we raise 
```"Both rasters must have the same shape, transform and CRS."```

I've had feedback that this is not very intuitive for beginners using the package, maybe we should raise: 
```"Both rasters must have the same shape, transform and CRS. For example, use raster1.reproject(raster2) to reproject both on the same grid and CRS."```

We could start a list on Docs (same as for the consistency of function arguments #432?) of those error cases that might be useful to be more explicit for every function, it would save us a lot of helping people do simple things down the road!

### PR DESCRIPTION
**Pull request by rhugonnet**
This PR:
- Adds a consistent error message across operations requiring two raster or one raster and one array that are aligned (arithmetic, NumPy array interface, indexing, index assignment),
- Removes the cropping behaviour of [] and makes it fully consistent with NumPy (allows any index supported by NumPy: integer, slices "1:5", ellipsis "...", or new axes.
- Improves significantly the tests for all errors raised and new behaviour.

Mysterious behaviour of NumPy makes the tests fail only on 4 specific array functions, everything else passing... :thinking: 

Resolves #433
Resolves #431

Opened #457

### DIFF
--- a/tests/test_raster.py
+++ b/tests/test_raster.py
@@ -3554,9 +3554,11 @@ class TestArithmetic:
         """
         # Rasters with different CRS, transform, or shape
         # Different shape
-        expected_message = "Both rasters must have the same shape, transform and CRS for an arithmetic operation. " \
-                           "For example, use raster1 = raster1.reproject(raster2) to reproject raster1 on the " \
-                           "same grid and CRS than raster2."
+        expected_message = (
+            "Both rasters must have the same shape, transform and CRS for an arithmetic operation. "
+            "For example, use raster1 = raster1.reproject(raster2) to reproject raster1 on the "
+            "same grid and CRS than raster2."
+        )
         with pytest.raises(ValueError, match=re.escape(expected_message)):
             getattr(self.r2, op)(self.r1_wrong_shape)
 
@@ -3569,11 +3571,13 @@ class TestArithmetic:
             getattr(self.r2, op)(self.r1_wrong_transform)
 
         # Array with different shape
-        expected_message = "The raster and array must have the same shape for an arithmetic operation. " \
-                           "For example, if the array comes from another raster, use raster1 = " \
-                           "raster1.reproject(raster2) beforehand to reproject raster1 on the same grid and CRS " \
-                           "than raster2. Or, if the array does not come from a raster, define one with raster = " \
-                           "Raster.from_array(array, array_transform, array_crs, array_nodata) then reproject."
+        expected_message = (
+            "The raster and array must have the same shape for an arithmetic operation. "
+            "For example, if the array comes from another raster, use raster1 = "
+            "raster1.reproject(raster2) beforehand to reproject raster1 on the same grid and CRS "
+            "than raster2. Or, if the array does not come from a raster, define one with raster = "
+            "Raster.from_array(array, array_transform, array_crs, array_nodata) then reproject."
+        )
         # Different shape, masked array
         with pytest.raises(ValueError, match=re.escape(expected_message)):
             getattr(self.r2, op)(self.r1_wrong_shape.data)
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: tests/test_raster.py [hunk 1]
```diff
@@ -940,29 +942,63 @@ def test_getitem_setitem(self, example: str) -> None:
...
         assert rst2.raster_equal(rst)
 
-        # Check that errors are raised for both indexing and index assignment
+        # -- Second, we test NumPy indexes (slices, integers, ellipses, new axes) --
+
+        # Indexing
+        assert np.ma.allequal(rst[0], rst.data[0])  # Test an integer
+        assert np.ma.allequal(rst[0:10], rst.data[0:10])  # Test a slice
+        assert np.ma.allequal(rst[np.newaxis, 0], rst.data[np.newaxis, 0])  # Test a new axis
+        assert np.ma.allequal(rst[...], rst.data[...])  # Test an ellipsis
```
adehecq, line 951: Just discovering ellipsis ! 😄

tests/test_raster.py
```diff
@@ -940,29 +942,63 @@ def test_getitem_setitem(self, example: str) -> None:
...
+
+        # Index assignment
+        rst[0] = 1
+        assert np.ma.allequal(rst.data[0], np.ones(np.shape(rst.data[0])))  # Test an integer
+        rst[0:10] = 1
+        assert np.ma.allequal(rst.data[0:10], np.ones(np.shape(rst.data[0:10])))  # Test a slice
+        rst[0, np.newaxis] = 1
```
adehecq, line 958: I have troubles understanding what this operation will do. Since `rst` has a dimension of 2 or 3, the second axis already exists... I assume it is equivalent to `rst[0] = 1` ?

tests/test_raster.py
```diff
@@ -987,10 +1023,6 @@ def test_crop(self, data: list[str]) -> None:
         r_cropped = r.crop(crop_geom2)
         assert r_cropped.raster_equal(r)
 
-        # Test with bracket call
-        r_cropped_getitem = r[crop_geom2]
```
adehecq, line 991: so this functionality is deleted now?

tests/test_raster.py
```diff
@@ -4008,3 +4059,67 @@ def test_ufunc_methods(self, method_str):
...
+        )
+        rst_wrong_crs = gu.Raster.from_array(ma, transform=self.transform, crs=32610, nodata=_default_nodata(ma.dtype))
+        rst_wrong_transform = gu.Raster.from_array(
+            ma, transform=self.wrong_transform, crs=4326, nodata=_default_nodata(ma_wrong_shape.dtype)
+        )
+
+        # Get ufunc
+        np_func = getattr(np, np_func_name)
+
+        # Strange errors happening only for these 4 functions...
```
adehecq, line 4084: Maybe just refer to the opened issue here?

adehecq (approved):
Nice that all these errors are consistently addressed now !
I only have a few minor comments.

tests/test_raster.py
rhugonnet: Yes! It was not in the documentation yet, and was not used much...

tests/test_raster.py
rhugonnet: It's adding a dimension, but because I only index 0 in the first one, it returns a 2D raster of shape (height, 1). Added a comment!

tests/test_raster.py
rhugonnet: Done!

rhugonnet:
I think it's because the lines don't exist anymore...
Since we added it a couple years ago, Coveralls doesn't seem to have picked up so well. Maybe it's time to switch to something that integrates more nicely with GitHub and saves us some time while reviewing!

rhugonnet:
Actually it looks like Coveralls is moving towards a GitHub integration too, maybe we can wait it out: https://coveralls.io/new-github-integration-beta

tests/test_raster.py
rhugonnet: The problem was that NumPy indexes can overlap with `crop_geom` types: If you pass [0, 1, 2, 3], you can't know if you are supposed to crop between [0, 2] and [1, 3], or index those 4 values?
I had repeated feedback from users that it would be more practical to be able to do raster[index] instead of raster.data[index] to manipulate the array.

adehecq: Good point ! I'm convinced now 😉

PR #455, opened 2024-01-26
Add `inplace` argument to all potential functions (`Raster` to modified `Raster`, same for `Vector`)

rhugonnet:
Ongoing

### DIFF
--- a/geoutils/raster/raster.py
+++ b/geoutils/raster/raster.py
@@ -2065,6 +2065,66 @@ np.ndarray or number and correct dtype, the compatible nodata value.
             newraster.tags["AREA_OR_POINT"] = "Area"
             return newraster
 
+    @overload
+    def reproject(
+            self: RasterType,
+            ref: RasterType | str | None = None,
+            crs: CRS | str | int | None = None,
+            size: tuple[int, int] | None = None,
+            bounds: dict[str, float] | rio.coords.BoundingBox | None = None,
+            res: float | abc.Iterable[float] | None = None,
+            nodata: int | float | tuple[int, ...] | tuple[float, ...] | None = None,
+            src_nodata: int | float | tuple[int, ...] | tuple[float, ...] | None = None,
+            dtype: DTypeLike | None = None,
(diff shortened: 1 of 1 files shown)


geoutils/raster/raster.py
```diff
@@ -2084,10 +2085,70 @@ def reproject(
...
+        force_source_nodata: int | float | None = None,
+        *,
+        inplace: bool = False,
+        silent: bool = False,
+        n_threads: int = 0,
+        memory_limit: int = 64,
+    ) -> RasterType | None:
+        ...
+
+    def reproject(
```
adehecq, line 2136: Why is there a need for a 4th overloading? The last two are identical apart from the star...

geoutils/raster/raster.py
```diff
@@ -2292,21 +2354,67 @@ def reproject(
...
         """
-        Shift the raster by a (x,y) offset.
+        Shift a raster by a (x,y) offset.
 
         The shifting only updates the geotransform (no resampling is performed).
 
         :param xoff: Translation x offset.
         :param yoff: Translation y offset.
         :param distance_unit: Distance unit, either 'georeferenced' (default) or 'pixel'.
+        :param inplace: Whether to modify the raster in-place.
```
adehecq, line 2417: The returns statement in missing.

adehecq (approved):
All good to me ! Thanks !! 
Just 6 additional lines of code missing test coverage according to [coveralls](https://coveralls.io/builds/65334491).

geoutils/raster/raster.py
rhugonnet: So it's 3 overloading + the normal function def

geoutils/raster/raster.py
rhugonnet: Good catch!

geoutils/raster/raster.py
adehecq: Those overloading are quite annoying, both to write and in term of space use... I hope they can find an easier syntax later.
Soon they will make up half of the code lines 😉

PR #458, opened 2024-01-30
Fix warnings and raise `UserWarning` as error in CI

### LINKED ISSUE #459: Test variable modified in `TestMask`
The test `TestMask:test_save` raises no `UserWarning` when run on its own, but does when run with all of `test_raster` or all tests.

That suggest that an input is likely modified by previous tests... which is not durable. 
A solution would be to transform all example data passed to tests into `pytext.fixtures`... See https://github.com/GlacioHack/xdem/issues/427

### LINKED ISSUE #460: Catching of `UserWarning` in tests degraded since homogenization of argument names
We should probably raise errors for all `UserWarning` not caught in tests, to notice and correct the test when we do a modification.

### LINKED ISSUE #461: Mask `dtype` not converted back to `bool` after reproject/crop/proximity operations
Converted to `uint8` to be supported in certain operations, but then not converted back

rhugonnet:
We had almost 100 warnings! Now all fixed except:
- A RunTime warning of NumPy that needs to be silenced in Rasterio and not here: 

```  
/home/atom/miniconda3/envs/geoutils-dev/lib/python3.11/site-packages/rasterio/dtypes.py:214: RuntimeWarning: overflow encountered in cast
    return np.allclose(values, values.astype(dtype), equal_nan=True)
```

We stopped catching a lot of warnings when modifying warning message during the hackathon (for instance, changing `dst_nodata` in `nodata`)! A good way to keep track of them from now on would be to consider `UserWarning` as an error in `pytest`: `pytest -W error::UserWarning`. I've added this as a pytest option in `pyproject.toml`, and so it will run by default during CI and locally when just writing `pytest`! :smile: 

Several warnings allowed to fix some deeper issues, for instance the modification of a `Mask`'s `dtype` without converting back in some functions! :sweat_smile: 

Also, I added a `NeedToImplementWarning` class (subclass of `FutureWarning`, to avoid triggering errors) to raise warning for us to remember to update the interface to GeoPandas as they add new vector methods.

And minor fixes to the API.

**EDIT: Change made here instead of #462 by mistake, but doesn't change much (independent)**
We mirror the script update in xDEM to parse package versions in CI to avoid failure in Windows and be more flexible in the future. See https://github.com/GlacioHack/xdem/pull/439. See #462 for justification.

Resolves #460 
Resolves #459 
Resolves #461


tests/test_geoviewer.py
```diff
@@ -40,9 +41,14 @@ def test_geoviewer_valid_1band(capsys, monkeypatch, filename, option):  # type:
     # To avoid having the plots popping up during execution
     monkeypatch.setattr(plt, "show", lambda: None)
 
+    # The everest example will raise errors
```
adehecq, line 44: Why is that? I don't get that warning when calling geoviewer locally on that dataset... Is this a geoutils' warning?

tests/test_raster.py
```diff
@@ -3072,8 +3092,19 @@ def test_save(self, mask: gu.Mask) -> None:
         # Save file to temporary file, with defaults opts
         temp_file = os.path.join(temp_dir.name, "test.tif")
         mask.save(temp_file)
-        saved = gu.Raster(temp_file)
-        assert mask.astype("uint8").raster_equal(saved)
+        saved = gu.Mask(temp_file)
+
+        # TODO: Generalize raster_equal for masks?
```
adehecq, line 3097: does it not work automatically for Masks?

tests/test_geoviewer.py
rhugonnet: Good point! I traced it back, and it's only when the `nodata=99` is used because this calls `set_nodata` in geoviewer, forcing a value that already exists in the raster.
I added a condition to filter the warning only for that specific nodata case in the tests!

tests/test_raster.py
rhugonnet: Opened #466 and looking into it...

PR #462, opened 2024-01-31
Pin max versions of all major dependencies for future compatibility

### LINKED ISSUE #426: Pin max major version number for dependencies to ensure durable compatibility
Right now our install is not strict enough, will break when a dependency releases a new major release.
To avoid this, see for instance some guidelines for NumPy 2 coming soon: https://pythonspeed.com/articles/numpy-2/

rhugonnet:
Pinning max major version only for future compatibility when they release a new version, we shouldn't have to update the max versions very often (and when we do, our functionalities will very likely require updating too because of breaking changes of major versions, so it's beneficial in any case! :smile: ).

Fixing `pytest` to `7.*` to avoid errors from `pytest` extension packages (lazy fixtures).

**EDIT: The change below is done in #458** 
Also mirroring the script update in xDEM to parse package versions in CI to avoid failure in Windows and be more flexible in the future.
**Reminder of the original reason we have to do all this:** Unfortunately the Python version cannot be pinned easily when building a conda cached environment in CI from the command line options (and having a cache save us up to 15min of build time on Mac/Windows, so we keep that for sure!). To circumvent the issue, we need to re-create a temporary YAML file that pins the python version (previously done by parsing the YAML argument into command line, but failed on Windows for symbols such as `<` that we need to pin max versions).

Resolves #426


adehecq (approved):
Great, this should make things more stable in the longer term. But is there an easy way to test whether or not a newer version that specifies exist? Apart from checking each module individually...
Should we have an environment file with no max version pinned, to be run as tests occasionally?

Arf, segmentation faults are not nice... Is it related to this PR only?

rhugonnet:
Found the cause, `pytest` just released `8.0` and it breaks things in `pytest-lazy-fixtures`... Because they did not pin their max major version of `pytest`!
This is very much in the theme of this PR :sweat_smile:

rhugonnet:
Actually, I realized that the parsing of versions such as `rasterio>1.3,<2` doesn't work on CI with Windows due to the script we used.
This was fixed in xDEM in https://github.com/GlacioHack/xdem/pull/439 by generating a temporary YAML during CI instead of parsing the package versions from the command line when building the `mamba` environment, I'm mirroring the changes here.

rhugonnet:
Made a mistake, mirroring the xDEM change in #458 instead. Doesn't change much, it's independent from the changes here.

rhugonnet:
Looks like it's a dependency problem upstream from testing packages...
Let's wait it out a couple days and see again

rhugonnet:
Finally found it!
Two causes: #473, and the latest `geopandas`, I forced an earlier version.

Issue #453, opened 2024-01-26
Test in test_reproject is never run


adehecq:
The test starting at [this line](https://github.com/GlacioHack/geoutils/blob/c9c9a843e95d34081a47a1dd3e36f0ba6dc8e712/tests/test_raster.py#L1347) is never run, because at the beginning of the test, we made sure nodata is never None.
Additionally, to really make that test, we would need to ensure that gaps are present in the data (this is not guaranteed right now) and that it can lead to possible overshoot for interpolation other than nearest neighbor.

PR #465, opened 2024-01-31
Two fixes in Raster.reproject

adehecq:
- solves #453 
- fixes an issue with reproject not enforcing bounds in some cases due to floating point precision (see [issue](https://github.com/rasterio/rasterio/issues/3016) raised in rasterio). Extracted window need to be rounded to nearest integer rather than up.

Weirdly the second change does not make any difference for tests. I guess in all tested cases, the rounding does not make any difference. This was probably an edge case that I encountered.
Not sure if there is any test to add for this...

### DIFF
--- a/tests/test_raster.py
+++ b/tests/test_raster.py
@@ -1381,6 +1381,13 @@ class TestRaster:
         assert r_test.crs.to_epsg() == 4326
 
         # -- Additional tests --
+        # First, make sure dst_bounds extend beyond current extent to create nodata
+        dst_bounds = rio.coords.BoundingBox(
+            left=bounds[0], bottom=bounds[1] - r.res[0], right=bounds[2] + 2 * r.res[1], top=bounds[3]
+        )
+        r_test = r.reproject(bounds=dst_bounds)
+        assert np.count_nonzero(r_test.data.mask) > 0
+
         # If nodata falls outside the original image range, check range is preserved (with nearest interpolation)
         r_float = r.astype("float32")  # type: ignore
         if r_float.nodata is None:
--- a/tests/test_raster.py
+++ b/tests/test_raster.py
@@ -1390,17 +1390,17 @@ class TestRaster:
 
         # If nodata falls outside the original image range, check range is preserved (with nearest interpolation)
         r_float = r.astype("float32")  # type: ignore
-        if r_float.nodata is None:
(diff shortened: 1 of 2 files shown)


tests/test_raster.py
```diff
@@ -1381,6 +1381,13 @@ def test_reproject(self, example: str) -> None:
         assert r_test.crs.to_epsg() == 4326
 
         # -- Additional tests --
+        # First, make sure dst_bounds extend beyond current extent to create nodata
```
rhugonnet, line 1384: So this is a replacement of the test never run? Should we delete that test that is never run then?

rhugonnet (approved):
If you still have the edge case that you can re-use to test the rounding, it'd be a nice quick addition to the tests.
Otherwise, all good for me!

tests/test_raster.py
adehecq: Sorry, there's a mistake I think. `r_test` should be used in the following lines, because I made sure it contains nodata. Before these lines were added, there was actually no gap so the test was pointless...
I'll check again!

adehecq: In the end, no it was correct, `r_test` is just used to check that reproject with these bounds will creates nodata in the following tests, not used directly in those tests.
I had to follow several layers of git blame to track down the origin of this test. It was created [here](https://github.com/GlacioHack/geoutils/pull/223), but then modified a few times, particularly [here](https://github.com/GlacioHack/geoutils/pull/284) by you @rhugonnet 😜, which made the test lose its usefulness...
In short, it was just added at a point where rasterio's default nodata values would add very large values. I'm not sure this test is super useful nowadays, but I will keep it just in case we break something very badly.

adehecq:
> If you still have the edge case that you can re-use to test the rounding, it'd be a nice quick addition to the tests. Otherwise, all good for me!

Unfortunately, I only have partial information in the rasterio issue I raised, so I cannot fully reproduce now. I will ignore it for the moment.

rhugonnet:
Perfect! Also, your tests are passing... While it's still failing on #462 with a segmentation error, even if I reverted back nearly everything, I don't understand what's going on... :exploding_head:

rhugonnet:
Actually I know! Your PR didn't have any environment file change, so it's still using the cached environment that was working...
OK I'll keep investigating in #462

PR #467, opened 2024-01-31
Check segmentation fault cause

rhugonnet:
Just to double-check it doesn't come from us


rhugonnet:
Indeed it does not! Fails with the same environment as always, also locally... Will wait it out in #462

PR #468, opened 2024-02-01
Make `raster_equal` accept `False` mask values of masked arrays

### LINKED ISSUE #466: `Mask` read from file is slightly different than saved `Mask`
This is why there was a failure in `TestMask::test_save` in #458, investigating...

rhugonnet:
And fixes `test_save` in `TestMask`, the difference came from the fact that in-memory rasters don't necessarily have the same masked `.data.data` as the ones read from disk, as they are replaced by the `nodata` value when writing.

Resolves #466


rhugonnet:
All passing locally, need to wait for the Segmentation Fault issue of Pytest to be solved for CI...

adehecq (approved):
Nice ! All good.

PR #472, opened 2024-02-01
Make `astype` set new `nodata` by default, and `data.setter` check for unmasked nodata

### LINKED ISSUE #394: `Raster.nodata` value is not updated during `astype()`
We should do this automatically?

### LINKED ISSUE #469: Creation or numerical operation can create unmasked `nodata` value in `Raster` without notice
Two cases:

- We can feed a masked array with -9999 that are not masked to `from_array()`, and it will create a `Raster` without the -9999 masked even if `nodata` is declared to be -9999,
- We can do have a `Raster` with a -9998 value, and a `nodata` of -9999. If we do `Raster - 1`, a value equal to the nodata will be created without notice.

### LINKED ISSUE #470: Remove `masked` option in `Raster.__init__` as it is no longer supported?
Doing `Raster(file, masked=False)` is not possible, as we force a masked array everywhere.

Given that the Xarray accessor will give the option of using a NaN array, maybe we don't need to do this in `Raster` as well?

### LINKED ISSUE #471: Arithmetic overloading replaces existing nodate with `None` for certain operation with `int` data types
For example, `int8` data with `nodata=126`:
```python
rst = rst + 1
```
Nodata is changed to `None`

rhugonnet:
That little change on `astype` ended up being a lot bigger than I anticipated!

I realized that:
- New nodata value could be created at any point without being masked, see #469,
- Some arithmetic overloading casted nodatas to `None`, see #471,
- We still had the `masked` option for `Raster` which is deprecated, see #470.

# What's changed

1. At the end of `data.setter`, we now check that there isn't any nodata value that is unmasked, otherwise we mask them (in case it is user input and not numerics) **BUT RAISE A WARNING** and recommend the user to either: Mask his array when creating the raster (if that was what he was doing), or Change the dtype to not have numerical operation derive the nodata value (if that was what he was doing). This mostly happens with `int`, and should be very easy to silence using a `float` array with properly masked values (or NaNs).
2. Changes the `nodata` when using `astype` by default. If the nodata is already in the array, it will be detected by `set_nodata()`.

Surprisingly, those two "big" changes didn't create so many new errors in tests, and actually rather allowed to identify a couple bugs! :smile: 
The rest is adding new tests, silencing or catching warnings, and updating the warning for the nodata setter (that were still not so clear when I re-read them).

Resolves #471
Resolves #470 
Resolves #469 
Resolves #394

### DIFF
--- a/tests/test_multiraster.py
+++ b/tests/test_multiraster.py
@@ -101,7 +101,9 @@ class TestMultiRaster:
         """Test stack_rasters"""
 
         # Silence the reprojection warning for default nodata value
-        warnings.filterwarnings("ignore", category=UserWarning, message="New nodata value cells already exist in the data array.*")
+        warnings.filterwarnings(
+            "ignore", category=UserWarning, message="New nodata value cells already exist in the data array.*"
+        )
         warnings.filterwarnings("ignore", category=UserWarning, message="For reprojection, nodata must be set.*")
 
         # Merge the two overlapping DEMs and check that output bounds and shape is correct
@@ -171,7 +173,9 @@ class TestMultiRaster:
         # Merge the two overlapping DEMs and check that it closely resembles the initial DEM
 
         # Silence the reprojection warning for default nodata value
-        warnings.filterwarnings("ignore", category=UserWarning, message="New nodata value cells already exist in the data array.*")
+        warnings.filterwarnings(
+            "ignore", category=UserWarning, message="New nodata value cells already exist in the data array.*"
+        )
         warnings.filterwarnings("ignore", category=UserWarning, message="For reprojection, nodata must be set.*")
 
         # Ignore warning already checked in test_stack_rasters
--- a/tests/test_multiraster.py
+++ b/tests/test_multiraster.py
@@ -101,7 +101,7 @@ class TestMultiRaster:
         """Test stack_rasters"""
 
         # Silence the reprojection warning for default nodata value
-        warnings.filterwarnings("ignore", category=UserWarning, message="New nodata value found in the data array.*")
+        warnings.filterwarnings("ignore", category=UserWarning, message="New nodata value cells already exist in the data array.*")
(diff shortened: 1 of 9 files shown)


tests/test_multiraster.py
```diff
@@ -101,7 +101,9 @@ def test_stack_rasters(self, rasters) -> None:  # type: ignore
         """Test stack_rasters"""
 
         # Silence the reprojection warning for default nodata value
-        warnings.filterwarnings("ignore", category=UserWarning, message="New nodata value found in the data array.*")
+        warnings.filterwarnings(
```
adehecq, line 104: How come sometimes you have these changes in syntax, that are not related to your PR, despite the linters being enforced?

tests/test_raster.py
```diff
@@ -574,6 +574,30 @@ def test_data_setter(self, dtype: str, nodata_init: str | None) -> None:
...
+
+        # Check that it can happen during a numerical operation
+        ma = np.ma.masked_array(arr, mask=mask)
+        ma.data[0, 0] = 125
+        ma.mask[0, 0] = False
+        raster = gu.Raster.from_array(data=ma, transform=transform, crs=None, nodata=126)
+        with pytest.warns(UserWarning, match="Unmasked values equal to the nodata value found in data array.*"):
+            # Issue during numerical operation
+            new_raster = raster + 1
+            print(new_raster.data[0, 0])
```
adehecq, line 597: remove the print statements?

adehecq (approved):
Good job finding all these bugs !
I went quickly through the tests, but the changes make sense.

tests/test_multiraster.py
rhugonnet: It's related to the PR but hard to see! There is a "cell" added in the sentence for the warning :wink:

tests/test_raster.py
rhugonnet: done!

PR #476, opened 2024-02-05
Add dependabot and pre-commit CI for automatic versioning updates

### LINKED ISSUE #327: Add dependabot to update GitHub actions automatically
There is a way to automate GitHub actions to keep the code up-to-date. I'm not sure exactly what actions are automated, but here is the documentation on it: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot.
I believe this can be used for example to keep the pre-commit up-to-date, as in #325.

### LINKED ISSUE #475: Update CI action versions that are deprecated
See warning in CI:

> Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/checkout@v3, conda-incubator/setup-miniconda@v2, actions/cache@v3, coverallsapp/github-action@master. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.

rhugonnet:
Resolves #475
Resolves #327 

However issue #327 by @adehecq was more about updating `pre-commit` automatically.

For this, dependabot does not work, see https://github.com/dependabot/dependabot-core/issues/2040.
But we could use either this: https://github.com/browniebroke/pre-commit-autoupdate-action?tab=readme-ov-file or better (= recommended by the previous link for an open project), this https://pre-commit.ci/.

However, it might be better to update pre-commit ourselves once a year or so, because updating it frequently means solving typing issues all the time instead of all at once (not sure that's so efficient), or running into bugs from minor versions changes of `pre-commit`. And we don't care so much if we're not at the "latest" on typing rules, as long as the package is properly typed with all the previous ones already, no?


rhugonnet:
OK, I've added https://pre-commit.ci/ (with the same setup as in `pandas`) but tuned it to update only quarterly instead of monthly, this seems more reasonable! :smile:

atedstone (approved):
That's some nice administration ;)

adehecq:
> OK, I've added https://pre-commit.ci/ (with the same setup as in `pandas`) but tuned it to update only quarterly instead of monthly, this seems more reasonable! 😄

Nice, very smart !

PR #484, opened 2024-02-07
Make `interp_points` robust to non-equal grids and impractical default SciPy arguments

### LINKED ISSUE #483: Several issues with `interp_points`
The `Raster.interp_points` function has several urgent issues:
1. The `mode` argument is never used, and actually defines the wrong thing (it is the `order` argument that defines the type of interpolation in `map_coordinates`, `mode` is for behaviour around the margin of the image),
2. The `prefilter` argument of `map_coordinates` runs by default: https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.map_coordinates.html#scipy-ndimage-map-coordinates. This spline pre-filter will actually expand NaNs for up to **20 PIXELS** (tested with `order=3` on this data: https://github.com/ICESat2-SlideRule/coregistration/blob/master/xdem_test.ipynb)
3. We need to clarify/merge behaviour with `value_at_coords`, see #395.

rhugonnet:
This PR:
- Makes the behaviour of `interp_points` consistent between equal and regular grid by using both `map_coordinates` and `interpn` from SciPy depending on the grid,
- Removes the default `prefilter` argument of `map_coordinates` that created large data gaps,
- Corrects the `method` argument for `map_coordinates` by mapping it into an `order` argument,
- Adds detailed tests with synthetic data to ensure all results are consistent, in addition to the ones existing in `test_xy2ij`.

Resolves #483 

TO-DO:

- [x] Test the consistency of the `AREA_OR_POINT` argument between the methods.

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -12,4 +12,4 @@ jobs:
     steps:
     - uses: actions/checkout@v4
     - uses: actions/setup-python@v5
-    - uses: pre-commit/action@v3.0.0
+    - uses: pre-commit/action@v3.0.1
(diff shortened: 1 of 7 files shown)


rhugonnet:
Merging this bug fix/improvement as it has pretty serious impact on xDEM, can be reviewed a posteriori.

PR #492, opened 2024-02-16
Add `PointCloud` subclass of `Vector`

### LINKED ISSUE #463: Add a `PointCloud` subclass of `Vector` to consistently deal with points?
Every type of point data (tuple of lists, list of tuples, array, geodataframe, etc) could be converted consistently into a geodataframe of only points that could fit into a `PointCloud()` subclass of `Vector`?

It could be the default input/output of:
- `Raster.to_points()`,
- `Raster.interp_points()`,
- `Raster.value_at_coords()`,
- etc...

And serve to have an `ElevationPointCloud` class in xDEM: https://github.com/GlacioHack/xdem/issues/463

rhugonnet:
This PR adds the `PointCloud` class to facilitate interfacing with this special sub-type of vector data very common in geospatial analysis.
In short, a point cloud is a subclass of `Vector` **containing only 2D point geometries and associated to a main data column** (+ optionally other auxiliary data columns). This main data column represents the values `.data` of the point cloud, that can be compared to the `.data` of a raster format (which is impossible for a typical vector), or of that of another point cloud if some of their coordinates are the same, or within close tolerance.

Additionally, all functionalities of `Raster` using `.data` that just need coordinates can be ported to a `PointCloud` (array interface with NumPy, subsampling, binning, zonal stats, variography, etc).

## Summary

The idea behind the implementation of the `PointCloud` class is two-fold:
1. Substantially improve the interfacing of point clouds with vectors or raster, which are currently based on passing volatile arguments of "coordinates" (that could be defined with many shapes/formats) or returning volative 1D arrays of data values associated with no coordinates: https://geoutils.readthedocs.io/en/stable/raster_vector_point.html#rasterpoint-operations.
2. Natively support large point cloud files of type LAS that are also used widely, are not supported by GeoPandas (didn't find a single issue on this, not planned at all), which can benefit from the same `load` mechanism as `Raster` and support chunked reading.

For this, we add the new optional dependency `laspy` for reading and writing LAS-type format. And we override some of the behaviour of `Vector` in `PointCloud` to allow for a `load()` functionality.

## Details

This PR adds the `PointCloud` class, including:
- A new `data_column` attribute containing the data column name of the point cloud,
- A new `data` attribute corresponding to the data of the point cloud,
- A new `is_loaded` attribute and `load()` method for when data is read from a LAS file,
- A new `point_count` attribute returning the number of points in the point cloud,
- A new set of `from_array()`, `from_tuples()` and `from_xyz()` methods to easily create a point cloud from different inputs, and their equivalent `to_array()`, `to_tuples()` and `to_xyz()`,
- A new `pointcloud_equal()` method simply checking `vector_equal()` and that the `data_column` is equal as well,
- A new `georeferenced_coords_equal()` method checking if the coordinates of the point clouds are the same (mirrors `georeferenced_grid_equal()` in `Raster`),
- Wraps the existing `subsample()` and `get_stats()` functionalities as classes method,
- Wraps the existing `grid()` functionality as a class method.

The new `PointCloud` class overrides the attributes `ds`, `crs`, `bounds` and `columns` to allow loading only metadata, and implicitly loading the point cloud.
We also add an `__array_interface__` and `__array_ufunc__` to `PointCloud`, as well as core arithmetic operations, to support those between similar point clouds. Finally, we add indexing and index assigment using a new `PointCloudMask` objects.

This PR updates existing input/output of some functions:
- `Raster.to_pointcloud()` now returns a `PointCloud` instead of a `Vector`,
- `Vector.create_mask()` now accepts a `PointCloud` in addition to a `Raster`, to return a masked `PointCloud`,
- `Raster.interp_points()` and `Raster.reduce_points()` now accept a point cloud as input as coordinates to use for interpolation, and can optionally return a point cloud for the output interpolated values + coordinates,
- `Raster.from_regular_pointcloud()` now accepts a `PointCloud` input,

## To-do for this PR

Small changes:
- [x] Add LAS file and Point-only vector file in `geoutils-data` to run tests,
- [x] Update `Raster.from_regular_pointcloud()` to accept a `PointCloud` input,
- [x] Update `Raster.interp_points()` and `Raster.reduce_points()` to accept a `PointCloud` as input for coordinates,
- [x] Update `Vector.create_mask()` to accept a `PointCloud` the manipulation of masking and booleans for point clouds,
- [x] Mirror every function relying only on adimensional `.data` (same whether 1D or 2D) to point cloud: `subsample()`, `get_stats()` (and function that work on 1D sets of coordinates like `variogram()`, once moved to GeoUtils).
- [x] Add `__array_interface__` support for point clouds with the same coordinates, simply pointing to that of `self.ds[data_column]`?
- [x] Populate documentation page "The georeferenced point cloud", update the "Raster-Point interface" sections, update Quick start and Feature overview, and add a gallery example?
- [x] Add indexing and index assignment for point clouds,
- [x] Add `plot()` function overriding `Vector.plot()` to always plot point-like with the data column,
- [x] Override `Vector.columns` for unloaded point clouds?
- [x] Finalize `write_laspy` function to save to LAS/LAZ,
- [x] Decide on namings: "z_column" or "data_columns", "nb_points" or "count" or "count_points".
- [x] Discuss optional point cloud return for `Raster.subsample()` and `PointCloud.subsample()`? Opened #707.
- [x] Discuss mask structuration: keep subclass or move to optional boolean `dtype` behaviour? Opened #706.

## Potential bigger change?

This last change could be to (**AFTER DISCUSSION: NOT FOR NOW, LEFT EXPLICIT TO USER**):
- Allow for arithmetic operations between `Raster` and `PointCloud`? (e.g., defaulting to using `interp_points` at the `PointCloud` coordinates, the comparison method could be tweaked in `geoutils.config`),

It could work as:
```python
pc = PointCloud("pc_example.shp")
raster = Raster("rast_example.tif")

# An arithmetic operation between a raster and point cloud would always return a PointCloud
pc_diff = pc - raster

# Users could change the behaviour in the config
geoutils.config["raster_point.comparison_method"] = "reduce_points"
geoutils.config["raster_point.resampling"] = "cubic"
```

The only problem I see is that interpolation is not ideal for a point cloud much denser than a raster (e.g., a lidar point cloud). In that case, it's better to grid the point cloud into a raster, and compare the two rasters. But that is computationally-intensive, so happening under-the-hood is not ideal.
We could have a criteria based on point cloud density relative to the raster size:
- Above a certain density, interpolate raster at the points; below that, grid the points,
- Or always follow the same behaviour, but raise a warning if the point density is not ideal?

Or we do not add such a functionality and always leave this to the user, which remains fairly short as long as we support `Raster`/`Raster` arithmetic and `PointCloud`/`PointCloud` arithmetic:
```python
rast_from_pc = pc.grid(rast)
my_op = (rast + rast_from_pc) / 4

pc_from_rast = rast.interp_points(pc)
my_op = (pc_from_rast + pc) / 4
```
And add a public function to calculate relative point density to help users decide when need be?

## Future functionalities

- Add pointcloud to pointcloud comparison with non-equal coordinates by grouping closest points within a certain tolerance,
- Add other common gridding/interpolation methods for points, or for raster-point comparison: IDW, kriging.
- Mirror the `PointCloud` class in a GeoPandas accessor (technically Pandas accessor, see here: https://github.com/geopandas/geopandas/pull/1952#issuecomment-1367889607).

Resolves #463


rhugonnet:
@adehecq @atedstone @erikmannerfelt This PR is now advanced enough to get your feedback! :blush:  
I think it's good we discuss it now before moving further forward, documenting and finalizing, to be sure we agree on all the concepts and behaviour introduced! I updated the description above to a good level of detail, which should allow you to grasp the idea behind this new object :wink:.

All tests passing locally with a LAS file I have. Will need to add one to `geoutils-data` for CI to pass.

atedstone:
@rhugonnet , this looks exceptionally well thought through. I read your descriptions and the changed files,but have not tried out the functionality myself as I don't immediately have files lying around that would be a good test case. But from my side, this looks good to progress to finishing the outstanding tasks you listed.

Concerning the bigger possible change about arithmetic ops between Raster and PointCloud: my instinct would be to require the user to decide, so that the comparison is explicit and clear.

geoutils/__init__.py
```diff
@@ -2,10 +2,12 @@
 GeoUtils is a Python package for the analysis of geospatial data.
 """
 
-from geoutils import examples, pointcloud, projtools, raster, vector  # noqa
+from geoutils import examples, projtools  # noqa
```
adehecq, line 5: so we don't want to load the raster submodule explicitely anymore? Is there nothing else to get apart from the Raster and Mask classes? (I'm not fully familiar with the new structure yet)

geoutils/raster/raster.py
```diff
@@ -28,6 +28,7 @@
 from rasterio.enums import Resampling
 from rasterio.plot import show as rshow
 
+import geoutils as gu
```
adehecq, line 31: Does it not lead to a circular import?? Would `from geoutils import Vector` be better?

geoutils/pointcloud/pointcloud.py
```diff
@@ -1 +1,393 @@
...
+from typing import Any, Iterable, Literal
+
+import geopandas as gpd
+import numpy as np
+import pandas as pd
+from pyproj import CRS
+from rasterio.coords import BoundingBox
+from shapely.geometry.base import BaseGeometry
+
+import geoutils as gu
```
adehecq, line 17: same comment as above

geoutils/pointcloud/pointcloud.py
```diff
@@ -1 +1,393 @@
...
+from geoutils._typing import ArrayLike, NDArrayNum, Number
+from geoutils.interface.gridding import _grid_pointcloud
+
+# from geoutils.raster.sampling import subsample_array
```
adehecq, line 21: to delete maybe?

geoutils/pointcloud/pointcloud.py
```diff
@@ -1 +1,393 @@
...
+
+        crs = f.header.parse_crs(prefer_wkt=False)
+        nb_points = f.header.point_count
+        bounds = BoundingBox(left=f.header.x_min, right=f.header.x_max, bottom=f.header.y_min, top=f.header.y_max)
+        columns_names = pd.Index(list(f.header.point_format.dimension_names))
+
+    return crs, nb_points, bounds, columns_names
+
+
+# def _write_laspy(filename: str, pc: gpd.GeoDataFrame):
```
adehecq, line 65: why is this commented? Not working?

geoutils/pointcloud/pointcloud.py
```diff
@@ -1 +1,393 @@
...
+        :param data_column: Name of main data column defining the point cloud.
+        """
+
+        self._ds: gpd.GeoDataFrame | None = None
+        self._name: str | None = None
+        self._crs: CRS | None = None
+        self._bounds: BoundingBox
+        self._data_column: str
+        self._data: NDArrayNum
+        self._nb_points: int
```
adehecq, line 118: Intuitively I would have called this "count", which I think is often used by other tools. But this may be ambiguous with the `Raster.count` which is about band count.

geoutils/pointcloud/pointcloud.py
```diff
@@ -1 +1,393 @@
...
+
+        :param filename_or_dataset: Path to vector file, or GeoPandas dataframe or series, or Shapely geometry.
+        :param data_column: Name of main data column defining the point cloud.
+        """
+
+        self._ds: gpd.GeoDataFrame | None = None
+        self._name: str | None = None
+        self._crs: CRS | None = None
+        self._bounds: BoundingBox
+        self._data_column: str
```
adehecq, line 116: Just a general note. I don't find the names "data_column" and "all_columns" very intuitive. How did you pick the names? Hence I am not sure to understand what is which, even with the PR description. This will most likely be clearer after reading the code, but it should be fairly intuitive to the user without searching through doc or code.  

I don't know what is the most used term for this.
In OGR this was named "fields" and similarly in QGIS. So in the old geoutils, we stuck to this term. 
Would it also not be easier to save a "fields" attribute as a panda dataframe, so the column names and data can be retrieved from one object instead of two?

geoutils/pointcloud/pointcloud.py
```diff
@@ -1 +1,393 @@
...
+    @data_column.setter
+    def data_column(self, new_data_column: str) -> None:
+        self.set_data_column(new_data_column=new_data_column)
+
+    def set_data_column(self, new_data_column: str) -> None:
+        """Set new column as point cloud data column."""
+
+        if new_data_column not in self.all_columns:
+            raise ValueError(
+                f"Data column {new_data_column} not found among columns, available columns "
```
adehecq, line 240: Maybe add a break. "not found among columns, available columns..." -> "not found among columns. Available columns..."

geoutils/pointcloud/pointcloud.py
adehecq: Ok, looking at the code below it is a bit more clear the meaning of the "columns" attributes. I understand that it originates from Pandas columns. 
I also realize that in the Vector class (and so in PointCloud too), the field attributes are not directly accessible from the Vector object, but from `Vector.ds` only. There is also no equivalent to the "all_columns" attribute in Vector.
Sorry I don't have a perfect solution to give, but maybe we should discuss this more and make sure everything is clear and consistent.

geoutils/pointcloud/pointcloud.py
```diff
@@ -1 +1,393 @@
...
+        return all([vector_eq, data_column_eq])
+
+    def grid(
+        self,
+        ref: gu.Raster | None,
+        grid_coords: tuple[NDArrayNum, NDArrayNum] | None,
+        resampling: Literal["nearest", "linear", "cubic"],
+        dist_nodata_pixel: float = 1.0,
+    ) -> gu.Raster:
+        """Grid point cloud into a raster."""
```
adehecq, line 359: I think the arguments should be defined, no?

geoutils/pointcloud/pointcloud.py
```diff
@@ -1 +1,393 @@
...
+            self.ds,
+            grid_coords=grid_coords,
+            data_column_name=self.data_column,
+            resampling=resampling,
+            dist_nodata_pixel=dist_nodata_pixel,
+        )
+
+        return gu.Raster.from_array(data=array, transform=transform, crs=self.crs, nodata=None)
+
+    # def subsample(self, subsample: float | int, random_state: int | np.random.Generator | None = None) -> PointCloud:
```
adehecq, line 381: not working?

geoutils/vector/vector.py
```diff
@@ -64,37 +63,105 @@ class Vector:
...
+    def name(self) -> str | None:
+        """Name on disk, if it exists."""
+        return self._name
+
+    @property
+    def geometry(self) -> gpd.GeoSeries:
+        return self.ds.geometry
+
+    @property
+    def columns(self) -> pd.Index:
```
adehecq, line 152: Should it not be consistent with the `PointCloud` class, which uses `all_columns`?

tests/test_pointcloud/test_pointcloud.py
```diff
@@ -1 +1,238 @@
...
+
+        pc = PointCloud(self.gdf1, data_column="b1")
+        # If the data column does not exist
+        with pytest.raises(ValueError, match="Data column column_that_does_not_exist not found*"):
+            pc.data_column = "column_that_does_not_exist"
+        # Equivalent syntax
+        with pytest.raises(ValueError, match="Data column column_that_does_not_exist not found*"):
+            pc.set_data_column("column_that_does_not_exist")
+
+    def test_pointcloud_equal(self) -> None:
```
adehecq, line 155: This test is empty 😉

tests/test_pointcloud/test_pointcloud.py
```diff
@@ -1 +1,238 @@
...
+        """Test errors raised during creation with array."""
+
+        array = np.ones((4, 5))
+        with pytest.raises(ValueError, match="Array must be of shape 3xN or Nx3."):
+            PointCloud.from_array(array=array, crs=4326)
+
+    def test_from_tuples(self) -> None:
+        """Test building point cloud from list of tuples."""
+
+        pc1 = PointCloud(self.gdf1, data_column="b1")
```
adehecq, line 180: the fact that you have to specify the data column each time makes me think. By default, if there is only one column, should it not pick that one?

adehecq (approved):
That is a very neat and useful addition to the existing set of classes, thanks a lot !!
Overall, this is well thought and I agree with most aspects. I only have some concerns about the attribute names "data_column" and "all_columns" (see my specific comment about it). 
I will reply to specific questions separately.

adehecq:
> Or we do not add such a functionality and always leave this to the user, which remains fairly short as long as we support `Raster`/`Raster` arithmetic and `PointCloud`/`PointCloud` arithmetic:

Regarding the arithmetic operations, I would make sure they work within objects of the same class, but not within objects of different classes. Otherwise, this would force us to make some implicit choices and we don't want "naive" users to do things without understanding it. Users need to understand what it implies to take the difference between a DEM and PC. I think we should only facilitate these operations, and as you show, if it takes only 2 lines to do it, that's perfectly fine. It should just be documented so that the users find the steps easily. 
It is the same question as to allow differencing two Rasters that are on a different grid. We could make some implicit assumptions and allow it, but we don't.

adehecq:
> * Add pointcloud to pointcloud comparison with non-equal coordinates by grouping closest points within a certain tolerance,
> 
> * Add other common gridding/interpolation methods for points, or for raster-point comparison: IDW, kriging.
> 
> * Mirror the `PointCloud` class in a GeoPandas accessor (technically Pandas accessor, see here:

All of this would be neat, especially the IDW and kriging!

geoutils/__init__.py
rhugonnet: It's actually not needed, and triggers circular import issues.

geoutils/raster/raster.py
rhugonnet: It's the opposite: `import geoutils as gu` works without necessarily getting submodules, so perfect to add `gu.Vector` or `gu.Raster` for typing purposes without circular import issues.

geoutils/pointcloud/pointcloud.py
```diff
@@ -1 +1,393 @@
...
+            raise ValueError("The dataset of a vector must be set with a GeoSeries or a GeoDataFrame.")
+
+    @property
+    def crs(self) -> CRS:
+        """Coordinate reference system of the vector."""
+        # Overriding method in Vector
+        if self.is_loaded:
+            return super().crs
+        else:
+            return self._crs
```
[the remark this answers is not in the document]
rhugonnet: If it's loaded we need to go search in the `.ds` in case the CRS has changed. Otherwise it's the one on disk from `self._crs`.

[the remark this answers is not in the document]
rhugonnet: Added a comment to clarify (but same mechanism as in `Raster`).

geoutils/pointcloud/pointcloud.py
rhugonnet: Done

geoutils/pointcloud/pointcloud.py
rhugonnet: Done.

geoutils/pointcloud/pointcloud.py
rhugonnet: I've removed "all_columns" and simply override "columns" to stay consistent with Pandas/GeoPandas.
For "data_column": I don't like "z" too much because it's usually used for 3D points (elevation), while a "data" associated to 2D points is more generic.
We could use "z" for the elevation point cloud class?

geoutils/pointcloud/pointcloud.py
rhugonnet: Changed to `point_count` for consistency with LAS naming.

geoutils/pointcloud/pointcloud.py
rhugonnet: Finalized now.

rhugonnet:
Thanks @atedstone, yes it was almost there but got caught in other work!
I've now almost finalized the documentation, and accounted from all previous comments.

Here's the link to the doc: TO ADD

Tagging @adehecq @adebardo @belletva for info/potential second review.
I don't think we need another full review on this, only your opinion on a couple points. Feel free to do a full review if you still like, I'll be merging in ~2 weeks top. Then I think we should release a `0.2`, and follow-up in xDEM with the `EPC` class and its own `0.2`.

In addition to the new class, here's a summary of changes that you probably don't expect, to check if it's OK with you:
1. I had to rename the `Mask` class into `RasterMask` to accomodate a `PointCloudMask` class, otherwise it was impossible to define one. We could think about removing the `Mask` class altogether in the future, I opened a discussion here: https://github.com/GlacioHack/geoutils/issues/706. The `Mask` class still works for now, but raises a deprecation warning.
2. Because point clouds can also have statistics computed on their `.data` (as well as arithmetics), I moved the `get_stats` core functionalities out of the `Raster` class (which we had discussed at the time) into a new `stats/` submodule, along with the `sampling` that lived in raster/ as well. Now `get_stats` (should we rename to simply `stats()`?) can be called from `Raster` or `PointCloud` interchangeably. The documentation reflects this with a new "Statistics" page that contains the common functionality, while the "Raster/Point cloud" pages simply give a short example and point to it.
3. Many functions that used to return an array of values that were actually points (`interp_points`, `reduce_points`, `subsample`, `to_pointcloud`) now return a `PointCloud` by default. But those functions all have an `as_array` argument to return an array if desired.

**Last point, probably the most critical one where I'd be curious of your opinion:** This PR adds support for LAS/LAZ files as a type of point cloud. However, those are quite often specific to elevation data, contrary to other point clouds. Should we rather move the optional `laspy` dependency and LAS file support to the elevation point cloud `EPC` class in xDEM that is inheriting from this?
I'm OK with both.

rhugonnet:
Thanks @belletva for the quick reaction!

I'm merging now, as structure changes conflict with #446 and I don't want to duplicate the amount of work by having to reconcile the two PRs constantly. 
I can account for potential comments of @adehecq @atedstone & others in a separate PR.

Following @belletva's feedback, I think the two important points before releasing a `0.2` and advertising our point cloud are:
1. Should we immediately remove the Mask classes as proposed in #706?
2. Should LAS support through `laspy` live only in xDEM, as it concerns almost exclusively elevation point clouds?

adehecq:
Thanks for finalizing this long PR ! 🥳 
The 3 points you mentioned above totally make sense. Thanks for the summary!
Regarding your last 2 questions, here's my point of view:


> I'm merging now, as structure changes conflict with #446 and I don't want to duplicate the amount of work by having to reconcile the two PRs constantly. I can account for potential comments of @adehecq @atedstone & others in a separate PR.
> 
> Following @belletva's feedback, I think the two important points before releasing a `0.2` and advertising our point cloud are:
> 
>     1. Should we immediately remove the Mask classes as proposed in [Potential refactoring of `Mask` classes ahead of accessors #706](https://github.com/GlacioHack/geoutils/issues/706)?

See my comment in the associated issue. I find the alternative solution sensible, so green light on my end.
 
>     2. Should LAS support through `laspy` live only in xDEM, as it concerns almost exclusively elevation point clouds?

Looking at the changed files, it looks like it is only added in the dev dependencies for now, is that normal?

I would say the answer mostly depends whether or not laspy adds a lot of inter-dependencies and pinned versions. Looking at the [setup file](https://github.com/laspy/laspy/blob/master/setup.py), it seems to have very loose dependency on numpy and pyproj, so no conflict with us and some optional dependencies for LAZ support. 
We also need to check the [license](https://github.com/laspy/laspy?tab=License-1-ov-file#readme). Although it's not standard, it seem very permissive, so compatible with ours.

Since the dependency does not seem to be an issue, I would say it makes sense to have it in Geoutils directly. LAS is another type of georeferenced data like GTiff or Geopackages. Even though it is mostly used for elevation data, it can actually contain any data (it even includes colored point clouds which are more similar to images). If we decide to have optional dependencies, laspy could of course be made optional.

PR #501, opened 2024-03-07
Make `Raster.to_pointcloud()` modular and add `Raster.get_mask()` for memory-efficient operations requiring mask

### LINKED ISSUE #499: Allow user to pass `z_name` and `random_state` for `Raster.to_points()`
Right now forced to "b1" without any argument.
Could rename the method into `to_pointcloud` for clarity?

### LINKED ISSUE #502: Operations requiring only `Raster.data.mask` load the entire data array
We can use `rasterio.read_masks()` to only get the mask and perform operations requiring only the data mask (and not the data array) in a much more memory-efficient way.
This includes for example:

- `subsample()` and any functionality that uses it like `to_points()`,
- `crop()` with option to crop only to nodata (not yet implemented, see #400).

Starting a `_load_only_mask` function in #501...

### LINKED ISSUE #504: `subsample_array` still runs random subsampling with `subsample=1`
Effectively "shuffles" the dataset, I think it is better to skip the subsampling entirely... Especially for testing purposes.

Ongoing in #501

### LINKED ISSUE #505: Deal consistently with `Raster.data.mask=False`
The mask of a masked array can be just `False` and not an array of `False` when there is no nodata. This can lead to some problems in functions.
The function `geoutils.raster.array.get_mask` circumvents that issue by using `np.isfinite(array)` on the array if the mask is not an array, but that requires the array (which we wouldn't have necessarily loaded, see #502) and mostly more CPU usage.

We should have a consistent way to deal with this, maybe something similar to what is done in `raster_equal` here: https://github.com/GlacioHack/geoutils/blob/d0cd168507a80f7c240fc075c8d856d7dbe6a2a4/geoutils/raster/raster.py#L988
Or a "good practice" from the NumPy documentation?

rhugonnet:
## Summary
This PR improves `Raster.to_pointcloud()` in terms of argument consistency, modularity, and makes its subsampling functionality only returns valid values (while before it returned any random sample) to also be consistent with other subsampling methods.
Additionally, this PR adds a non-public `Raster._load_only_mask()` function called by a new public function `Raster.get_mask()`, which allows to get the nodata mask without loading the array into memory. This is used in `.to_pointcloud()` to perform sampling of only valid values without loading the array in memory :smile:.

The PR also adds a lot of tests!

## Detailed changes

This PR:
- Renames `Raster.to_points()` into `Raster.to_pointcloud()` for consistent naming in preparation for #492,
- Define a main `data_column_name` argument for `Raster.to_pointcloud()` to associate the point cloud to a certain `data_band` (instead of having a hard-coded "b1", "b2", etc), and proposes to store all columns using the new argument `store_auxiliary_bands` (naming to be made consistent with `PointCloud` class but should be fairly close),
- Adds a `random_state` argument for `subsample`,
- Changes the subsampling done to rely on `subsample_array()`, after calling a new `_load_only_mask` function to get the valid mask from disk without loading the raster array,
- Modifies `_load_rio` to allow to return only the mask by calling `rasterio.read_masks()`,
- Adds a `Raster.get_mask()` function that returns the raster mask consistently as a boolean array (to circumvent the issue of having a single `False` value, see #505) and without loading the raster array.

Resolves #499
Resolves #504
Resolves #505
Resolves #502 

TO-DO:

- [x] Add tests for `subsample=1` in `subsample_array()`,
- [x] Add tests for `_load_only_mask`,
- [x] Define a function `get_mask` that works without the array? See #505.

### DIFF
--- a/examples/handling/interface/topoints.py
+++ b/examples/handling/interface/topoints.py
@@ -19,7 +19,7 @@ rast.crop([rast.bounds.left, rast.bounds.bottom, rast.bounds.left + 500, rast.bo
 rast.plot(cmap="terrain")
 
 # %%
-# We convert the raster to points. By default, this returns a vector with columb geometry burned.
+# We convert the raster to points. By default, this returns a vector with column geometry burned.
 
 pts_rast = rast.to_pointcloud()
 pts_rast
@@ -27,4 +27,4 @@ pts_rast
 # %%
 # We plot the point vector.
 
-pts_rast.ds.plot(column="b1", cmap="terrain", legend=True)
+pts_rast.plot(ax="new", column="b1", cmap="terrain", legend=True)
--- a/examples/handling/interface/topoints.py
+++ b/examples/handling/interface/topoints.py
@@ -27,4 +27,4 @@ pts_rast
 # %%
 # We plot the point vector.
 
-pts_rast.plot(column="b1", cmap="terrain", legend=True)
+pts_rast.ds.plot(column="b1", cmap="terrain", legend=True)
--- a/examples/handling/interface/topoints.py
+++ b/examples/handling/interface/topoints.py
@@ -21,7 +21,7 @@ rast.plot(cmap="terrain")
 # %%
 # We convert the raster to points. By default, this returns a vector with columb geometry burned.
 
-pts_rast = rast.to_points()
+pts_rast = rast.to_pointcloud()
 pts_rast
 
 # %%
--- a/geoutils/raster/raster.py
+++ b/geoutils/raster/raster.py
@@ -192,6 +192,7 @@ _default_rio_attrs = [
 
 def _load_rio(
     dataset: rio.io.DatasetReader,
+    only_mask: bool = False,
     indexes: int | list[int] | None = None,
     masked: bool = False,
     transform: Affine | None = None,
@@ -205,6 +206,7 @@ def _load_rio(
     Ensure that ``self.data.ndim=3`` for ease of use (needed e.g. in show).
 
     :param dataset: Dataset to read (opened with :func:`rasterio.open`).
+    :param only_mask: Read only the dataset mask.
     :param indexes: Band(s) to load. Note that rasterio begins counting at 1, not 0.
     :param masked: Whether the mask should be read (if any exists) to use the nodata to mask values.
     :param transform: Create a window from the given transform (to read only parts of the raster)
@@ -242,11 +244,16 @@ def _load_rio(
             window = None
 
     if indexes is None:
-        data = dataset.read(masked=masked, window=window, **kwargs)
+        if only_mask:
+            data = dataset.read_masks(masked=masked, window=window, **kwargs)
+        else:
+            data = dataset.read(masked=masked, window=window, **kwargs)
     else:
-        data = dataset.read(indexes=indexes, masked=masked, window=window, **kwargs)
+        if only_mask:
+            data = dataset.read_masks(indexes=indexes, masked=masked, window=window, **kwargs)
+        else:
+            data = dataset.read(indexes=indexes, masked=masked, window=window, **kwargs)
     return data
-
 
 def _get_reproject_params(
     raster: RasterType,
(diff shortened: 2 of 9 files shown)


examples/handling/interface/topoints.py
```diff
@@ -21,10 +21,10 @@
...
 
-pts_rast = rast.to_points()
+pts_rast = rast.to_pointcloud()
 pts_rast
 
 # %%
 # We plot the point vector.
 
-pts_rast.plot(column="b1", cmap="terrain", legend=True)
+pts_rast.ds.plot(column="b1", cmap="terrain", legend=True)
```
adehecq, line 30: Why do we need to add `ds`? Isn't `pts_rast` a vector instance? In which case, the method `plot` should work directly...

geoutils/raster/raster.py
```diff
@@ -3363,96 +3443,186 @@ def split_bands(self: RasterType, copy: bool = False, bands: list[int] | int | N
...
 
-        If the raster is not loaded, sampling will be done from disk without loading the entire Raster.
+        If the raster is not loaded, sampling will be done from disk using rasterio.sample after loading only the masks
+        of the dataset.
 
         Formats:
             * `as_array` == False: A vector with dataframe columns ["b1", "b2", ..., "geometry"],
             * `as_array` == True: A numpy ndarray of shape (N, 2 + count) with the columns [x, y, b1, b2..].
 
+        :param data_column_name: Name of point cloud data column to use, defaults to "b1".
```
adehecq, line 3532: I would rephrase: "Name to use for point cloud data column" or it sounds like we need to specify the name from some inputs.

geoutils/raster/raster.py
```diff
@@ -3363,96 +3443,186 @@ def split_bands(self: RasterType, copy: bool = False, bands: list[int] | int | N
...
+        random_state: np.random.RandomState | int | None = None,
+        force_pixel_offset: Literal["center", "ul", "ur", "ll", "lr"] = "ul",
+    ) -> NDArrayNum | Vector:
+        ...
+
+    def to_pointcloud(
+        self,
+        data_column_name: str = "b1",
+        data_band: int = 1,
+        store_auxiliary_bands: bool = False,
```
adehecq, line 3500: One question that comes up. For those bands, the name will be "bX" whatever the name chosen for `data_column_name`? Or would it make sense to either use a prefix, or a list of names for multibands?

geoutils/raster/raster.py
```diff
@@ -3363,96 +3443,186 @@ def split_bands(self: RasterType, copy: bool = False, bands: list[int] | int | N
...
         :param as_array: Return an array instead of a vector.
-        :param pixel_offset: The point at which to associate the pixel coordinate with ('corner' == upper left).
+        :param random_state: Random state or seed number.
+        :param force_pixel_offset: Force offset to derive point coordinate with. Raster coordinates normally only
+            associate to upper-left corner "ul" ("Area" definition) or center ("Point" definition).
 
         :raises ValueError: If the sample count or fraction is poorly formatted.
 
-        :returns: A GeoDataFrame, or ndarray of the shape (N, 2 + count) where N is the sample count.
+        :returns: A point cloud, or ndarray of the shape (N, 2 + count) where N is the sample count.
```
adehecq, line 3545: It is not clear what a "point cloud" is.

geoutils/raster/raster.py
```diff
@@ -3363,96 +3443,186 @@ def split_bands(self: RasterType, copy: bool = False, bands: list[int] | int | N
...
-            sample = int(data_size * subsample)
-        else:
-            sample = int(subsample)
-        if sample > data_size:
-            raise ValueError(f"sample cannot exceed the size of the dataset ({sample} vs {data_size})")
+        # Input checks
+        if not isinstance(data_column_name, str):
+            raise ValueError("Data column name must be a string.")
+        if not isinstance(data_band, int) and (1 < data_band or self.count < data_band):
+            raise ValueError(f"Data band number must be between 1 and the total number of bands of {self.count}.")
```
adehecq, line 3552: Rephrase "of {self.count}" into "({self.count})".

geoutils/raster/raster.py
```diff
@@ -3363,96 +3443,186 @@ def split_bands(self: RasterType, copy: bool = False, bands: list[int] | int | N
...
+        ma_valid = np.ma.masked_array(data=np.ones(np.shape(valid_mask), dtype=bool), mask=~valid_mask)
+        # Take a subsample within the valid values
+        indices = subsample_array(array=ma_valid, subsample=subsample, random_state=random_state, return_indices=True)
 
-        # Extract the coordinates of the pixels and filter by the chosen pixels.
-        x_coords, y_coords = (np.array(a) for a in self.ij2xy(rows, cols, offset=pixel_offset))
+        # Extract the coordinates at subsampled pixels with valid data
+        # To extract data, we always use "upper left" which rasterio interprets as the exact coordinates of the raster
+        # Further below we redefine output coordinates based on point interpretation
+        x_coords, y_coords = (np.array(a) for a in self.ij2xy(indices[0], indices[1], offset="ul"))
```
adehecq, line 3581: It seems like `x_coords` and `y_coords` are only used in the second block of the following if statement, so could be moved there no?

tests/test_raster.py
```diff
@@ -2966,51 +2984,122 @@ def test_proximity_parameters(self) -> None:
...
+        assert isinstance(points_arr, np.ndarray)
 
-        assert img0.to_points(0.2, as_array=True).shape == (5, 3)
+        # Check that both outputs (array or vector) are fully consistent, order matters here
+        assert np.array_equal(points.ds.geometry.x.values, points_arr[:, 0])
+        assert np.array_equal(points.ds.geometry.y.values, points_arr[:, 1])
+        assert np.array_equal(points.ds["b1"].values, points_arr[:, 2])
+
+        # Validate that 25 points were sampled (equating to img1.height * img1.width) with x, y, and band0 values.
+        assert isinstance(points_arr, np.ndarray)
```
adehecq, line 3010: this was already tested at L 3002.

tests/test_raster.py
```diff
@@ -2966,51 +2984,122 @@ def test_proximity_parameters(self) -> None:
...
+        assert points_arr.shape == (25, 3)
+        assert points.ds.shape == (25, 2)  # One less column here due to geometry storing X and Y
+        # Check that X, Y and Z arrays are equal to raster array input independently of value order
+        x_coords, y_coords = img0.ij2xy(i=np.arange(0, 5), j=np.arange(0, 5))
```
adehecq, line 3014: can't you just use `Raster.coords`? https://geoutils.readthedocs.io/en/latest/gen_modules/geoutils.Raster.coords.html

adehecq (approved):
Apart some minor comments, looking good to me. Thanks !!

examples/handling/interface/topoints.py
rhugonnet: Indeed, changed back to `plot`!

geoutils/raster/raster.py
rhugonnet: Done

geoutils/raster/raster.py
rhugonnet: You're right! Just took some more time and extended the functionality to support any `auxiliary_data_band` as a list and `auxiliary_column_names` as a list too. Now the user can control everything! :smile:

geoutils/raster/raster.py
rhugonnet: It will soon be with #492

rhugonnet: (and it is specified at the start of the function description)

geoutils/raster/raster.py
rhugonnet: Done

geoutils/raster/raster.py
rhugonnet: Yes, done!

tests/test_raster.py
rhugonnet: Good catch

tests/test_raster.py
rhugonnet: No as it didn't use ij2xy consistently yet, this was corrected in #509. I take a note and I'll see if it makes sense to change when merging conflicts.

geoutils/raster/raster.py
adehecq: In that case, should it not be "PointCloud" instead of "point cloud"?

rhugonnet: All functions descriptions use lower-case to describe objects since 6 months ("array", "raster", "dataframe", etc) to match what is done in other packages (have lighter but still understand descriptipons).
And the typing points exactly to the class so the real object is unmistakable!

Issue #438, opened 2024-01-18
Add default plotting styles


adehecq:
As in seaborn, we could have default plotting styles when using geoutils/xdem.
Here are some ideas:
- create a `type` or `style` attribute and have default plotting parameters for each (e.g. a hillshade should be b/w, a 1-band image b/w etc)
- we could make changes to the rcParams, so that for example that 'image.interpolation' is always set to "None" by default

PR #509, opened 2024-03-13
Add pixel interpretation and casting logic based on `AREA_OR_POINT` and package-level config defaults

### LINKED ISSUE #500: Add `shift_area_or_point` to `ij2xy` to ensure reflectivity of operations under any AREA_OR_POINT interpretation
Right now only `xy2ij` deals with the argument.

### LINKED ISSUE #503: Deal consistently with raster's "AREA_OR_POINT" metadata across all functionalities
I think the best summaries out there on this issue are:
- https://github.com/opengeospatial/ogcapi-coverages/issues/92,
- https://gis.stackexchange.com/questions/122670/is-there-a-standard-for-the-coordinates-of-pixels-in-georeferenced-rasters.

And the standard source defining this is here: https://docs.ogc.org/is/19-008r4/19-008r4.html#_requirements_class_gtrastertypegeokey.

In short: Except if Rasterio is doing some of this through GDAL (I don't think it is, as `RasterIsPoint` or `AREA_OR_POINT` don't appear in their documentation), we need to be careful about it, especially when comparing point data and raster data.

For this, we need:
- To be able to get/set an attribute `Raster.area_or_point` (that would update in the `.tags()` of Rasterio, and carry it across Raster operations if possible),
- Write it to file properly with `save`,
- Assess if it is considered by Rasterio functions through some tests (for instance their `xy()`),
- Implement a consistent behaviour for all functions relating to points or coordinates, the core ones being `xy2ij()`, `ij2xy()`, and `coords()`; which then propagate to `interp_points()`, `to_points()`, etc.

### LINKED ISSUE #508: Merge `_check_cast_raster_array()` and `_overloading_check()` into a consistent implementation
Right now, the two functions are used for similar purposes, but output different things, making things inconsistent.

For instance, calling a NumPy ufunc on two input rasters (let's say `np.arctan2()`), the output nodata cast is forced to the first raster (as coded in `Raster.__array_ufunc__()` which only uses `_check_cast_raster_array()`) while calling a division between two input rasters casts the nodata logically depending on the types by calling `_overloading_check`.

They should be made into a single function with consistent behaviour!

rhugonnet:
Resolves #508
Resolves #500
Resolves #503

## TODO

- [ ] Merge casting functions following #508,
- [ ] Add `shift_area_or_point` transformation to all coordinates function still missing it (`coords` and `ij2xy`),
- [ ] Add warning on `area_or_point` when using a match-reference raster to crop/reproject/proximity another raster?
- [ ] Modify `shift_area_or_point` to be `None` by default and call `config["shift_area_or_point"]`.
- [ ] Add function to convert interpretation? Something like `to_area_or_point()`?

### DIFF
--- a/geoutils/raster/raster.py
+++ b/geoutils/raster/raster.py
@@ -48,7 +48,7 @@ from geoutils.projtools import (
 )
 from geoutils.raster.sampling import subsample_array
 from geoutils.vector import Vector
-from geoutils import config
+from geoutils._config import config
 
 # If python38 or above, Literal is builtin. Otherwise, use typing_extensions
 try:
@@ -355,21 +355,51 @@ def _get_reproject_params(
 
     return dst_transform, dst_size
 
-
-def _check_cast_array_raster(
-    input1: RasterType | NDArrayNum, input2: RasterType | NDArrayNum, operation_name: str
-) -> None:
+def _cast_pixel_interpretation(area_or_point1: Literal["Area", "Point"] | None,
+                               area_or_point2: Literal["Area", "Point"] | None) -> Literal["Area", "Point"] | None:
     """
-    Check the casting between an array and a raster, or raise an (helpful) error message.
+    Cast two pixel interpretations and warn if not castable.
+
+    Casts to:
+     - "Area" if both are "Area",
+     - "Point" if both are "Point",
+     - None if any of the interpretation is None, or
+     - None if one is "Area" and the other "Point" (and raises a warning).
+     """
+
+    # If one is None, cast to None
+    if area_or_point1 is None or area_or_point2 is None:
+        area_or_point_out = None
+    # If both are equal and not None
+    elif area_or_point1 == area_or_point2:
+        area_or_point_out = area_or_point1
+    else:
+        area_or_point_out = None
+        msg = 'One raster has a pixel interpretation "Area" and the other "Point". To silence this warning, ' \
+              'either correct the pixel interpretation of one raster (see ), or deactivate ' \
+              'warnings of pixel interpretation with geoutils.config["warn_area_or_point"]=False.'
+        if config["warn_area_or_point"]:
+            warnings.warn(message=msg, category=UserWarning)
+
+    return area_or_point_out
+
+def _cast_array_raster(
+    input1: RasterType | NDArrayNum, input2: RasterType | NDArrayNum, operation_name: str
+) -> Literal["Area", "Point"] | None:
+    """
+    Cast between an array and a raster, or raise an (helpful) error message. Return pixel interpretation.
 
     :param input1: Raster or array.
     :param input2: Raster or array.
     :param operation_name: Name of operation to raise in the error message.
 
-    :return: None.
+    :return: Pixel interpretation.
     """
 
     if isinstance(input1, Raster) and isinstance(input2, Raster):
+
+        # Cast area or point interpretation
+        area_or_point = _cast_pixel_interpretation(input1.area_or_point, input2.area_or_point)
 
         # Check that both rasters have the same shape and georeferences
         if input1.georeferenced_grid_equal(input2):
@@ -386,8 +416,10 @@ def _check_cast_array_raster(
         # The shape compatibility should be valid even when squeezing
         if isinstance(input1, np.ndarray):
             input1 = input1.squeeze()
+            area_or_point = input2.area_or_point
         elif isinstance(input2, np.ndarray):
             input2 = input2.squeeze()
+            area_or_point = input1.area_or_point
 
         if input1.shape == input2.shape:
             pass
@@ -399,6 +431,8 @@ def _check_cast_array_raster(
                 "than raster2. Or, if the array does not come from a raster, define one with raster = "
                 "Raster.from_array(array, array_transform, array_crs, array_nodata) then reproject."
             )
+
+    return area_or_point
 
 
 class Raster:
@@ -475,6 +509,7 @@ class Raster:
             self.data = filename_or_dataset["data"]
             self.transform: rio.transform.Affine = filename_or_dataset["transform"]
             self.crs: rio.crs.CRS = filename_or_dataset["crs"]
+            self.area_or_point = filename_or_dataset["area_or_point"]
             for key in filename_or_dataset:
                 if key in ["data", "transform", "crs", "nodata", "area_or_point"]:
                     continue
@@ -700,15 +735,21 @@ class Raster:
         """Pixel interpretation of the raster."""
 
         # Check input
-        if new_area_or_point is not None and isinstance(new_area_or_point, str) and new_area_or_point.lower() in ["area", "point"]:
+        if new_area_or_point is not None and not (isinstance(new_area_or_point, str) and new_area_or_point.lower() in ["area", "point"]):
             raise ValueError("New pixel interpretation must be 'Area', 'Point' or None.")
 
-        # Set new input
-        self._area_or_point = new_area_or_point
+        # Set new input as exactly "Area" or "Point"
+        if new_area_or_point is not None:
+            self._area_or_point = new_area_or_point[0].upper() + new_area_or_point[1:].lower()
+        else:
+            self._area_or_point = None
 
         # Update tag only if not None
         if new_area_or_point is not None:
             self.tags.update({"AREA_OR_POINT": new_area_or_point})
+        else:
+            if "AREA_OR_POINT" in self.tags:
+                self.tags.pop("AREA_OR_POINT")
 
     @property
     def driver(self) -> str | None:
@@ -780,6 +821,7 @@ class Raster:
         crs: CRS | int | None,
         nodata: int | float | tuple[int, ...] | tuple[float, ...] | None = None,
         area_or_point: Literal["Area", "Point"] | None = None,
+        tags: dict[str, Any] = None,
     ) -> RasterType:
         """Create a raster from a numpy array and the georeferencing information.
 
@@ -790,6 +832,7 @@ class Raster:
             or an EPSG integer.
         :param nodata: Nodata value.
         :param area_or_point: Pixel interpretation of the raster, will be stored in AREA_OR_POINT metadata.
+        :param tags: Metadata stored in a dictionary.
 
         :returns: Raster created from the provided array and georeferencing.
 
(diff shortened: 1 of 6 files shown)


rhugonnet:
@adehecq @erikmannerfelt The new `config` could also be used to set default matplotlib interpolation, or number of threads used during reprojecting :stuck_out_tongue_winking_eye:
Which we discussed in https://github.com/GlacioHack/xdem/issues/383 and https://github.com/GlacioHack/geoutils/issues/438

geoutils/raster/array.py
```diff
@@ -90,8 +90,8 @@ def get_xy_rotated(raster: gu.Raster, along_track_angle: float) -> tuple[NDArray
 
     myang = np.deg2rad(along_track_angle)
 
-    # Get grid coordinates
-    xx, yy = raster.coords(grid=True)
+    # Get grid coordinates (only relative is important, we don't care about offsets, so let's fix lower-left)
+    xx, yy = raster.coords(grid=True, force_offset="ll")
```
adehecq, line 94: Idf we don't care, why set `force_offset` and not just use the default?

geoutils/raster/raster.py
```diff
@@ -354,23 +355,78 @@ def _get_reproject_params(
...
+    if area_or_point1 is None or area_or_point2 is None:
+        area_or_point_out = None
+    # If both are equal and not None
+    elif area_or_point1 == area_or_point2:
+        area_or_point_out = area_or_point1
+    else:
+        area_or_point_out = None
+        msg = (
+            'One raster has a pixel interpretation "Area" and the other "Point". To silence this warning, '
+            "either correct the pixel interpretation of one raster (see ), or deactivate "
```
adehecq, line 381: is there a missing text in the brackets, after "see ..."?

geoutils/raster/raster.py
```diff
@@ -379,15 +435,18 @@ def _check_cast_array_raster(
...
+    # If other is an array
+    elif isinstance(other, np.ndarray):
 
-        # The shape compatibility should be valid even when squeezing
-        if isinstance(input1, np.ndarray):
-            input1 = input1.squeeze()
-        elif isinstance(input2, np.ndarray):
-            input2 = input2.squeeze()
+        # Squeeze first axis of other data if possible
+        if len(other.shape) == 3 and other.shape[0] == 1:
```
adehecq, line 442: No big deal, but you could use `other.ndim`.

geoutils/raster/raster.py
```diff
@@ -398,6 +457,38 @@ def _check_cast_array_raster(
...
+    # 1/ Output nodata depending on common data type
+    out_dtype = np.promote_types(dtype1, dtype2)
+
+    out_nodata = None
+    # Priority to nodata of first raster if both match the output dtype
+    if (nodata2 is not None) and (out_dtype == dtype2):
+        out_nodata = nodata2
+    elif (nodata1 is not None) and (out_dtype == dtype1):
+        out_nodata = nodata1
+    # For some cases the promoted output type is none if the inputs (e.g. inputs uint8 and int8 = int16 output)
```
adehecq, line 479: There is something missing in this line...

adehecq: The original text below was "For some cases the promote_types is neither (uint8 and int8 = int16),"

geoutils/raster/raster.py
```diff
@@ -3192,23 +3373,81 @@ def xy2ij(
...
-        x, y = rio.transform.xy(self.transform, i, j, offset=offset)
+        # If undefined, default to the global system config
+        if shift_area_or_point is None:
+            shift_area_or_point = config["shift_area_or_point"]
+
+        # Shift by half a pixel back for "Point" interpretation
+        if shift_area_or_point and force_offset is None:
+            if self.area_or_point is not None and self.area_or_point == "Point":
+                i = np.asarray(i) - 0.5
+                j = np.asarray(j) - 0.5
```
adehecq, line 3407: Why is the sign different from the shift applied at line 845?

geoutils/raster/raster.py
```diff
@@ -3272,7 +3512,7 @@ def interp_points(
             transformer = pyproj.Transformer.from_crs(init_crs, dest_crs)
             x, y = transformer.transform(x, y)
 
-        i, j = self.xy2ij(x, y, op=np.float32, shift_area_or_point=shift_area_or_point)
+        i, j = self.xy2ij(x, y, shift_area_or_point=shift_area_or_point)
```
adehecq, line 3515: why has the argument `op` been removed?

tests/test_raster.py
```diff
@@ -19,6 +19,7 @@
 from pylint.lint import Run
 from pylint.reporters.text import TextReporter
 
+import geoutils
```
adehecq, line 22: why don't use the import at the following line?

tests/test_raster.py
```diff
@@ -596,6 +597,81 @@ def test_data_setter(self, dtype: str, nodata_init: str | None) -> None:
...
+        assert "AREA_OR_POINT" in raster_point.tags and raster_point.tags["AREA_OR_POINT"] == "Point"
+
+        # 2.5/ Setter shift and using set_area_or_point() with options
+        old_transform = raster_point.transform
+        # Setting "Area" from "Point" will shift the transform by default
+        raster_point.area_or_point = "Area"
+        assert raster_point.area_or_point == "Area"
+        assert raster_point.transform != old_transform
+
+        # Setting "Pixel" should give back the original transform
```
adehecq, line 629: Replace "Pixel" by "Point"

tests/test_raster.py
```diff
@@ -596,6 +597,81 @@ def test_data_setter(self, dtype: str, nodata_init: str | None) -> None:
...
+        assert raster_point.area_or_point == raster_point_fromarray.area_or_point
+
+        # Copy
+        raster_point_copy = raster_point.copy()
+        assert raster_point.area_or_point == raster_point_copy.area_or_point
+
+        # 4/ With function casting from several Rasters
+
+        # with pytest.warns(UserWarning, match='One raster has a pixel interpretation "Area" and the other "Point".*'):
+        #     r
```
adehecq, line 673: Some missing code here?

tests/test_raster.py
```diff
@@ -681,19 +757,25 @@ def test_downsampling(self, example: str) -> None:
 
             # - Test that xy2ij are consistent with new image -
             # Upper left
-            assert rst_down.xy2ij(rst_down.bounds.left, rst_down.bounds.top) == (0, 0)
+            assert rst_down.xy2ij(rst_down.bounds.left, rst_down.bounds.top, shift_area_or_point=False) == (0, 0)
```
adehecq, line 760: Just out of curiosity, here and all following tests, the argument `shift_area_or_point` is set to False. Is it only for this particular test that it is needed, or would it generally be needed? It would be strange to have to think about this when using xy2ij and hence the default argument should be the intuitive one.

tests/test_raster.py
```diff
@@ -1721,40 +1834,73 @@ def test_intersection(self, example: list[str]) -> None:
...
+        xmin, ymin, xmax, ymax = rst.bounds
 
         # We generate random points within the boundaries of the image
+        np.random.seed(42)
         xrand = np.random.randint(low=0, high=rst.width, size=(10,)) * list(rst.transform)[0] + xmin
         yrand = ymax + np.random.randint(low=0, high=rst.height, size=(10,)) * list(rst.transform)[4]
 
-        # Test reversibility (only works with default upper-left offset)
+        # Test reversibility for any point or area interpretation
+        print(rst.area_or_point)
```
adehecq, line 1903: to remove?

adehecq (approved):
A welcome clarification of the pixel interpretation ! 👏 Thanks !
It's still not fully clear in my head why the signs of the half pixels shifts are sometimes different, but maybe you'll be able to help answer.
I'm also curious how often this pixel interpretation will matter. I rarely encountered a case when I needed to care about it.
Oh and maybe this would require adding a small section about this in the documentation?

geoutils/raster/raster.py
rhugonnet: it's the function default, so we don't need to specify!

tests/test_raster.py
rhugonnet: good point :sweat_smile:, corrected!

tests/test_raster.py
rhugonnet: Done

tests/test_raster.py
rhugonnet: Done

geoutils/raster/array.py
rhugonnet: The tests were written to check in comparison to a rotation of an array matching values with starting 0,0 coordinate (lower-left), which is why I kept it! Don't want to re-write all the tests.
Added a comment to further clarify :smile:

geoutils/raster/raster.py
rhugonnet: You're right, I changed it in the entire package to be more consistent :wink:

geoutils/raster/raster.py
rhugonnet: Will add a link to the documentation here once done. Opened #523

geoutils/raster/raster.py
rhugonnet: Made clearer, might have to change with #517

tests/test_raster.py
rhugonnet: Ah yes, hadn't finished this! I just added the tests for these raised warnings in the `TestArithmetic` and `TestArrayInterface` functions `test_raise_errors`, where they can use the parametrized loops and be more thorough :)

geoutils/raster/raster.py
rhugonnet: That is the last (and quite important) thing that I need to check (last box in the to-do list of the PR). I was wondering about the best way to check this, I could just rely on GDAL's documentation... Or GDAL's printed extents depending on the metadata as shown in https://github.com/GlacioHack/geoutils/issues/503 (first link)

rhugonnet: Right now it's probably wrong

tests/test_raster.py
rhugonnet: It's forced only for this particular test and a couple others, because we made assumptions based on "Area" (upper left) for the values of coordinates checked during the test.
Otherwise we shouldn't have to think about it, so the default is `True`. Most "Point" interpretation are for DEMs anyway, and we co-register them so that usually solves any difference :wink:

geoutils/raster/raster.py
rhugonnet: It was actually right!
The shift in `set_area_or_point` shifts coordinates, so needs the negative sign of the Y, hence the different sign. While in `xy2ij`, it's only an index with (0, 0) starting at the upper left corner.
I added a new suite of test to check this against #503, all fully consistent and passing!

tests/test_raster.py
```diff
@@ -1222,6 +1304,23 @@ def test_crop(self, data: list[str]) -> None:
...
 
+        # - Check related to pixel interpretation -
+
+        # Check warning for a different area_or_point for the match-reference geometry works
+        r.set_area_or_point("Area", shift_area_or_point=False)
+        r2 = r.copy()
+        r2.set_area_or_point("Point", shift_area_or_point=False)
+
+        with pytest.warns(UserWarning, match='One raster has a pixel interpretation "Area" and the other "Point".*'):
+            r.crop(r2)
```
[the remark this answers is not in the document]
rhugonnet: GDAL reports a different extent (as it is based on coordinates) for a "Point" or "Area" raster, even if the georeferencing is the same. See the `gdalinfo` calls in the first link of #503.
So I think we can leave as is!

geoutils/raster/raster.py
rhugonnet: See [4cf8899](https://github.com/GlacioHack/geoutils/pull/509/commits/4cf8899f717c119d0f511c104a48f7260292534a) for changes

rhugonnet:
Thanks for the review! 
I clarified why the half-pixel shift sign differs between the two functions and added a lot of tests in [4cf8899](https://github.com/GlacioHack/geoutils/pull/509/commits/4cf8899f717c119d0f511c104a48f7260292534a). All working as intended! :partying_face: (based on the references in #503) 
And opened issue #523 to add a page in the documentation soon.

And for the use: Basically, you should never have to touch the default parameters, except if you get a warning :smile:

PR #521, opened 2024-03-17
Add `Raster.from_xarray()` to create raster from a `xr.DataArray`

### LINKED ISSUE #519: Raster - support to instantiate a raster directly from an Xarray object
Awesome utility! I am using it in combination with xDEM. 

However, it would be great if there was support for `xr.DataArray` object type in `Raster.from_array()`. I want to be able to directly convert from Xarray to Raster without having to store and read from a file or performing other conversions. 

Currently, I solved this by reading the data and `Affine` transformation using `rio` from my `xr.DataArray`:
```python
dem = xr.DataArray(...)
crs = CRS.from_wkt(dem['spatial_ref'].attrs['crs_wkt'])
epsg = crs.to_epsg()
affine = dem.rio.transform()

raster = gu.Raster.from_array(
    data=dem.data,
    transform=affine,
    crs=epsg
)

x_dem = xdem.DEM(raster)
```

It would be nice to have something like this instead:
```python
dem = xr.DataArray(...)

raster = gu.Raster.from_array(dem)
OR  
raster = gu.Raster.from_xarray(dem)

x_dem = xdem.DEM(raster)
```

### LINKED ISSUE #522: `Raster.to_xarray()` not masking NaN nodata values
It seems like NaNs are not masked when using `Raster.to_xarray()`. The `Raster` object contains the `nodata` attribute which is set via `Raster.to_rio_dataset()`, so not sure what the issue is and if that if that is a common issue, or only happens in my case.

I solved this by manually masking NaN values `slope_xr.where(slope_xr != slope_xr.rio.nodata)`.

```python
%matplotlib inline
import pdemtools
import xdem
from geoutils import Raster
from pyproj import CRS

# download some ArcticDEM data 
bounds = (-2789094.0, 708249.0, -2789044.0, 708299.0)
dem = pdemtools.load.mosaic(
    dataset='arcticdem',
    resolution=2,
    bounds=bounds,
    version='v4.1'
)

# Xarray to Raster
crs = CRS.from_wkt(dem['spatial_ref'].attrs['crs_wkt'])
epsg = crs.to_epsg()
affine = dem.rio.transform()

raster = Raster.from_array(
    data=dem.data,
    crs=epsg,
    transform=affine,
    nodata=-9999.           # set nodata value
)

# Raster to xDEM 
x_dem = xdem.DEM(raster)
print(type(x_dem))
print(x_dem.nodata)

# Compute terrain parameter
slope = xdem.terrain.slope(x_dem)

# Convert back to Xarray
slope_xr = slope.to_xarray()
print(slope_xr.data[0,:3,:3])
```

This code outputs:
geoutils_to_xarray

rhugonnet:
This PR adds `Raster.from_xarray()` to create a raster from an Xarray dataset (this loads the data, in the future will be able to cast functions directly on the Xarray object through #446).

Correct the behaviour of `to_xarray()` which was not converting the data to floating when converting, and not masking nodata values (as default was `False` for reading mask in `open_rasterio()`), leading to unmasked nodata values for integer raster types. Now converts to floating by defaults to mask properly as NaNs (only way of masking with Xarray).

Additionally, this function adds two new arguments to `Raster.raster_equal()`: `strict_masked` and `warn_failure_reason`. The former allows to check for less strict equality regarding the mask (masked cells can have different underlying values, as those are never used anyway), and the latter is to raise an understandable error when `raster_equal()` assertions fail.
We lose information when converting to an Xarray dataset with NaNs (masked values are transformed to NaNs and cannot be retrieved back when re-creating a raster), so we can only check equality with `strict_masked=False`.

Resolves #519
Resolves #522

### DIFF
--- a/geoutils/raster/raster.py
+++ b/geoutils/raster/raster.py
@@ -807,9 +807,9 @@ class Raster:
             driver="GTiff",
         ) as ds:
             if self.count == 1:
-                ds.write(self.data[np.newaxis, :, :])
+                ds.write(self.data.filled(self.nodata)[np.newaxis, :, :])
             else:
-                ds.write(self.data)
+                ds.write(self.data.filled(self.nodata))
 
         # Then open as a DatasetReader
         return mfh.open()
@@ -969,13 +969,17 @@ class Raster:
             self._data[:, ind] = assign  # type: ignore
         return None
 
-    def raster_equal(self, other: RasterType) -> bool:
+    def raster_equal(self, other: RasterType, strict_masked: bool = True, warn_failure_reason: bool = False) -> bool:
         """
         Check if two rasters are equal.
 
         This means that are equal:
         - The raster's masked array's data (including masked values), mask, fill_value and dtype,
         - The raster's transform, crs and nodata values.
+
+        :param other: Other raster.
+        :param strict_masked: Whether to check if masked pixels (in .data.mask) have the same value (in .data.data).
+        :param warn_failure_reason: Whether to warn for the reason of failure if the check does not pass.
         """
 
         # If the mask is just "False", it is equivalent to being equal to an array of False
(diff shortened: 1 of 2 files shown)


geoutils/raster/raster.py
```diff
@@ -1523,6 +1529,7 @@ def set_nodata(
 
         # Update the nodata value
         self._nodata = new_nodata
+        self.data.fill_value = new_nodata
```
rhugonnet, line 1532: This little addition was necessary to make `raster_equal` happy when `set_nodata` is called!

tests/test_raster.py
```diff
@@ -3204,7 +3232,7 @@ def test_reproject(self, mask: gu.Mask) -> None:
             match="Reprojecting a mask with a resampling method other than 'nearest', "
             "the boolean array will be converted to float during interpolation.",
         ):
-            mask.reproject(resampling="bilinear")
+            mask.reproject(res=50, resampling="bilinear", force_source_nodata=2)
```
adehecq, line 3235: why was a change needed here?

adehecq (approved):
Nice !

tests/test_raster.py
rhugonnet: I am not sure, but the wrong `UserWarning` was raised (the "returning self" one because nothing changed for this reprojection). And looking at the test, this makes sense. So the question is why did it pass in the first place...

rhugonnet: The test should be more robust now, I didn't get the courage to investigate more

PR #528, opened 2024-03-27
Refactor `Raster.dtypes` tuple to a `Raster.dtype` string

### LINKED ISSUE #520: Remove iterable `Raster.dtypes` into a single `dtype` attribute
We started with both `nodata` and `dtypes` as tuples to mirror Rasterio, but in the end having this is never possible and making things more complex (need to call `dtypes[0]` every time).

To be consistent with `nodata` that is now a single value, we should change the tuple `dtypes` into a single value `dtype`.

rhugonnet:
Having `dtypes` as a tuple is effectively never used, even in Rasterio if I'm not mistaken, like for `nodata`.

Having a single value makes things much more straightforward for the user (and us).

Resolves #520


adehecq (commented):
Alright !
I've never encountered rasters with multiple bands that had different dtypes... I guess if a multiple dtypes functionality is added in rasterio in the future, we'll just have to reverse this PR 😅

PR #531, opened 2024-03-29
Make `nodata` casting consistent based on new array type

### LINKED ISSUE #516: User-input type checks of `nodata` bypassed in `from_array()` and other dependent functions
The `from_array()` function expects the `nodata` to be checked when calling the class with the dictionary of input values: https://github.com/GlacioHack/geoutils/blob/main/geoutils/raster/raster.py#L788
And rightfully so, as we have a `nodata.setter()` with a good list of checks: https://github.com/GlacioHack/geoutils/blob/main/geoutils/raster/raster.py#L1463

However the attribute is set during class instantiation by-passing the setter: https://github.com/GlacioHack/geoutils/blob/main/geoutils/raster/raster.py#L464
We need to use `set_nodata(nodata, update_array=False, update_mask=False)` during instantiation, or we could have any nodata from user input!

It would probably be good practice to this consistently for other data as well (CRS, transform):
- Either have it in the `.setter()` and not every function that relies on the attribute as input,
- Or have a consistent input check function written outside of the `Raster` class that is called at every occurence.

### LINKED ISSUE #517: Nodata should be cast depending on operation type and not just input
Right now we cast nodata values depending on input types of "raster1" and "raster2" or "array2" or "numeric2" during any operation (through NumPy functions, arithmetic functions).
But, in practice, most operations change the type of the output array, and the derived nodata type might not be compatible with that output. 
This bug was facilitated by the fact that we check the `nodata` compatibility with a raster's data type in `set_nodata()` which is bypassed by `from_array()` in class instantiation, see #516.

We need to either:
1. Read the output array dtype after operation, then cast the nodata logically **but this might be a problem for lazy operations through the Xarray accessor**, or,
2. Use the NumPy table of data type casting to know in advance which output data type to expect. See for example our current tests that use these tables in `ufunc.types` for input: https://github.com/GlacioHack/geoutils/blob/main/tests/test_raster.py#L4077.

rhugonnet:
This PR fixes the fact that `set_nodata` was circumvented during `from_array()` creation, making the behaviour fully consistent for checking nodata input and setting. 
Additionally, for practicality, it adds a `cast_nodata` boolean argument to `from_array()` defaulting to `True` which re-casts the input `nodata` to the default of the array type, if the nodata value was incompatible with the input array type.

The impact of this last change was actually much less of a bigger than I thought, because all integer nodata values are compatible with floating array types and the other direction rarely happens. There is almost no NumPy operations that convert from `float` to `int` (even operations like floordiv keep the output a float even if the output of the operation could technically be integers without loss of information, for instance).

So this mostly addresses the cases where the user is copying the data with a new array that has different data type (fairly common), or provided a non-compatible nodata value directly to from_array (less common).

Resolves #517
Resolves #516

TO-DO:

- [x] Use `set_nodata()` in from_array section of `__init__()`,
- [x] Add optional casting of nodata for new array type in `Raster.copy()`,
- [x] Add automatic casting of nodata for NumPy operations.
- [x] Add tests of new behaviour.

### DIFF
--- a/geoutils/raster/raster.py
+++ b/geoutils/raster/raster.py
@@ -390,14 +390,29 @@ def _cast_pixel_interpretation(
     return area_or_point_out
 
 
+def _cast_nodata(out_dtype: DTypeLike, nodata: int | float | None) -> int | float | None:
+    """
+    Cast nodata value for output data type to default nodata if incompatible.
+
+    :param out_dtype: Dtype of output array.
+    :param nodata: Nodata value.
+
+    :return: Cast nodata value.
+    """
+
+    if out_dtype == bool:
+        nodata = None
+    if nodata is not None and not rio.dtypes.can_cast_dtype(nodata, out_dtype):
+        nodata = _default_nodata(out_dtype)
+    else:
+        nodata = nodata
+
+    return nodata
+
+
 def _cast_numeric_array_raster(
     raster: RasterType, other: RasterType | NDArrayNum | Number, operation_name: str
-) -> tuple[
-    MArrayNum,
-    MArrayNum | NDArrayNum | Number,
-    float | int | tuple[int, ...] | tuple[float, ...] | None,
-    Literal["Area", "Point"] | None,
-]:
+) -> tuple[MArrayNum, MArrayNum | NDArrayNum | Number, float | int | None, Literal["Area", "Point"] | None]:
     """
     Cast a raster and another raster or array or number to arrays with proper metadata, or raise an error message.
 
@@ -1019,9 +1034,10 @@ class Raster:
         data: NDArrayNum | MArrayNum | NDArrayBool,
         transform: tuple[float, ...] | Affine,
         crs: CRS | int | None,
-        nodata: int | float | tuple[int, ...] | tuple[float, ...] | None = None,
+        nodata: int | float | None = None,
         area_or_point: Literal["Area", "Point"] | None = None,
         tags: dict[str, Any] = None,
+        cast_nodata: bool = True,
     ) -> RasterType:
         """Create a raster from a numpy array and the georeferencing information.
 
@@ -1034,6 +1050,8 @@ class Raster:
         :param nodata: Nodata value.
         :param area_or_point: Pixel interpretation of the raster, will be stored in AREA_OR_POINT metadata.
         :param tags: Metadata stored in a dictionary.
+        :param cast_nodata: Automatically cast nodata value to the default nodata for the new array type if not
+            compatible. If False, will raise an error when incompatible.
 
         :returns: Raster created from the provided array and georeferencing.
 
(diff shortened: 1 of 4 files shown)


geoutils/raster/satimg.py
```diff
@@ -428,8 +428,8 @@ def __parse_metadata_from_file(self, fn_meta: str | None) -> None:
 
         return None
 
-    def copy(self, new_array: NDArrayNum | None = None) -> SatelliteImage:
-        new_satimg = super().copy(new_array=new_array)  # type: ignore
+    def copy(self, new_array: NDArrayNum | None = None, cast_nodata: bool = False) -> SatelliteImage:
```
adehecq, line 431: why is the default False, whereas it was True for the Raster class?

tests/test_raster.py
```diff
@@ -1061,14 +1061,32 @@ def test_copy(self, example: str) -> None:
...
 
         # -- Fifth test: check that the new_array argument works when providing a new dtype ##
+        # For an integer dataset cast to float, or opposite (the exploradores dataset will cast from float to int)
         if "int" in r.dtype:
             new_dtype = "float32"
         else:
             new_dtype = "uint8"
-        r2 = r.copy(new_array=r_arr.astype(dtype=new_dtype))
 
+        # This should work all the types by default due to automatic casting
```
adehecq, line 1070: typo "all the types" -> "for all the types"

tests/test_raster.py
```diff
@@ -2927,6 +2945,21 @@ def test_from_array(self, example: str) -> None:
         with pytest.raises(TypeError, match="The transform argument needs to be Affine or tuple."):
             gu.Raster.from_array(data=img.data, transform="lol", crs=None, nodata=None)  # type: ignore
 
+    def test_from_array__nodata_casting(self) -> None:
+        """Check nodata casting of from_array that affects of all other functionalities (copy, etc)"""
+
+        rst = gu.Raster(self.landsat_b4_path)
+        warnings.filterwarnings("ignore", message="New nodata value cells already exist*")
+        rst.set_nodata(255)
```
adehecq, line 2953: or you can use 0 that is not used?

adehecq (approved):
All good ! There's just a block of code that was deleted but I don't understand why.

geoutils/raster/raster.py
```diff
@@ -1038,20 +1064,14 @@ def from_array(
             >>> transform = (30.0, 0.0, 478000.0, 0.0, -30.0, 3108140.0)
             >>> myim = Raster.from_array(data, transform, 32645)
         """
-        if not isinstance(transform, Affine):
```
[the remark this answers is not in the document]
rhugonnet: Because those same checks already exist in the `data.setter` of `transform` and `crs`

geoutils/raster/satimg.py
rhugonnet: good catch! fixed

tests/test_raster.py
rhugonnet: fixed

tests/test_raster.py
rhugonnet: Looks like there are some...

Issue #383, opened 2023-08-08
Add Xarray accessor `gu`


rhugonnet:
Big task! As we discussed, using `rioxarray` and `geocube`, and maybe eventually [geoxarray](https://github.com/geoxarray/geoxarray/)?
Looks like the accessor name `gu` is available, `geo` being taken by `geoxarray`.

See discussions started in: https://github.com/pydata/xarray/issues/8041, https://github.com/corteva/rioxarray/issues/687 and https://github.com/geoxarray/geoxarray/issues/3#issuecomment-1668698026.

PR #537, opened 2024-04-25
Add Dask-delayed raster `subsample()`, `reproject()` and `interp_points()` 

rhugonnet:
## Summary
(Moved from https://github.com/GlacioHack/xdem/pull/508. Some functions were written with help from @ameliefroessl, after discussions that stemmed in https://github.com/GlacioHack/xdem/discussions/501.)

This PR adds three **out-of-memory** raster operations:

1. **Subsampling** (that respects nodata values), 
2. **Georeferenced reprojection** (inspired by https://github.com/opendatacube/odc-geo/pull/88) and
3. **Point interpolation**, which harnesses the fact that rasters are NOT on a rectilinear grid but on a regular grid, to greatly reduce memory usage of `xarray.interp()`.

## Summary of reasoning behind implementations

**For both subsampling and point interpolation**, we had to deal with the difficulty of ragged outputs (varying 1D length, and different than the 2D input chunks). See this blogpost for a nice example of why this can be tricky: https://blog.dask.org/2021/07/02/ragged-output, and the issue it originated from: https://github.com/dask/dask/issues/7589.
The `map_blocks(drop_axis=)` solution proposed in the blogpost unfortunately loads a lot of the chunks in memory at once by dropping the axis (rechunking in 1D essentially). Hence why we inspired ourselves from the `dask.delayed` solution instead.

**For subsampling**: Using `dask.vindex` (or vectorized Xarray indexing) allows to do pointwise sampling, but we cannot know the nodata values in advance. 
In this PR, we compute the number of valid values per chunk in advance, then create a random subsample of the total number of valid values for the array (no matter which chunk they fall in), then identify which chunks they fall in, and sample them out-of-memory per chunk.
The only downside of the method is that the random subsample is chunksize-dependent (for a fixed random state, it will give a different random subsample for different chunk sizes). If the user has multiple arrays to sample from at the same points, the argument `return_indices` can solve this. So this seems fine as long as it is clear in the function description.

**For interpolation**: 
Interpolation in the chunked dimensions was recently added to Xarray: https://github.com/pydata/xarray/pull/4155, but the memory usage is really big https://github.com/pydata/xarray/issues/6799 (we checked this ourselves too, see https://github.com/GlacioHack/xdem/discussions/501#discussioncomment-9171932), because the grid can be rectilinear.
In this PR, we rely on the fact that a raster grid is of equal length in each dimension, which allows to both map the location of points to interpolate using less memory and, for per-chunk interpolation, only need to pass a couple values (step in X/Y and starting X/Y) instead of a full vector of coordinate.
The memory usage is bigger than a single chunk because we transform the input delayed array into overlapping arrays (with a depth of overlap depending on the interpolation method), so overlapping chunks have to be loaded as well. This is fine as long as the chunking is small enough.

**For reprojection**:
The only implementation of this to my knowledge is in https://github.com/opendatacube/odc-geo/pull/88 (and maybe also in `pyresample` and `geowombat`?), but it depends on a long class of classes in `odc-geo` (starting with `GeoBox` and `VariableSizedTiles`). As it didn't seem like so much work, I made a concise stand-alone implementation with the idea that it could potentially be moved to Rioxarray, or be used in GeoUtils's Xarray accessor without adding any new dependency.

## Possible contribution to upstream packages

I will start discussions in Rioxarray/Xarray/Dask to see if it makes sense to integrate some there directly, I was thinking:
- Rioxarray for `reproject()`, but would requires shapely as optional dependency with current implementation; 
- Xarray for `interp()`, on the assumptions that the grid is not rectilinear but equal/regular along the interpolated dimensions, and would have to extend to N-D!
- Dask for `subsample()`: to mirror their `dask.dataframe.sample()` functionality, while `dask.array` doesn't have one.

## Next steps

This is a big step forward for #383. 
We should not describe too much these functions in the API just yet, but wait until the Xarray accessor is out (as it will facilitate greatly reading the chunked array properly through Rioxarray).

## TODO

- [x]  Furnish tests a bit more,
- [x] Split tests into two: memory checks (big dataset) and output accuracy (small datasets),
- [x]  Use https://github.com/itamarst/dask-memusage in `test_delayed` to consistently check out-of-memory operation work as intended,
- [x] Find out why `rio.warp.reproject()` fails with an error during multi-threading only? (Forced to single-thread)
- [x] Make output tests run faster by not relying on the cluster fixture for all tests.


geoutils/raster/delayed.py
```diff
@@ -0,0 +1,746 @@
...
+# Here we harness the fact that rasters are always on regular (or sometimes equal) grids to efficiently map
+# the location of the blocks required for interpolation, which requires little memory usage.
+
+# Code structure inspired by https://blog.dask.org/2021/07/02/ragged-output and the "block_id" in map_blocks
+
+
+def _get_interp_indices_per_block(interp_x, interp_y, starts, num_chunks, chunksize, xres, yres):
+    """Map blocks where each pair of interpolation coordinates will have to be computed."""
+
+    # TODO 1: Check the robustness for chunksize different and X and Y
```
ameliefroessl, line 252: I was testing this function out and unfortunately ran into a `IndexError`. 

 ```shell
IndexError                                Traceback (most recent call last)
Cell In[3], [line 8](vscode-notebook-cell:?execution_count=3&line=8)
      [4](vscode-notebook-cell:?execution_count=3&line=4) xs, ys = np.unravel_index(samples, shape=dask_array.shape)
      [6](vscode-notebook-cell:?execution_count=3&line=6) print(xs,ys)
----> [8](vscode-notebook-cell:?execution_count=3&line=8) output = delayed_interp_points(darr=dask_array, points=(xs, ys), resolution=(0.3,0.3))

File [~/projects/xdem/geoutils/geoutils/raster/delayed.py:336](https://file+.vscode-resource.vscode-cdn.net/home/amelie/projects/xdem/geoutils/examples/~/projects/xdem/geoutils/geoutils/raster/delayed.py:336), in delayed_interp_points(darr, points, resolution, method)
    [333](https://file+.vscode-resource.vscode-cdn.net/home/amelie/projects/xdem/geoutils/examples/~/projects/xdem/geoutils/geoutils/raster/delayed.py:333) num_chunks = expanded.numblocks
    [335](https://file+.vscode-resource.vscode-cdn.net/home/amelie/projects/xdem/geoutils/examples/~/projects/xdem/geoutils/geoutils/raster/delayed.py:335) # Get samples indices per blocks
--> [336](https://file+.vscode-resource.vscode-cdn.net/home/amelie/projects/xdem/geoutils/examples/~/projects/xdem/geoutils/geoutils/raster/delayed.py:336) ind_per_block = _get_interp_indices_per_block(
    [337](https://file+.vscode-resource.vscode-cdn.net/home/amelie/projects/xdem/geoutils/examples/~/projects/xdem/geoutils/geoutils/raster/delayed.py:337)     points[0, :], points[1, :], starts, num_chunks, chunksize, resolution[0], resolution[1]
    [338](https://file+.vscode-resource.vscode-cdn.net/home/amelie/projects/xdem/geoutils/examples/~/projects/xdem/geoutils/geoutils/raster/delayed.py:338) )
    [340](https://file+.vscode-resource.vscode-cdn.net/home/amelie/projects/xdem/geoutils/examples/~/projects/xdem/geoutils/geoutils/raster/delayed.py:340) # Create a delayed object for each block, and flatten the blocks into a 1d shape
    [341](https://file+.vscode-resource.vscode-cdn.net/home/amelie/projects/xdem/geoutils/examples/~/projects/xdem/geoutils/geoutils/raster/delayed.py:341) blocks = expanded.to_delayed().ravel()

File [~/projects/xdem/geoutils/geoutils/raster/delayed.py:265](https://file+.vscode-resource.vscode-cdn.net/home/amelie/projects/xdem/geoutils/examples/~/projects/xdem/geoutils/geoutils/raster/delayed.py:265), in _get_interp_indices_per_block(interp_x, interp_y, starts, num_chunks, chunksize, xres, yres)
    [260](https://file+.vscode-resource.vscode-cdn.net/home/amelie/projects/xdem/geoutils/examples/~/projects/xdem/geoutils/geoutils/raster/delayed.py:260) for i, (x, y) in enumerate(zip(interp_x, interp_y)):
    [261](https://file+.vscode-resource.vscode-cdn.net/home/amelie/projects/xdem/geoutils/examples/~/projects/xdem/geoutils/geoutils/raster/delayed.py:261)     # Because it is a regular grid, we know exactly in which block ID the coordinate will fall
    [262](https://file+.vscode-resource.vscode-cdn.net/home/amelie/projects/xdem/geoutils/examples/~/projects/xdem/geoutils/geoutils/raster/delayed.py:262)     block_i_1d = int((x - starts[0][0]) [/](https://file+.vscode-resource.vscode-cdn.net/) (xres * chunksize[0])) * num_chunks[1] + int(
    [263](https://file+.vscode-resource.vscode-cdn.net/home/amelie/projects/xdem/geoutils/examples/~/projects/xdem/geoutils/geoutils/raster/delayed.py:263)         (y - starts[1][0]) [/](https://file+.vscode-resource.vscode-cdn.net/) (yres * chunksize[1])
    [264](https://file+.vscode-resource.vscode-cdn.net/home/amelie/projects/xdem/geoutils/examples/~/projects/xdem/geoutils/geoutils/raster/delayed.py:264)     )
--> [265](https://file+.vscode-resource.vscode-cdn.net/home/amelie/projects/xdem/geoutils/examples/~/projects/xdem/geoutils/geoutils/raster/delayed.py:265)     ind_per_block[block_i_1d].append(i)
    [267](https://file+.vscode-resource.vscode-cdn.net/home/amelie/projects/xdem/geoutils/examples/~/projects/xdem/geoutils/geoutils/raster/delayed.py:267) return ind_per_block

IndexError: list index out of range
``` 

I first thought it was because of the uneven chunksizes of my array. As far as I understood you are assuming a regular grid here. I think it would be beneficial to make this robust for chunks that are of varying sizes (at least around the edges).  I'm assuming it's a common use case that a raster doesn't perfectly fit the chunksize.  

However! The actual error also appeared when I was testing with a small 4x4 array and the chunksize was 2 (and the chunks are even) ... not sure I completely understand how `block_i_1d` is calculated so I can't say for sure why it's running into the error but to help reproduce, this is my input data: 

```python
data = np.array(
        [
            [1, 2, 3, 4],
            [5, 6, 7, 8],
            [9, 10, 11, 12],
            [13, 14, 15, 16],
        ]
    )
dask_array = da.from_array(data, chunks=(chunk, chunk))

samples = np.random.default_rng().choice(
    dask_array.shape[0] * dask_array.shape[1],
    size=10,
    replace=False,
)
xs, ys = np.unravel_index(samples, shape=dask_array.shape)


output = delayed_interp_points(
    darr=dask_array,
    points=np.array([xs, ys]).reshape((-1, 2)),
    resolution=(0.3, 0.3),
)
```
 
I think using `DaskArray.chunksize` is a bit problematic here because it returns the max of `self.chunks` per dimension. For example: 

```python
dask_array = da.from_array(data, chunks=((1,1,2),(2,2)))
dask_array.chunksize
>>> (2,2)
```

rhugonnet: I think your error might be coming from the fact that `delayed_interp_points` expects coordinates (not indexes like `delayed_subsample`), and your specified resolution is 0.3, so your 4x4 array coordinates are 0 to 1.2 and some of your coordinates have a value of 2 or 3.

To generate random coordinates, you can use the same routine as what is currently done in `test_delayed_interp_points` (which also tests for uneven chunksizes and passes successfully :wink:)! Feel free to add any parameters/inputs to the tests if you noticed I missed something, I explain the logic in the delayed tests at the start of `TestDelayed` and in the description of each `test_delayed_xxx`.

When this function will be wrapped in the Xarray accessor `interp_points` function (like `interp_points` is in the `Raster` class right now), there will be a list of input checks, one of them verifying which input points belongs in the array and returning NaN for those who don't with a warning (and running only the ones within the array extent)! Ideally these checks should not live in the subfunction (depends if we make `delayed_interp_points` public, or just the wrapper function in the Xarray accessor), which is why I didn't add them here :smile:

geoutils/raster/delayed.py
```diff
@@ -0,0 +1,777 @@
...
+    # For destination, we need to create the chunks based on destination chunksizes
+    if dst_chunksizes is None:
+        dst_chunksizes = darr.chunksize
+    dst_chunks = _chunks2d_from_chunksizes_shape(chunksizes=dst_chunksizes, shape=dst_shape)
+    dst_geotiling = ChunkedGeoGrid(grid=dst_geogrid, chunks=dst_chunks)
+
+    # 2/ Get footprints of tiles in CRS of destination array, with a buffer of 2 pixels for destination ones to ensure
+    # overlap, then map indexes of source blocks that intersect a given destination block
+    src_footprints = src_geotiling.get_block_footprints(crs=dst_crs)
+    dst_footprints = dst_geotiling.get_block_footprints().buffer(2 * max(dst_geogrid.res))
```
ameliefroessl, line 719: How is the 2 pixels buffer decided ? Is this enough for reprojections?

ameliefroessl (commented):
Really nice tests!! Looking very good :)

geoutils/raster/delayed.py
rhugonnet: For the Xarray accessor, it will be finalizing this: #446 (the issue in Xarray was just to ask for approval/keep them posted, no code needs to live there), but we'll likely open a small PR simply to add our accessor in their doc to direct interested users here :wink: .

environment.yml
```diff
@@ -12,4 +12,5 @@ dependencies:
   - scipy=1.*
   - tqdm
   - xarray
+  - dask
```
adehecq, line 15: Should we maybe make it an optional dependency? I don't know how much more dependencies dask involves...

geoutils/raster/delayed.py
```diff
@@ -0,0 +1,777 @@
...
+    """Define random state based on varied user input."""
+
+    # Define state for random sampling (to fix results during testing)
+    # Not using the legacy random call is crucial for RAM usage: https://github.com/numpy/numpy/issues/14169
+    if random_state is None:
+        rnd: np.random.RandomState | np.random.Generator = np.random.default_rng()
+    elif isinstance(random_state, np.random.RandomState):
+        rnd = random_state
+    else:
+        rnd = np.random.default_rng(seed=42)
```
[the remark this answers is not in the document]
adehecq: I thought the same

geoutils/raster/delayed.py
```diff
@@ -0,0 +1,777 @@
...
+
+    # If value is between 0 and 1, use a fraction
+    if (subsample <= 1) & (subsample > 0):
+        npoints = int(subsample * total_nb_valids)
+    # Otherwise use the value directly
+    elif subsample > 1:
+        # Use the number of valid points if larger than subsample asked by user
+        npoints = min(int(subsample), total_nb_valids)
+        if subsample > total_nb_valids:
+            warnings.warn(
```
adehecq, line 57: Is it really needed? Or maybe we should have an option to silence it as it could show up frequently...

geoutils/raster/delayed.py
```diff
@@ -0,0 +1,777 @@
...
+    ix, iy = np.unravel_index(np.argwhere(np.isfinite(arr_chunk.flatten()))[subsample_indices], shape=arr_chunk.shape)
+
+    # Convert to full-array indexes by adding the row and column starting indexes for this block
+    ix += block_id["xstart"]
+    iy += block_id["ystart"]
+
+    return np.hstack((ix, iy))
+
+
+def delayed_subsample(
```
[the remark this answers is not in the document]
adehecq: It would be great to compare at least runtime between the non delayed and delayed approaches, for arrays that fit in memory. To check that the method does not add too much overhead for that functionality.

geoutils/raster/delayed.py
```diff
@@ -0,0 +1,777 @@
...
+import numpy as np
+import pandas as pd
+import rasterio as rio
+from scipy.interpolate import interpn
+
+from geoutils._typing import NDArrayNum
+from geoutils.projtools import _get_bounds_projected, _get_footprint_projected
+
+# 1/ SUBSAMPLING
+# Getting an exact subsample size out-of-memory only for valid values is not supported directly by Dask/Xarray
```
adehecq, line 21: Maybe mention "not supported as of 30/04/2024" or "as of version xxx" ? Here and also for the other items below.

geoutils/raster/delayed.py
```diff
@@ -0,0 +1,777 @@
...
+    :return: Array of raster value(s) for the given points.
+    """
+
+    # TODO: Replace by a generic 2D point casting function accepting multiple inputs (living outside this function)
+    # Convert input to 2D array
+    points_arr = np.vstack((points[0], points[1]))
+
+    # Map depth of overlap required for each interpolation method
+    # TODO: Double-check this window somewhere in SciPy's documentation
+    map_depth = {"nearest": 1, "linear": 2, "cubic": 3, "quintic": 5}
```
adehecq, line 335: Ideally, this should be defined outside the function.

geoutils/raster/delayed.py
```diff
@@ -0,0 +1,777 @@
...
+    def bounds(self) -> rio.coords.BoundingBox:
+        return rio.coords.BoundingBox(*rio.transform.array_bounds(self.height, self.width, self.transform))
+
+    def bounds_projected(self, crs: rio.crs.CRS = None) -> rio.coords.BoundingBox:
+        if crs is None:
+            crs = self.crs
+        return _get_bounds_projected(bounds=self.bounds, in_crs=self.crs, out_crs=crs)
+
+    @property
+    def footprint(self) -> gpd.GeoDataFrame:
```
adehecq, line 447: For easier maintenance, this function could simply call `footprint_projected`. Same for `bounds` and `bounds_projected`.

geoutils/raster/delayed.py
```diff
@@ -0,0 +1,777 @@
...
+        dx, b, xmin, d, dy, ymax = list(self.transform)[:6]
+
+        # Convert pixel offsets to georeferenced units
+        if distance_unit == "pixel":
+            # Multiplying the offset by the resolution might amplify floating precision issues
+            # xoff *= self.res[0]
+            # yoff *= self.res[1]
+
+            # Using the boundaries instead!
+            xoff = xoff / self.shape[1] * (self.bounds.right - self.bounds.left)
```
adehecq, line 481: How do you know this will have less floating precision issues? Also, maybe first multiplying by the image range then dividing might cause less issues?

geoutils/raster/delayed.py
```diff
@@ -0,0 +1,777 @@
...
+    # https://github.com/dask/dask/blob/24493f58660cb933855ba7629848881a6e2458c1/dask/array/core.py#L908
+    from dask.utils import cached_cumsum
+
+    starts = [cached_cumsum(c, initial_zero=True) for c in chunks]
+    nb_blocks = num_chunks[0] * num_chunks[1]
+    ixi, iyi = np.unravel_index(np.arange(nb_blocks), shape=(num_chunks[0], num_chunks[1]))
+    block_ids = [
+        {
+            "num_block": i,
+            "ys": starts[0][ixi[i]],
```
adehecq, line 505: took me a bit of time to understand the meaning of "s" and "e". Would it not be more straightforward to call them "ystart" and "yend" (or "ystop")?

geoutils/raster/delayed.py
```diff
@@ -0,0 +1,777 @@
...
+
+        block_ids = self.get_block_locations()
+
+        list_geogrids = []
+        for bid in block_ids:
+            # We get the block size
+            block_shape = (bid["ye"] - bid["ys"], bid["xe"] - bid["xs"])
+            # Build a temporary geogrid with the same transform as the full grid
+            geogrid_tmp = GeoGrid(transform=self.grid.transform, crs=self.grid.crs, shape=block_shape)
+            # And shift it to the right location (X is positive in index direction, Y is negative)
```
adehecq, line 551: Why not directly have GeoGrid take as arguments "xstart" and "ystart"?

geoutils/raster/delayed.py
```diff
@@ -0,0 +1,777 @@
...
+def _delayed_reproject_per_block(
+    *src_arrs: tuple[NDArrayNum], block_ids: list[dict[str, int]], combined_meta: dict[str, Any], **kwargs: Any
+) -> NDArrayNum:
+    """
+    Delayed reprojection per destination block (also rebuilds a square array combined from intersecting source blocks).
+    """
+
+    # If no source chunk intersects, we return a chunk of destination nodata values
+    if len(src_arrs) == 0:
+        dst_arr = np.zeros(combined_meta["dst_shape"], dtype=np.dtype("float64"))
```
adehecq, line 625: why is dtype set to float64 here?

geoutils/raster/delayed.py
```diff
@@ -0,0 +1,777 @@
...
+    if dst_chunksizes is None:
+        dst_chunksizes = darr.chunksize
+    dst_chunks = _chunks2d_from_chunksizes_shape(chunksizes=dst_chunksizes, shape=dst_shape)
+    dst_geotiling = ChunkedGeoGrid(grid=dst_geogrid, chunks=dst_chunks)
+
+    # 2/ Get footprints of tiles in CRS of destination array, with a buffer of 2 pixels for destination ones to ensure
+    # overlap, then map indexes of source blocks that intersect a given destination block
+    src_footprints = src_geotiling.get_block_footprints(crs=dst_crs)
+    dst_footprints = dst_geotiling.get_block_footprints().buffer(2 * max(dst_geogrid.res))
+    dest2source = [np.where(dst.intersects(src_footprints).geometry.values)[0] for dst in dst_footprints]
```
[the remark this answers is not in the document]
adehecq: Just thinking in term of dependency to geopandas, if this was to be moved in another package like rioxarray. It seems like the only dependence is through `_get_footprint_projected` which relies solely on Shapely, and this `intersect` method (if I understand correctly).

geoutils/raster/delayed.py
```diff
@@ -0,0 +1,777 @@
...
+
+def _random_state_from_user_input(
+    random_state: np.random.RandomState | int | None = None,
+) -> np.random.RandomState | np.random.Generator:
+    """Define random state based on varied user input."""
+
+    # Define state for random sampling (to fix results during testing)
+    # Not using the legacy random call is crucial for RAM usage: https://github.com/numpy/numpy/issues/14169
+    if random_state is None:
+        rnd: np.random.RandomState | np.random.Generator = np.random.default_rng()
```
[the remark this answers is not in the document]
rhugonnet: A relic from old code! Now adjusted mirroring #538

geoutils/raster/delayed.py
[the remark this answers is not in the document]
rhugonnet: Code removed, now mirrors what is done in #538 (only relies on `np.random.default_rng()` that already accepts an integer, a generator or None as input)

geoutils/raster/delayed.py
```diff
@@ -0,0 +1,777 @@
...
+        relative_xindex = x - first_xindex_block
+        relative_ind_per_block[k].append(relative_xindex)
+
+    return relative_ind_per_block
+
+
+@dask.delayed  # type: ignore
+def _delayed_nb_valids(arr_chunk: NDArrayNum) -> NDArrayNum:
+    """Count number of valid values per block."""
+    return np.array([np.count_nonzero(np.isfinite(arr_chunk))]).reshape((1, 1))
```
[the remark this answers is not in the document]
rhugonnet: I think we can assume that the nodata are NaNs or masked (both covered by `np.isfinite`). We'll enforce the `masked` option in our accessor to ensure all NaN-related operations are natively supported (while Rioxarray leaves the option to the user).

geoutils/raster/delayed.py
```diff
@@ -0,0 +1,777 @@
...
+        ]
+        subsamples = np.concatenate(dask.compute(*list_subsamples_delayed), axis=0)
+
+        return subsamples
+
+    # To return indices
+    else:
+        # Get robust list of starts (using what is done in block_id of dask.array.map_blocks)
+        # https://github.com/dask/dask/blob/24493f58660cb933855ba7629848881a6e2458c1/dask/array/core.py#L908
+        from dask.utils import cached_cumsum
```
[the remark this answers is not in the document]
rhugonnet: Moved up with other imports

geoutils/raster/delayed.py
```diff
@@ -0,0 +1,777 @@
...
+            "xres": resolution[0],
+            "yres": resolution[1],
+        }
+        for i in range(len(blocks))
+    ]
+
+    # Compute values delayed
+    list_interp = [
+        _delayed_interp_points_block(blocks[i], block_ids[i], points_arr[:, ind_per_block[i]])
+        for i, data_chunk in enumerate(blocks)
```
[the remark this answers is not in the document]
rhugonnet: Good catch!

geoutils/raster/delayed.py
rhugonnet: Yes normally no resampling algorithm uses more than +2/-2, this is the same in ODC. We could do more checking on this if the Rasterio reprojection issues were solved... It'll have to wait.

geoutils/raster/delayed.py
rhugonnet: I think it's critical to have it especially on an out-of-memory dataset where you can't quickly check how many valid values there are. If nothing is raised, the user would wonder why the length of the subsample returned is different.
Added an option to silence.

geoutils/raster/delayed.py
```diff
@@ -0,0 +1,777 @@
...
+) -> list[list[int]]:
+    """
+    Get list of 1D valid subsample indices relative to the block for each block.
+
+    The 1D valid subsample indices correspond to the subsample index to apply for a flattened array of valid values.
+    Relative to the block means converted so that the block indexes for valid values starts at 0 up to the number of
+    valid values in that block (while the input indices go from zero to the total number of valid values in the full
+    array).
+
+    :param xxs: Subsample 1D indexes among a total number of valid values.
```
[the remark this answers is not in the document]
rhugonnet: Renamed to `indices_1d` and adjusted other variable names too

geoutils/raster/delayed.py
rhugonnet: Added

geoutils/raster/delayed.py
rhugonnet: Done

geoutils/raster/delayed.py
rhugonnet: It might not matter at all, this is a relic from when I thought the Rasterio issue came from this code. Changed the comment.

geoutils/raster/delayed.py
rhugonnet: Added a comment to explain, trying to minimize the side of these nested dictionaries that can blow up memory fast for big dask graphs

rhugonnet: It's also described again right below.

geoutils/raster/delayed.py
rhugonnet: I'm not sure how that would work for the transform, a GeoGrid is independent of a chunk size, and "xstart" and "ystart" are indexes not coordinates.
Maybe a class method `GeoGrid.from_corner_xy()` where I pass `bounds.left + xs * xres` and same for y? It's the equivalent of what's done in another order, but if it is clearer it'd be better.

rhugonnet: I think both are equally clear to me, I added more comments to describe the new shape + shift method already coded

geoutils/raster/delayed.py
rhugonnet: Good point, I forgot this bit! Only needs float32 to return NaN, will be cast automatically to float64 if that's the source array type (which is not known here because there are no source arrays intersecting.

geoutils/raster/delayed.py
```diff
@@ -0,0 +1,777 @@
...
+    # Now, we can simply call Rasterio!
+
+    # We build the combined transform from tuple
+    src_transform = rio.transform.Affine(*combined_meta["src_transform"])
+    dst_transform = rio.transform.Affine(*combined_meta["dst_transform"])
+
+    # Reproject
+    dst_arr = np.zeros(combined_meta["dst_shape"], dtype=comb_src_arr.dtype)
+
+    _ = rio.warp.reproject(
```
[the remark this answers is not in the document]
rhugonnet: Done!

environment.yml
rhugonnet: Looks like it's pretty much pandas/numpy, which we already have. Could install only `dask-core` to ensure it's minimal.

geoutils/raster/delayed.py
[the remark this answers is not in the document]
rhugonnet: Fixed

environment.yml
rhugonnet: Keeping `dask` for now.

geoutils/raster/delayed.py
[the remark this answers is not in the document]
rhugonnet: To check runtime we'd need to compare to `Raster.interp_points`, `Raster.reproject` etc, which currently live in the raster class directly and not as a separate function, so it's impractical.
This change in architecture needs to change for #383 anyway, opening two issues.

PR #538, opened 2024-04-27
Remove any use of legacy NumPy random number generator

### LINKED ISSUE #536: Consistently use NumPy's default random number generator to avoid RAM usage issues from legacy call
Turns out using the legacy NumPy random generator
```python
x = np.random.choice()
```
or its equivalent for consistency in the random output:
```python
rnd = np.random.RandomState(np.random.MT19937(np.random.SeedSequence(random_state)))
rnd.choice()
```
and without replacement (`replace=False`) is leaking RAM usage by creating in memory an array the size of the sample size requested, which is not cleared after the function call. So, for instance, asking for 100 random points with a value between 0 and 1 billion, without replacement, will create an array of size 1 billion in memory which stays for a while. See https://github.com/numpy/numpy/issues/14169 and https://github.com/GlacioHack/xdem/discussions/501#discussioncomment-9149146.
And we are currently doing this everywhere :scream: 

We need to replace everywhere in GeoUtils and xDEM by:
```python
rnd = np.random.default_rng(seed=random_state)
```

rhugonnet:
For durability, and was creating some RAM issues for choice without replacement on large arrays. Removed `np.random.RandomState` input type who was just for legacy seeds which we don't need. (more details in https://github.com/GlacioHack/xdem/pull/511)

Resolves #536

### DIFF
--- a/examples/analysis/point_extraction/interpolation.py
+++ b/examples/analysis/point_extraction/interpolation.py
@@ -24,9 +24,9 @@ import geopandas as gpd
 import numpy as np
 
 # Replace by Raster function once done (valid coords)
-np.random.seed(42)
-x_coords = np.random.uniform(rast.bounds.left + 50, rast.bounds.right - 50, 50)
-y_coords = np.random.uniform(rast.bounds.bottom + 50, rast.bounds.top - 50, 50)
+rng = np.random.default_rng(42)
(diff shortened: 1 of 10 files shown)


rhugonnet:
Merging as needed to finalize #537 and https://github.com/GlacioHack/xdem/pull/511

adehecq:
Good to know that the old way (e.g. `np.random.randin`) should be avoided, not just for RAM issues on large arrays!

PR #553, opened 2024-05-31
Add point cloud gridding

### LINKED ISSUE #552: Add point cloud gridding into raster with nodata value setting
Right now we only have raster creation from a point cloud using `from_pointcloud_regular` (see #547), but for the generic case of any point cloud (irregular coordinates), we should have a gridding function based on `scipy.interpolate.griddata`.

rhugonnet:
Resolves #552

### DIFF
--- a/geoutils/pointcloud.py
+++ b/geoutils/pointcloud.py
@@ -1,20 +1,22 @@
 """Module for point cloud manipulation."""
 
-from typing import Literal
 import warnings
+from typing import Literal
 
+import geopandas as gpd
 import numpy as np
 from scipy.interpolate import griddata
-import geopandas as gpd
-
 
 from geoutils._typing import NDArrayNum
 
-def _grid_pointcloud(pc: gpd.GeoDataFrame,
-                     grid_coords: tuple[NDArrayNum],
-                     data_column_name: str = "b1",
-                     resampling: Literal["nearest", "linear", "cubic"] = "linear",
-                     dist_nodata_pixel: float = 1.) -> NDArrayNum:
+
+def _grid_pointcloud(
+    pc: gpd.GeoDataFrame,
+    grid_coords: tuple[NDArrayNum, NDArrayNum],
+    data_column_name: str = "b1",
+    resampling: Literal["nearest", "linear", "cubic"] = "linear",
+    dist_nodata_pixel: float = 1.0,
+) -> NDArrayNum:
     """
     Grid point cloud (possibly irregular coordinates) to raster (regular grid) using delaunay triangles interpolation.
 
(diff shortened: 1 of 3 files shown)


rhugonnet:
Will be wrapped and documented in #492

geoutils/pointcloud.py
```diff
@@ -0,0 +1,77 @@
...
+import numpy as np
+from scipy.interpolate import griddata
+
+from geoutils._typing import NDArrayNum
+
+
+def _grid_pointcloud(
+    pc: gpd.GeoDataFrame,
+    grid_coords: tuple[NDArrayNum, NDArrayNum],
+    data_column_name: str = "b1",
```
adehecq, line 16: Why is the default "b1"? Should it be None instead?

tests/test_pointcloud.py
```diff
@@ -0,0 +1,107 @@
...
+        # 2.1/ Grid points outside the convex hull of all points should always be nodata
+
+        # We convert the full raster to a point cloud, keeping all cells even nodata
+        rst_pc = rst.to_pointcloud(skip_nodata=False).ds
+
+        # We define a multi-point geometry from the individual points, and compute its convex hull
+        poly = geometry.MultiPoint([[p.x, p.y] for p in pc.geometry])
+        chull = poly.convex_hull
+
+        # We compute the index of grid cells interesting the convex hull
```
adehecq, line 49: interesting -> intersecting

adehecq (commented):
Amazing !! Nice functionality to have!

geoutils/pointcloud.py
rhugonnet: It's the same default of "band1" as for the `to_pointcloud()` function and `PointCloud()` class, but it's true it doesn't need a name here (like when creating a `gpd.GeoDataFrame`), so I guess if the default could be "None" and choose the first band by default...
Need to think a bit more on it... We can still adjust this during the `PointCloud` PR + release :wink:

geoutils/pointcloud.py
```diff
@@ -0,0 +1,77 @@
...
+    )
+    with warnings.catch_warnings():
+        warnings.filterwarnings("ignore", category=UserWarning, message="Geometry is in a geographic CRS.*")
+        near = gpd.sjoin_nearest(grid_pc, pc)
+        # In case there are several points at the same distance, it doesn't matter which one is used to compute the
+        # distance, so we keep the first index of closest point
+        index_right = near.groupby(by=near.index)["index_right"].min()
+
+    # Compute distance between points as a function of the pixel sizes in X and Y
+    res_x = np.abs(grid_coords[0][1] - grid_coords[0][0])
```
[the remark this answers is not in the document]
rhugonnet: Added that it is always regular grid in the argument description (is also already mentioned in the first line of docstring).

tests/test_pointcloud.py
rhugonnet: Good catch!

geoutils/pointcloud.py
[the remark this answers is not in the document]
rhugonnet: And adding input check!

Issue #539, opened 2024-04-30
Re-organize raster and vector functionalities as separate functions


rhugonnet:
Already mentioned when discussing #383, we need to move out all function of `Raster` and `Vector` to live outside the class, both to facilitate the Xarray accessor and tests.

Here's an idea of structure, adding new modules:

In _geoutils/_: 

_raster/_
**-- accessor.py** (Xarray raster accessor)
**-- distance.py** (proximity)
**-- georef.py** (for reproject, crop)

_vector/_
**-- geometric.py** (buffer_with_overlap)
**-- georef.py** (reproject, crop)

**raster_vector_interface.py** (rasterize, polygonize, create_mask, to_pointcloud)

adehecq:
I guess all these submodules would never have to be called by the user, as all the functions would be implemented as class methods? If so, this sounds good !

rhugonnet:
Yes exactly!

PR #560, opened 2024-06-07
Define consistent nodata propagation for `interp_points` and allow to return interpolator

### LINKED ISSUE #559: Add option to return interpolator of `interp_points` to optimize speed of some iterative functionalities
Should probably be a non-public functionality, so need to write a `_interp_points` wrapper outside of the `Raster` class.

Very useful for instance to optimize DEM coregistration.

rhugonnet:
Most of the line changes in this PR are due to re-structuring, moving out functions of the `Raster` class (see #539, in preparation for #383) into their own module, here only those linked to `interp_points`.

This PR also adds a nodata propagation scheme (as `scipy.interpolate.interpn` does not support nodata propagation natively except for methods "nearest" and "linear") and allows to return an interpolator that contains this nodata propagation (to optimize speed of iterative methods calling the same interpolator several times, such as DEM coregistration).

It turned out that interpolation of `map_coordinates` with splines of order 3/5 does not match "cubic" or "quintic" in `interpn`, only order 0/1 do match for "nearest" and "linear", so all other are dropped. The nodata propagation and accuracy of results for values near the edge of NaNs is now tested thoroughly! :smile: 

Resolves #559
Starts addressing #539 for functions related to `interp_points`

TO-DO:

- [x] Add tests on `return_interpolator`,
- [x] Add tests validating that the interpolator generated produces exactly the same results as `interpn`,
- [x] Add distance of nodata spreading as an argument,
- [x] Enforce the same NaN propagation between map_coordinates and interpn.


rhugonnet:
This PR is finalized, and ready for review! :smile: @adehecq @erikmannerfelt @atedstone 
**Important:** don't lose time looking at changes from functions moved from `raster.py` to `georeferencing.py`, etc... This is simply restructuring linked to #539, which was necessary to have functionalities separate from the `Raster` class (to avoid always re-building a `Raster` object for performance, for instance in coregistration), and will facilitate building the Xarray accessor as well.
All the core changes are in `interpolate.py` and `test_interpolate.py`.

We now have a fully consistent point interpolation with all benefits and (normally) none of the limitations of existing implementations: 
- Very fast on an equal grid (by relying on `scipy.ndimage.map_coordinates`) which supports "nearest" or "linear" methods,
- Otherwise (for a regular grid, or other interpolation methods), it relies on the equivalent of `scipy.interpolate.interpn` but coded as an interpolator function based on SciPy's code (so that it can be constructed only once and used iteratively, to greatly improve performance for instance in coregistration),
- Supports modular NaN propagation for all methods by adding our own consistent NaN propagation scheme on top of SciPy (that barely supports any NaN propagation in interpolation, only "nearest" and "linear" do in `interpn` and inconsistently with `map_coordinates`), with our default method tested for its high accuracy (propagate NaNs for as many pixels as half the method order rounded up).

All the tests allow to check the robustness of the implementation:
- Exact same results for `interpn` and `map_coordinates`, and the interpolator optionally returned,
- Exact same NaN propagation,
- High accuracy of interpolation near NaN edges (10e-4 relative) for the default NaN propagation.

geoutils/raster/interpolate.py
```diff
@@ -0,0 +1,329 @@
...
+    if dist_nodata_spread == "half_order_up":
+        dist_nodata_spread = int(np.ceil(order / 2))
+    elif dist_nodata_spread == "half_order_down":
+        dist_nodata_spread = int(np.floor(order / 2))
+
+    return dist_nodata_spread
+
+
+def _interpn_interpolator(
+    coords: tuple[NDArrayNum, NDArrayNum],
```
adehecq, line 34: arguments are not described in the docstring

geoutils/raster/interpolate.py
```diff
@@ -0,0 +1,329 @@
...
+    # REMOVED (native NaN support not reliable)
+    # Adding masking of NaNs for methods not supporting it
+    # method_support_nan = method in ["nearest", "linear"]
+
+    # Derive distance to spread nodata to depending on method order
+    order = method_to_order[method]
+    d = _dist_nodata_spread(order=order, dist_nodata_spread=dist_nodata_spread)
+
+    # If NaNs are not supported
+    if True:
```
adehecq, line 63: is the condition here really needed?

geoutils/raster/interpolate.py
```diff
@@ -0,0 +1,329 @@
...
+        # We compute the mask and dilate it to the distance to spread nodatas
+        mask_nan = ~np.isfinite(values)
+        if d != 0:
+            new_mask = binary_dilation(mask_nan, iterations=d).astype("uint8")
```
adehecq, line 67: at this stage, it is assumed that the array is 2D no? However, this is not indicated in the docstring and there is no check that this is actually the case I believe. Until now, I assumed `values` could be a 1D array.

geoutils/raster/interpolate.py
```diff
@@ -0,0 +1,329 @@
...
+        # Zero iterations has a different behaviour in binary_dilation than doing nothing, we want the original array
+        else:
+            new_mask = mask_nan.astype("uint8")
+
+        # We create an interpolator for the mask using nearest
+        interp_mask = RegularGridInterpolator(
+            coords, new_mask, method="nearest", bounds_error=bounds_error, fill_value=1
+        )
+
+        # We replace all NaN values by nearest neighbours to minimize interpolation errors near NaNs
```
adehecq, line 77: Since interpolated values are exactly at a pixel, which neighbor is selected if all 4 closest neighbors are valid? 
Why not do a bilinear interpolation, is it too costly?

geoutils/raster/interpolate.py
```diff
@@ -0,0 +1,329 @@
...
+            results = np.atleast_1d(result.reshape(xi_shape[:-1]))
+            results[invalids.astype(bool)] = np.nan
+
+            return results
+
+        return rectbivariate_interpolator_with_fillvalue
+
+
+def _map_coordinates_nodata_propag(
+    values: NDArrayNum,
```
adehecq, line 144: same comment about arguments not described in docstring

tests/test_raster/test_interpolate.py
```diff
@@ -0,0 +1,568 @@
...
+    def test_dist_nodata_spread(self) -> None:
+        """Test distance of nodata spreading computation based on interpolation order."""
+
+        assert _dist_nodata_spread(0, "half_order_up") == 0
+        assert _dist_nodata_spread(0, "half_order_down") == 0
+        assert _dist_nodata_spread(0, 5) == 5
+
+        assert _dist_nodata_spread(1, "half_order_up") == 1
+        assert _dist_nodata_spread(1, "half_order_down") == 0
+        assert _dist_nodata_spread(5, 5) == 5
```
adehecq, line 39: I assume you meant (1, 5) instead of (5, 5). Same for the following block (3).

tests/test_raster/test_interpolate.py
```diff
@@ -0,0 +1,568 @@
...
+
+        # Check that, outside the edge, the interpolation fails and returns a NaN
+        index_x_edge_rand = [-0.5, -0.5, -0.5, 25, 25, 49.5, 49.5, 49.5]
+        index_y_edge_rand = [-0.5, 25, 49.5, -0.5, 49.5, -0.5, 25, 49.5]
+
+        points_x_rand, points_y_rand = raster.ij2xy(
+            i=index_x_edge_rand, j=index_y_edge_rand, shift_area_or_point=shift_aop
+        )
+
+        # Nearest doesn't apply, just linear and above
```
adehecq, line 198: linear is not in the list, as stated in the comment

adehecq: Also I do not understand why the test is not suited for nearest and linear? Because it is only half a pixel?

tests/test_raster/test_interpolate.py
```diff
@@ -0,0 +1,568 @@
...
+        r = r.crop((r.bounds.left, r.bounds.bottom, r.bounds.left + r.res[0] * 50, r.bounds.bottom + r.res[1] * 50))
+        r.set_area_or_point("Area", shift_area_or_point=False)
+
+        # 1/ Test for an individual point (shape can be tricky in 1 dimension)
+        itest = 10
+        jtest = 10
+        x, y = r.ij2xy(itest, jtest)
+        val = r.interp_points((x, y), method=method, force_scipy_function="map_coordinates")[0]
+        val_img = r.data[itest, jtest]
+        if "nearest" in method or "linear" in method:
```
adehecq, line 238: why exclude the others? Also the line could be `if method in ["nearest", "linear"]:`.

tests/test_raster/test_interpolate.py
```diff
@@ -0,0 +1,568 @@
...
+        # Create a big NaN area in the bottom right (for more complex NaN propagation below)
+        r[80:90, 80:90] = np.nan
+
+        # All surrounding pixels with distance half the method order rounded up should be NaNs
+        order = method_to_order[method]
+        d = _dist_nodata_spread(order=order, dist_nodata_spread=dist)
+
+        # Get indices of NaNs within the distance from NaNs
+        indices_nan = [
+            (i0 + i, j0 + j) for i in np.arange(-d, d + 1) for j in np.arange(-d, d + 1) if (np.abs(i) + np.abs(j)) <= d
```
adehecq, line 308: is the distance calculated using the L1 norm? (because of `np.abs(i) + np.abs(j) <= d`)

tests/test_raster/test_interpolate.py
```diff
@@ -0,0 +1,568 @@
...
+        ]
+        i, j = list(zip(*indices_nan))
+        x, y = r.ij2xy(i, j)
+        vals = r.interp_points((x, y), method=method, force_scipy_function="map_coordinates", dist_nodata_spread=dist)
+        vals2 = r.interp_points((x, y), method=method, force_scipy_function="interpn", dist_nodata_spread=dist)
+
+        assert all(np.isnan(np.atleast_1d(vals))) and all(np.isnan(np.atleast_1d(vals2)))
+
+        # Same check for values within one pixel of the exact coordinates
```
adehecq, line 317: Is this not already included in the previous test?

adehecq (approved):
I have a few comments, but otherwise looking good to me!
Thanks! 🙌

geoutils/raster/interpolate.py
```diff
@@ -0,0 +1,329 @@
...
+from scipy.interpolate import RectBivariateSpline, RegularGridInterpolator
+from scipy.ndimage import binary_dilation, distance_transform_edt, map_coordinates
+
+from geoutils._typing import NDArrayNum, Number
+from geoutils.raster.georeferencing import _coords, _outside_image, _res, _xy2ij
+
+method_to_order = {"nearest": 0, "linear": 1, "cubic": 3, "quintic": 5, "slinear": 1, "pchip": 3, "splinef2d": 3}
+
+
+def _dist_nodata_spread(order: int, dist_nodata_spread: Literal["half_order_up", "half_order_down"] | int) -> int:
```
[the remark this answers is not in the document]
rhugonnet: Very good catch! Fixed

geoutils/raster/interpolate.py
rhugonnet: Removed! Relic from a first iteration

geoutils/raster/interpolate.py
rhugonnet: I leave those checks to SciPy, as this is a non-public subfunction. The public function `Raster.interp_points` always extracts the right shape based on `.coords()` and `get_nanarray`, so there should never be any error there.
To answer the shape: the function works with the same shape for all arrays as described in SciPy. So passing 1D coords tuple and 1D values, or 2D coords tuple and 2D values, and supposedly for ND.
We only test 1D in GeoUtils, and the 2D in xDEM with the new `_reproject_horizontal_xxx` function.

rhugonnet: Correction: values is always 2D, but the point coordinates can have any 1D/2D/ND shape!

geoutils/raster/interpolate.py
rhugonnet: Same as above: pointing towards `scipy.ndimage.map_coordinates` as they are the same arguments 1:1.

geoutils/raster/interpolate.py
rhugonnet: You mean in the case where there is a NaN surrounded by 4 valid pixels? It's true that this filling method would attribute any nearest value then... But the NaN propagation would also yield a NaN everywhere in the surrounding of this pixel, so it wouldn't really create an issue.

The idea in this bit of code here is just to avoid giving a very far-off value (let's say 0 or -9999) compare to the raster value in this area (let's say 8000 if it's a DEM near Everest) during cubic/quintic interpolation, as these interpolations do not support NaNs at all and requires to fill the raster with values everywhere. By feeding the nearest value, we give a value within the same order of magnitude, and so we don't create any effects near the edges.

We're talking about avoiding errors of 10e-2/e-3 relative to the signal.
I'm adding a comment to clarify this in the code!

rhugonnet: And for the default nodata propagation method (half order up), this is also not too much of an issue as this nearest value should be largely out of the interpolation kernel, but it mitigates errors for shorter distance of nodata propagation :)

tests/test_raster/test_interpolate.py
rhugonnet: Good catch, fixed!

tests/test_raster/test_interpolate.py
rhugonnet: Good catch, this was a relic of the previous implementation/tests; now it works for all methods. I have changed to use all :)

tests/test_raster/test_interpolate.py
```diff
@@ -0,0 +1,568 @@
...
+        # With the interpolator (coordinates are re-ordered automatically, as it happens often for rasters)
+        interpolator = _interpn_interpolator(coords=coords, values=values, method=method)
+        vals2 = interpolator((i, j))
+
+        assert np.array_equal(vals, vals2, equal_nan=True)
+
+    @pytest.mark.parametrize("tag_aop", [None, "Area", "Point"])  # type: ignore
+    @pytest.mark.parametrize("shift_aop", [True, False])  # type: ignore
+    def test_interp_points__synthetic(self, tag_aop: str | None, shift_aop: bool) -> None:
+        """Test interp_points function with synthetic data."""
```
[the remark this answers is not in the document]
rhugonnet: Good idea! I've added an outline for all the "interp" tests.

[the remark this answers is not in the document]
rhugonnet: We could open an issue on this, if we wanted to update old tests...

tests/test_raster/test_interpolate.py
rhugonnet: I added a comment to clarify: it is because "cubic" methods and order high orders do not yield the same value as that of the raster grid at its coordinate, only "nearest" and "linear" do.

tests/test_raster/test_interpolate.py
rhugonnet: I added a comment to clarify: previous test checked for points falling right on grid coordinates, this one is for points falling anywhere within half a pixel of the previous grid coordinates.

tests/test_raster/test_interpolate.py
```diff
@@ -0,0 +1,568 @@
...
+            for j in np.arange(-d - 1, d + 2)
+            if (np.abs(i) + np.abs(j)) == d + 1
+        ]
+        i, j = list(zip(*indices_edge))
+        x, y = r.ij2xy(i, j)
+        # And get their interpolated value
+        vals = r.interp_points((x, y), method=method, force_scipy_function="map_coordinates", dist_nodata_spread=dist)
+        vals2 = r.interp_points((x, y), method=method, force_scipy_function="interpn", dist_nodata_spread=dist)
+
+        # Then we fill the NaNs in the raster with a placeholder value of the DEM mean
```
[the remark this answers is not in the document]
rhugonnet: Good catch, fixed

tests/test_raster/test_interpolate.py
rhugonnet: The distance comes the `binary_dilation`... And after some tests it looks like it indeed matches the L1 norm and not the euclidean distance.
We could think of another way to propagate the NaNs than `binary_dilation` (computing the distance for each pixel, like in proximity, and keeping everything smaller than the euclidean distance?).

I'm opening an issue to discuss this in the meantime.

rhugonnet: Opened #603

PR #561, opened 2024-06-08
Run Dask-delayed tests last to avoid asynchronous teardown errors

### LINKED ISSUE #545: Silence asynchronous `PluggyTeardownRaisedWarning` in `test_delayed` due to using `dask-memusage`
The memusage monitoring works but is polluting the test outputs right now, and seems to trigger random errors in subsequent tests if not properly ordered. Might need to force that test to happen last as a temporary fix... :cry:

rhugonnet:
Hopefully a sufficient hack to resolve #545

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2,8 +2,9 @@
 
 from pytest import DoctestItem
 
+
 # To order test modules logically during execution
-def pytest_collection_modifyitems(items):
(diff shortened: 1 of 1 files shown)


rhugonnet:
Passing for now, we'll only know for sure if future PRs don't fail randomly!

PR #578, opened 2024-06-13
Add `.footprint` attribute to `Raster` and `Vector`

### LINKED ISSUE #563: Add `footprint` as an attribute of both `Raster` and `Vector`
Instead of having to write `self.get_footprint_projected(self.crs)`

rhugonnet:
Resolves #563


adehecq (approved):
Great !

PR #579, opened 2024-06-13
Rename `Raster.shift()` into `Raster.translate` and add `inplace` argument to `Vector.translate`

### LINKED ISSUE #569: Add `Vector.shift()` function mirroring `Raster.shift()`

rhugonnet:
Turns out `Vector.translate` already existed (from Geopandas), so renaming Raster method to match instead.

Resolves #569


adehecq (approved):
You will need to update the documentation again ! 😅 
And I will need to update many of my script 😢 
But it's sensible to match geopandas.

rhugonnet:
Yes :sweat_smile:. I checked and there's only two calls in a test function in xDEM thankfully

PR #593, opened 2024-08-02
Fix pip install and add more exhaustive CI checks

### LINKED ISSUE #594: Missing dependencies after pip install
Mirrors https://github.com/GlacioHack/xdem/issues/557, fiona is now only an optional dependency of geopandas, so our code breaks

rhugonnet:
This PR removes `fiona` as a dependency of GeoUtils as it was also dropped by GeoPandas in favor of `pyogrio` (which was breaking the `pip` install, and probably would have eventually broken the `conda` one as well).

This PR also adds a new (fast) CI action that simply checks that the `pip install` works and can import the package (usually runs in less than 2min).

Resolves #594


tests/test_misc.py
```diff
@@ -118,8 +118,8 @@ def useless_func() -> int:
 
     def test_diff_environment_yml(self, capsys) -> None:  # type: ignore
         # Test with synthetic environment
-        env = {"dependencies": ["python==3.9", "numpy", "fiona"]}
-        devenv = {"dependencies": ["python==3.9", "numpy", "fiona", "opencv"]}
+        env = {"dependencies": ["python==3.9", "numpy", "pandas"]}
```
adehecq, line 121: why did you add pandas instead here?

rhugonnet: It is just a synthetic list of dependencies for running the tests (so it would still have worked without this change), but I didn't want to leave "fiona" in there to avoid confusion, as it's not a dependency to our packages anymore, so I simply modified the synthetic list using "pandas" instead.

PR #605, opened 2024-09-05
Add NumPy 2 support

rhugonnet:


rhugonnet:
Update: Opened issue https://github.com/numpy/numpy/issues/27679

rhugonnet:
Confirmation this is a bug from NumPy, we'll have to wait until it is fixed there to update to 2.0. Will likely take a while!

adehecq:
> Confirmation this is a bug from NumPy, we'll have to wait until it is fixed there to update to 2.0. Will likely take a while!

Thanks for looking into this! It's a pity the issue prevents us from upgrading. Some colleagues have had the issue that numpy 2 was installed in there environment and xdem was not working anymore...

rhugonnet:
That's annoying indeed... We pinned the max version of NumPy a bit late, so solving for GeoUtils/xDEM in a NumPy 2.0 env will install older versions, that will indeed fail.

It might take a very long time to get a fix in NumPy, and it's a small aspect, so we could look for a way to circumvent in the meantime.

rhugonnet:
@adehecq @atedstone @adebardo @vschaffn This PR is ready for your review, should be quick :wink:

tests/test_raster/test_raster.py
```diff
@@ -1057,17 +1049,13 @@ def test_is_modified(self, example: str) -> None:
         r = gu.Raster(example)
         assert not r.is_modified
 
-        # This should not trigger the hash
```
adehecq, line 1060: Why are these changes necessary?

tests/test_raster/test_raster.py
```diff
@@ -971,7 +964,6 @@ def test_copy(self, example: str) -> None:
 
         # Open data, modify, and copy
         r = gu.Raster(example)
-        r.data += 5
```
adehecq, line 974: The test does not fit the comment anymore, as the raster is not modified before being copied...

tests/test_raster/test_raster.py
```diff
@@ -1929,20 +1917,14 @@ def test_split_bands(self) -> None:
             red.data.data.squeeze().astype("float32"), img.data.data[0, :, :].astype("float32"), equal_nan=True
         )
 
-        # Modify the red band and make sure it propagates to the original img (it's not a copy)
-        red.data += 1
```
adehecq, line 1933: Why did you delete this test? Did you mean to replace "red.data += 1" with "red.data = red.data + 1"?

adehecq (approved):
Seems all good to me! Apart from the two main points below:
- Maybe we just need to clarify briefly in the code itself why we are casting the nodata type now (when we used to raise an error before) so we don't revert by accident.
- There are just a few tests that seem to have been deleted and I don't understand why.

Thanks a lot! 🙌

tests/test_raster/test_raster.py
rhugonnet: Good point (I made some of these modifs back in September...). This was failing because casting with a Python scalar now changed the output `dtype` of the array.
I added the test back but replaced `0` by `np.array([0], dtype=r.dtype)` and that now works again.

geoutils/raster/raster.py
```diff
@@ -1690,7 +1690,7 @@ def data(self, new_data: NDArrayNum | MArrayNum) -> None:
 
         1. Writes the data in a masked array, whether the input is a classic array or a masked_array,
         2. Reshapes the data to a 2D array if it is single band,
-        3. Raises an error if the dtype is different from that of the Raster, and points towards .copy() or .astype(),
+        3. If new dtype is different from Raster and nodata is not compatible, casts the nodata value to new dtype,
```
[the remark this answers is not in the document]
rhugonnet: It's not strictly needed by 2.0, it's a design choice that was already inconsistent with NumPy (not being able to cast array to a new dtype), and just became more frequent with the new casting rules from Python scalars.

tests/test_raster/test_raster.py
rhugonnet: Added back! Good catch.

tests/test_raster/test_raster.py
rhugonnet: Added back too. Old change from September...

PR #617, opened 2024-10-30
Add tight layout by default in `Raster.plot()`

### LINKED ISSUE #616: Default `Raster.plot()` can have out-of-frame colorbar
This can lead to the colorbar being out-of-frame by default, for instance in the documentation of xDEM (Sphinx-gallery example of Nuth and Kääb, before `0.1`).

Should we default to using `plt.tight_layout()`?

rhugonnet:
Resolves #616

### DIFF
--- a/geoutils/raster/raster.py
+++ b/geoutils/raster/raster.py
@@ -3304,6 +3304,7 @@ class Raster:
             cbar = None
 
         plt.sca(ax0)
+        plt.tight_layout()
 
         # If returning axes
         if return_axes:


adehecq (commented):
Just seeing this. I think it could potentially lead to issues difficult to track later on with people trying to have multiple subplots, no?

rhugonnet:
Yes, I wondered that too. But I haven't had any since then using multiple subplots at many instances (our documentations or for the book chapter).
Sometimes I have to re-call `tight_layout()` for the plot containing subplots to look nice too, but that seems to work.

We can open an issue in the future in we bump into it I guess (or if you have an idea of a case that could fail, test it quickly now)

adehecq:
No, I don't have a problematic case in mind at the moment. So let's wait and see ! :wink:

adehecq:
No, I don't have a problematic case in mind at the moment. So let's wait and see ! :wink:

Issue #595, opened 2024-08-02
Consolidate a function that can open either a raster or vector with error handling


rhugonnet:
Previously used a fiona exception, removed in #593 and https://github.com/GlacioHack/xdem/pull/559 because Fiona is not the default dependency of GeoPandas anymore.

PR #628, opened 2024-11-17
Remove `SatelliteImage` class for reading sensor metadata in favor of a `Raster` option

### LINKED ISSUE #532: Transform `SatelliteImage` class into an optional dictionary attribute in `Raster`
After more thinking, the organization `Raster` / `SatelliteImage` / `DEM` does not make so much sense anymore: some DEMs can be purely modelled, and are not necessarily from an acquisition.

In the end, the `SatelliteImage` class is essentially just an extra metadata parser on top of `Raster`, which could live as simply an `__init__` option of the `Raster` class and store the metadata in an optional dictionary attribute such as the `tags` attribute for raster metadata.

In short, we could:
- Remove the `SatelliteImage` class,
- Add the functionalities as a `parse_sensor_metadata` option to `Raster`, true by default, which stores all in a `.sensor` dictionary attribute,
- Subclass `DEM` from `Raster`.

### LINKED ISSUE #626: Option to turn off name parsing
We had the issue with some students that if we save a DEM with a filename close to the "standard names", the automatic name parsing in xdem.DEM fails because it is not consistent with the expected name. One of the cases that failed was I believe something like "NASADEM_XXX.tif". I can provide a reproducible example if needed, but I don't have the time just now.
I think there should be an option to turn-off name parsing to avoid such conflicts.

rhugonnet:
This PR removes the `SatelliteImage` class (child of `Raster`) and replaces it by a `parse_sensor_metadata` argument (`False` by default) and `silent` argument of `Raster.__init__`, as discussed in #532.

All sensor metadata is saved in `Raster.tags` along other metadata, with practical datatypes when relevant (datetime, float). By necessity, the `tags` metadata is written as `str` on disk, so this PR also adds a function that automatically "decodes'" sensor metadata to convert it back to its datatype.
In short, the sensor metadata can be written/read from file consistently!

Additionally, this PR removes the old option of `Vector.reproject()` accepting a `str` as reference (which is inconsistent with the rest of the package that never accepts strings as method input), as it sometimes triggered errors depending on `fiona`/`pyogrio` reading. 
We could add back this functionality consistently everywhere after solving #595, if desired.

Resolves #626 
Resolves #532

## TO-DO

- [x] Add new tests for `parse_sensor_metadata` behaviour,
- [x] Test if `tags` are conserved after writing to file and re-reading.


adehecq:
Thanks !

PR #630, opened 2024-11-19
Fix buffer without overlap

rhugonnet:
Ongoing


adehecq:
Can you briefly explain what was the issue here?

PR #643, opened 2024-12-17
Change default plot interpolation to avoid large PDF sizes

### LINKED ISSUE #641: Forced `interpolation="None"` in `plot()` to avoid NaN artefacts causes large vectorized file size
And, as we override the default, it's not parametrizable with the `rcParams` of `matplotlib` anymore, the only way is to override everywhere with the `interpolation=` parameter at every call of `plot()`, which is unpractical.

To give an idea of the impact: Plotting a 100MB DEM with our forced `interpolation="None"`: no matter the `figsize` or `dpi` arguments, the PDF figure file is always around 80MB.

rhugonnet:
Partiallly resolves #641

### DIFF
--- a/geoutils/raster/raster.py
+++ b/geoutils/raster/raster.py
@@ -2795,7 +2795,7 @@ class Raster:
 
         # Set matplotlib interpolation to None by default, to avoid spreading gaps in plots
         if "interpolation" not in kwargs.keys():
-            kwargs.update({"interpolation": "None"})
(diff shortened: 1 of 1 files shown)


rhugonnet:
Merging this for the Geomorphology book chapter to be easily re-runnable.

evetion:
Strictly this is not compatible with the current xdem, as it pins 0.1.12, while this is 0.1.13. Found while trying to create a reproducible environment.

PR #644, opened 2025-01-24
Add percentage valid points in `get_stats()`

vschaffn:
Resolves GlacioHack/xdem#679.
Resolves GlacioHack/xdem#685

## Description
- Add `valid points` and `percentage valid points` statistic calculation, as the number of finite values divided by the number of values in the unmasked Raster.
- Add `valid points` and `percentage valid points` in the `dict` returned by `get_stats`, add aliases.
- Add `valid points` and `percentage valid points` in `test_stats` method in `test_raster`.
- Add `valid points no mask` and `percentage valid points no mask` to compute the same stats as above but without masking 
values.
- Add `LE90` in `stats.py` and in `Raster.get_stats`, and test it in `test_stats.py`.

### DIFF
--- a/doc/source/raster_class.md
+++ b/doc/source/raster_class.md
@@ -361,8 +361,9 @@ rast_reproj.to_xarray()
 
 ## Obtain Statistics
 The `get_stats()` method allows to extract key statistical information from a raster in a dictionary.
-Supported statistics are : mean, median, max, mean, sum, sum of squares, 90th percentile, nmad, std, valid points and
-percentage valid points.
+Supported statistics are : mean, median, max, mean, sum, sum of squares, 90th percentile, nmad, rmse, std, valid points
+and percentage valid points.
+The RMSE is only relevant when the raster represents a difference of two objects.
 Callable functions are supported as well.
 
 ### Usage Examples:
--- a/doc/source/raster_class.md
+++ b/doc/source/raster_class.md
@@ -360,10 +360,23 @@ rast_reproj.to_xarray()
 ```
 
 ## Obtain Statistics
-The `get_stats()` method allows to extract key statistical information from a raster in a dictionary.
-Supported statistics are : mean, median, max, mean, sum, sum of squares, 90th percentile, nmad, rmse, std, valid points,
-percentage valid points, valid points all data, percentage valid points all data.
-The RMSE is only relevant when the raster represents a difference of two objects.
+The {func}`~geoutils.Raster.get_stats` method allows to extract key statistical information from a raster in a dictionary.
+Supported statistics are :
+- **Mean:** arithmetic mean of the data, ignoring masked values.
+- **Median:** middle value when the valid data points are sorted in increasing order, ignoring masked values.
+- **Max:** maximum value among the data, ignoring masked values.
+- **Min:** minimum value among the data, ignoring masked values.
+- **Sum:** sum of all data, ignoring masked values.
+- **Sum of squares:** sum of the squares of all data, ignoring masked values.
+- **90th percentile:** point below which 90% of the data falls, ignoring masked values.
+- **NMAD (Normalized Median Absolute Deviation):** robust measure of variability in the data, less sensitive to outliers compared to standard deviation. Ignore masked values.
+- **RMSE (Root Mean Square Error):** commonly used to express the magnitude of errors or variability and can give insight into the spread of the data. Only relevant when the raster represents a difference of two objects. Ignore masked values.
+- **Std (Standard deviation):** Measures the spread or dispersion of the data around the mean, ignoring masked values.
+- **Valid count:** The number of valid (unmasked) data points in the array. It counts the non-masked elements.
+- **Total count:** The total number of finite data points in the array, including masked elements.
+- **Percentage valid points:** The percentage of unmasked data points out of the total number of finite points. Useful for classification statistics.
+- **Size:** Total size of the raster.
+
 Callable functions are supported as well.
 
 ### Usage Examples:
--- a/doc/source/raster_class.md
+++ b/doc/source/raster_class.md
@@ -374,9 +374,12 @@ Supported statistics are :
 - **RMSE (Root Mean Square Error):** commonly used to express the magnitude of errors or variability and can give insight into the spread of the data. Only relevant when the raster represents a difference of two objects. Ignore masked values.
 - **Std (Standard deviation):** Measures the spread or dispersion of the data around the mean, ignoring masked values.
 - **Valid count:** The number of valid (unmasked) data points in the array. It counts the non-masked elements.
-- **Total count:** The total number of finite data points in the array, including masked elements.
-- **Percentage valid points:** The percentage of unmasked data points out of the total number of finite points. Useful for classification statistics.
 - **Size:** Total size of the raster.
+
+If an inlier mask is passed:
+- **Total count:** The number of unmasked data points in the array before applying the inlier mask.
+- **Percentage valid points:** The ration between `valid_count` and `total_count` Useful for classification statistics.
+
 
 Callable functions are supported as well.
 
@@ -401,4 +404,12 @@ rast.get_stats(["mean", "max", "std"])
 def custom_stat(data):
     return np.nansum(data > 100)  # Count the number of pixels above 100
 rast.get_stats(custom_stat)
+```
+
+- Passing an inlier mask:
+```{code-cell} ipython3
+inlier_mask = np.array([[False, False , True],
+                        [False, True , False],
+                        [True, False , True]])
+rast.get_stats(inlier_mask=inlier_mask)
 ```
--- a/doc/source/raster_class.md
+++ b/doc/source/raster_class.md
@@ -361,7 +361,7 @@ rast_reproj.to_xarray()
 
 ## Obtain Statistics
 The `get_stats()` method allows to extract key statistical information from a raster in a dictionary.
-Supported statistics are : mean, median, max, mean, sum, sum of squares, 90th percentile, nmad, std, count and
+Supported statistics are : mean, median, max, mean, sum, sum of squares, 90th percentile, nmad, std, valid points and
 percentage valid points.
 Callable functions are supported as well.
 
--- a/doc/source/raster_class.md
+++ b/doc/source/raster_class.md
@@ -260,14 +260,14 @@ The {func}`~geoutils.Raster.icrop` function accepts only a bounding box in pixel
 By default, {func}`~geoutils.Raster.crop` and {func}`~geoutils.Raster.icrop` return a new Raster unless the inplace parameter is set to True, in which case the cropping operation is performed directly on the original raster object.
 For more details, see the {ref}`specific section and function descriptions in the API<api-geo-handle>`.
 
-### Example for `crop`
+### Example for {func}`~geoutils.Raster.crop`
 ```{code-cell} ipython3
 # Crop raster to smaller bounds
 rast_crop = rast.crop(bbox=(0.3, 0.3, 1, 1))
 print(rast_crop.bounds)
 ```
 
-### Example for `icrop`
+### Example for {func}`~geoutils.Raster.icrop`
 ```{code-cell} ipython3
 # Crop raster using pixel coordinates
 rast_icrop = rast.icrop(bbox=(2, 2, 6, 6))
--- a/doc/source/raster_class.md
+++ b/doc/source/raster_class.md
@@ -361,8 +361,8 @@ rast_reproj.to_xarray()
 
 ## Obtain Statistics
 The `get_stats()` method allows to extract key statistical information from a raster in a dictionary.
-Supported statistics are : mean, median, max, mean, sum, sum of squares, 90th percentile, nmad, rmse, std, valid points
-and percentage valid points.
+Supported statistics are : mean, median, max, mean, sum, sum of squares, 90th percentile, nmad, rmse, std, valid points,
+percentage valid points, valid points all data, percentage valid points all data.
 The RMSE is only relevant when the raster represents a difference of two objects.
 Callable functions are supported as well.
 
--- a/doc/source/raster_class.md
+++ b/doc/source/raster_class.md
@@ -369,6 +369,7 @@ Supported statistics are :
 - **Sum:** sum of all data, ignoring masked values.
 - **Sum of squares:** sum of the squares of all data, ignoring masked values.
 - **90th percentile:** point below which 90% of the data falls, ignoring masked values.
+- **LE90 (Linear Error with 90% confidence):** Difference between the 95th and 5th percentiles of a dataset, representing the range within which 90% of the data points lie. Ignore masked values.
 - **NMAD (Normalized Median Absolute Deviation):** robust measure of variability in the data, less sensitive to outliers compared to standard deviation. Ignore masked values.
 - **RMSE (Root Mean Square Error):** commonly used to express the magnitude of errors or variability and can give insight into the spread of the data. Only relevant when the raster represents a difference of two objects. Ignore masked values.
 - **Std (Standard deviation):** Measures the spread or dispersion of the data around the mean, ignoring masked values.
--- a/doc/source/raster_class.md
+++ b/doc/source/raster_class.md
@@ -361,7 +361,8 @@ rast_reproj.to_xarray()
 
 ## Obtain Statistics
 The `get_stats()` method allows to extract key statistical information from a raster in a dictionary.
-Supported statistics are : mean, median, max, mean, sum, sum of squares, 90th percentile, nmad, rmse, std.
+Supported statistics are : mean, median, max, mean, sum, sum of squares, 90th percentile, nmad, std, count and
+percentage valid points.
 Callable functions are supported as well.
 
 ### Usage Examples:
@@ -377,7 +378,7 @@ rast.get_stats("mean")
 
 - Get multiple statistics in a dict:
 ```{code-cell} ipython3
-rast.get_stats(["mean", "max", "rmse"])
+rast.get_stats(["mean", "max", "std"])
 ```
 
 - Using a custom callable statistic:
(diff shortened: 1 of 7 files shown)


rhugonnet:
Nice addition, and good catch on the redundancy of the RMSE in this case! :smile: 

Two small remarks:
-  For this type of statistic I think "count" is the name often used, for example the statistic argument of this SciPy function: https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.binned_statistic.html, or the name of this NumPy function: https://numpy.org/doc/2.1/reference/generated/numpy.count_nonzero.html. We could use this here too?
- It might be useful to some users to know the total point count (NaNs included)? If we report a "total count" and "valid count", then users can derive the percent of valid points themselves by dividing the two.

Last thought: I didn't see any function to calculate the valid points in the changes, maybe it's not there yet! In that case I would recommend `np.count_nonzero` applied to `np.isfinite`, instead of `~np.isnan`, as the latter considers only NaNs but not +/- infinity that are often unusable in stats (and I've had some misadventures with those being propagated in raster data before!).

adehecq:
Good addition!
Two other thoughts:
- I believe your implementation does not exclude masked values. Remember that `self.data` is a masked array, so invalid pixels are masked instead of being set to NaN. One quick way to get total number of unmasked values is through `self.data.compressed().size`. Maybe it would be good to add 1-2 tests. For example, the "exploradores_aster_dem" example has data gaps.
- the name of your variable is incorrect. It should not be "percentile" but "percentage". 

Regarding @rhugonnet's comment: 
> It might be useful to some users to know the total point count (NaNs included)? If we report a "total count" and "valid count", then users can derive the percent of valid points themselves by dividing the two.

Not so easy to figure out which ones are the most useful between total count, valid count and fraction of valid pixels. It does not make so much sense to give all 3 as one can derive the 3rd from the 2 others... But I think the percentage of valid pixel is more useful than a count (which is generally gonna be large and not very convenient). Also, total count can be easily derived from the data shape... So I would go for either just the percentage of valid pixels, or percentage + total valid count.

vschaffn:
@rhugonnet the function to calculate the valid point was already there, but I used `~np.isnan`, then I have changed for `np.isfinite` following your feedback.

@adehecq the stats are applied on the unmasked data : https://github.com/GlacioHack/geoutils/blob/514ea820dbfc660a0c2f498ebe75b41b0910667a/geoutils/raster/raster.py#L1891
Following your feedback I have changed `percentile` to `percentage`, and I added the `count` 😃

adebardo:
image
This PR is ready for me; I have no inspiration to improve the naming. @rhugonnet , @adehecq , @duboise-cnes , do you have any ideas?  
Feel free to merge this PR during the week to facilitate progress.

rhugonnet:
Actually, this also reminds me that we should use `np.ma` functions everywhere for computational performance: They can be more efficient than doing `np.func(data.compressed())`.

So replacing `np.nanmean(compressed_data)` (note that the `nanmean` was not necessary!) into `np.ma.mean(data)` directly, and same for all other functions :wink:.
And normally our `nmad` should also support this.

geoutils/raster/raster.py
```diff
@@ -1897,69 +1897,55 @@ def _statistics(self, band: int = 1) -> dict[str, np.floating[Any]]:
...
             data = self.data
         else:
             data = self.data[band - 1]
 
-        # If data is a MaskedArray, use the compressed version (without masked values)
-        if isinstance(data, np.ma.MaskedArray):
-            data = data.compressed()
-
         # Compute the statistics
+        compressed_data = data.compressed()
```
duboise-cnes, line 1910: ok i understand that compressed takes only non masked data --> be explicit in naming, commentary. non_masked_data ? or coherent with chosen naming.

geoutils/raster/raster.py
```diff
@@ -1897,69 +1897,55 @@ def _statistics(self, band: int = 1) -> dict[str, np.floating[Any]]:
...
+            "Median": np.nanmedian(compressed_data),
+            "Max": np.nanmax(compressed_data),
+            "Min": np.nanmin(compressed_data),
+            "Sum": np.nansum(compressed_data),
+            "Sum of squares": np.nansum(np.square(compressed_data)),
+            "90th percentile": np.nanpercentile(compressed_data, 90),
+            "NMAD": nmad(compressed_data),
+            "RMSE": np.sqrt(np.nanmean(np.square(compressed_data))),
+            "Standard deviation": np.nanstd(compressed_data),
+            "Valid points": np.count_nonzero(np.isfinite(compressed_data)),
```
duboise-cnes, line 1924: why compute two time and not put valid_points directly here ?

geoutils/raster/raster.py
```diff
@@ -1969,8 +1955,8 @@ def get_stats(
 
         :param stats_name: Name or list of names of the statistics to retrieve. If None, all statistics are returned.
                    Accepted names include:
-                   - "mean", "median", "max", "min", "sum", "sum of squares", "90th percentile", "nmad", "rmse", "std"
-                   You can also use common aliases for these names (e.g., "average", "maximum", "minimum", etc.).
+                   - "mean", "median", "max", "min", "sum", "sum of squares", "90th percentile", "nmad", "rmse", "std",
+                   "valid points", "percentage valid points".
```
duboise-cnes, line 1959: see renaming. proposition: "non-masked valid" points and "non-masked valid points percentage"... why not the all data present ?

geoutils/raster/raster.py
```diff
@@ -1994,16 +1979,17 @@ def get_stats(
...
             "rmse": "RMSE",
+            "rms": "RMSE",
             "std": "Standard deviation",
-            "stddev": "Standard deviation",
-            "standarddev": "Standard deviation",
             "standarddeviation": "Standard deviation",
+            "validpoints": "Valid points",
+            "percentagevalidpoints": "Percentage valid points",
+            "validpointsalldata": "Valid points all data",
+            "percentagevalidpointsalldata": "Percentage valid points all data",
```
duboise-cnes, line 1992: same remarks with renaming explicitely

duboise-cnes:
Great work and remarks @rhugonnet @adebardo @vschaffn 
I have reviewed it see above. 

I would only say that the naming are not that explicit as you asked @adebardo :) :  
- at minima, explain it in documentation precisely (nan, mask, ...) to have at least a RTFM possibility for users ;) 
- and for naming, "valid" is fuzzy: what is valid ? can you describe it to find a naming ?
- also, "all data" what is the meaning. If all data, what is valid ? 

Find something the most explicit with what does the code and really clear for the user. 

My proposition (not perfect...): 
"Number of non-masked (ou unmasked) valid points " "Percentage of non-masked points"
"Number of valid points" and "Percentage of valid points" 
and explain in documentation that valid is all finite values without np.nan, np.inf et -np.inf (if correct ?)
(if compressed is https://numpy.org/doc/stable/reference/generated/numpy.ma.MaskedArray.compressed.html it is explicit ;) )

Hope I understand well and that it helps. Anyway, be explicit in documentation and naming with what is done.

rhugonnet:
I agree with @duboise-cnes. 
My favourite namings would be "Valid count" and "Total count" (see my first comment on top of the thread on the naming "count" in NumPy and SciPy). And in the description/documentation, we can explain that "valid" means non-masked values, which are always finite in a `Raster`.

vschaffn:
@rhugonnet @dubois Thank you for your feedback and for the clarification on how rasters handle masking.

I have updated the code by replacing all instances of `np.nanstat` with `np.ma.stat`, except for the 90th percentile, as `np.ma.percentile` does not exist.

Additionally, I incorporated your suggestions regarding the final statistics (`valid count`, `total count`, and `percentage valid points`) and provided more detailed explanations in the documentation. However, there is a distinction between the `total count` and the `size` of the raster.

For example, in areas like slopes, there is often a 1-pixel `NaN` padding, which we prefer to exclude from the `total count`. Otherwise, the percentage of valid points would not be 100% even when no mask is applied. To address this, I defined `total count` as the number of finite values in the raster and added an additional statistic: `size`, which returns the total size of the raster. These two values may be the same in some cases, but not always.

vschaffn:
@rhugonnet Perhaps I misspoke, when I said no mask, I meant no additional mask (on top of the ‘classic’ mask for a raster which hides nan/inf values), for classification for example.
The purpose of `total_count` is to separate out the masked values, from nan/inf, or from other contributions (such as a classification mask, for example).  It then gives a percentage of valid points which does not take into account nan/inf due to the calculation of slope for example.

Here is an example:

image
image

rhugonnet:
OK I see now! Thanks a lot for the clarification :smile: 

So I think the underlying issue here is that we need to clearly differentiate **nodata masking** (`.mask` and `set_mask()` to set/update valid raster values) and **zonal masking**/**binning** (classification = evaluating the raster on a mask).
Ideally, for evaluating the `Raster` on zonal/value bins and then compute associated statistics (especially counts), it's much better to do `statistic(Raster[Mask])`, otherwise using `set_mask()` makes it impossible to differentiate "invalid values" (NaNs/inf in the data) with "values we don't want to bin". It's also less practical for multiple call (need to re-set the mask every time).

Sorry, I didn't realize this limitation when we framed the design!
It joins my earlier comment for `get_stats()` to support binning (zonal or by other values) in the functionality: https://github.com/GlacioHack/xdem/issues/620#issuecomment-2450873328.

To solve this specific issue here, I think we could simply add an `inlier_masks` argument to `Raster.get_stats()`, which would expect a `list[Mask | np.ndarray[bool]]` and evaluate the statistics on every mask passed from that list (optionally of length 1 for a single mask, and `None` by default which evaluates the whole raster).
This could later be extended to support any kind of binning, not just zonal.

So instead of doing:
```python
rst.set_mask(mask1)
stats1 = rst.get_stats()
rst.set_mask(mask2)
stats2 = rst.get_stats()
```
We'd do:
```python
all_stats = rst.get_stats(inlier_masks=[mask1, mask2])
```

What do you think? :stuck_out_tongue:

vschaffn:
`LE90` have been added (see GlacioHack/xdem#685)
@rhugonnet Thanks for this proposal, but for now I am happy with 
```bash
rst.set_mask(mask1)
stats1 = rst.get_stats()
```
As we plan to develop a high-level classification in xDEM (internal tickets, comming on github soon), we will think about ways to recover statistics from each class easily in classification classes 😄

rhugonnet:
OK. Just to clarify: My proposition was not to add multiple classification classes (this is a bonus), but to solve the issue of ambiguity of the masked values raised above. 

It is impossible to define the `total_count` with the current approach of using `set_mask(classif_mask)` then `get_stats()`. It mixes two masked value concepts: **invalid values** and **classif values**, that once intertwined cannot be recovered.
To do this properly, we would have to pass a `classif_mask` argument directly to `get_stats()` directly, hence my proposition.

geoutils/raster/raster.py
adebardo: https://github.com/GlacioHack/geoutils/pull/644#issuecomment-2665290918

geoutils/raster/raster.py
adebardo: https://github.com/GlacioHack/geoutils/pull/644#issuecomment-2665290918

geoutils/raster/raster.py
adebardo: https://github.com/GlacioHack/geoutils/pull/644#issuecomment-2665290918

geoutils/raster/raster.py
adebardo: https://github.com/GlacioHack/geoutils/pull/644#issuecomment-2665290918

vschaffn:
image
image
image

geoutils/raster/raster.py
```diff
@@ -1890,102 +1890,101 @@ def set_mask(self, mask: NDArrayBool | Mask) -> None:
...
         :param band: The index of the band for which to compute statistics. Default is 1.
-
+        :param total_count: The total number of finite data points in the array.
         :returns: The requested statistic or a dictionary of statistics if multiple or all are requested.
         """
         if not self.is_loaded:
             self.load()
-        stats_dict = self._statistics(band=band)
+        if mask is not None:
+            total_count = np.count_nonzero(~self.get_mask())
```
rhugonnet, line 1977: I think at this line, instead of `~self.get_mask()`, it should be `mask`?
All the values we sample from (so the `total_count`) is based on the `mask`.

rhugonnet (approved):
Great! All good for me after the small changes mentioned above :slightly_smiling_face:

geoutils/raster/raster.py
vschaffn: For the total count, we want to retrieve all the finite values (i.e. all the False values in the outlier mask). As `mask` represents a classification mask for example (inlier mask), it's `self.get_mask()` that we need here 😃

geoutils/raster/raster.py
```diff
@@ -1890,102 +1890,101 @@ def set_mask(self, mask: NDArrayBool | Mask) -> None:
...
         to calculate custom stats.
 
         :param stats_name: Name or list of names of the statistics to retrieve. If None, all statistics are returned.
                    Accepted names include:
-                   - "mean", "median", "max", "min", "sum", "sum of squares", "90th percentile", "nmad", "rmse", "std"
-                   You can also use common aliases for these names (e.g., "average", "maximum", "minimum", etc.).
+                   - "mean", "median", "max", "min", "sum", "sum of squares", "90th percentile", "LE90", "nmad", "rmse",
+                    "std", "valid count", "total count", "percentage valid points", "size".
                    Custom callables can also be provided.
+        :param mask: A mask on which statistics would be applied.
```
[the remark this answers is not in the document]
vschaffn: I agree, I made the modifications

geoutils/raster/raster.py
rhugonnet: Sorry it looks like we have a different definition in mind for `total_count` and/or `valid_count` :sweat_smile: 

Here's how I usually define those, so that we can identify the discrepancy between our definitions :wink::
- The total count represents all the values (valid and invalid) that are sampled from the raster to compute the statistic. When a classif mask is passed, the sampled raster values are limited to the inlier mask (equivalent to running all statistics on the smaller array `Raster[inlier_mask]`). So the `total_count` is `np.count_nonzero(inlier_mask)`. With no inlier mask, on the entire raster, the `total_count` is always all values `np.prod(self.shape)`.
- The valid count represents the valid values among all those sampled to compute the statistic. When a classif mask is passed, these values are limited to the valid values among raster values subsampled by the inlier mask, so `np.count_nonzero(~Raster.mask[inlier_mask])` (which in this case is equivalent the way you compute it with just `np.isfinite(~self.mask)`, because you've used `set_mask()` before). With no inlier mask, on the entire raster, it's always all the valid values `np.count_nonzero(~Raster.mask)`.

rhugonnet: To better show the link between the two with these definition:
Without a classif mask: `total_count` is all values of the masked-array, `valid_count` is all non-masked values of the masked-array.
With a classif mask: Exactly the same calculations as above, but on the masked-array subsampled on the inlier mask `Raster.data[inlier_mask]`, instead of the full masked-array `Raster.data`.

rhugonnet:
More specifically, I was thinking that we should add a `total_inlier_points`, `valid_inlier_points` and `percentage_valid_inlier_points` for when `inlier_mask` is passed (which would correspond to the calculation described here https://github.com/GlacioHack/geoutils/pull/644#discussion_r1969376880 **with** a classif mask).
And we disable those when no classif mask if passed, as you just said!

And then we always keep `total_points`, `valid_points` and `percentage_valid_points` that are the statistics on the full raster, even when an inlier mask is passed (calculations described in https://github.com/GlacioHack/geoutils/pull/644#discussion_r1969376880 **without** a classif mask).

vschaffn:
@rhugonnet : I redefined the follwing stats:
- **Valid points:** number of finite data points in the array.
- **Total points:** total size of the raster.
- **Percentage valid points:** ration between `Valid points` and `Total points`.


If an inlier mask is passed:
- **Valid inlier points:** number of unmasked data points in the array after applying the inlier mask.
- **Percentage valid inlier points:** The ration between **Valid inlier points** and **Valid points**. Useful for classification statistics.

vschaffn:
@rhugonnet :
If an inlier mask is passed:
- **Inlier points:** number of unmasked data points in the inlier mask.
- **Valid inlier points:** number of unmasked data points in the array after applying the inlier mask.
- **Percentage inlier points:** ratio between **Valid inlier points** and **Valid points**. Useful for classification statistics.
- **Percentage valid inlier points:** rati between **Valid inlier points** and **Inlier points**.

doc/source/raster_class.md
```diff
@@ -360,8 +360,31 @@ rast_reproj.to_xarray()
...
+- **RMSE (Root Mean Square Error):** commonly used to express the magnitude of errors or variability and can give insight into the spread of the data. Only relevant when the raster represents a difference of two objects. Ignore masked values.
+- **Std (Standard deviation):** measures the spread or dispersion of the data around the mean, ignoring masked values.
+- **Valid points:** number of finite data points in the array. It counts the non-masked elements.
+- **Total points:** total size of the raster.
+- **Percentage valid points:** ratio between **Valid points** and **Total points**.
+
+
+If an inlier mask is passed:
+- **Inlier points:** number of unmasked data points in the inlier mask.
+- **Valid inlier points:** number of unmasked data points in the array after applying the inlier mask.
```
rhugonnet, line 383: Great!
Small final remarks:
- We could use "Total inlier points" instead of "Inlier points" to keep consistent naming,
- I think the description of "Inlier points" can be simplified to "Number of data points in the inlier mask" (no mention of mask as it includes both masked and unmasked).

And finally, my preference was towards the naming "Count" (used by NumPy, SciPy, etc) rather than "Points" which is ambiguous. I think @duboise-cnes had mentioned this too further above proposing "Number of points" instead, a longer version of "Count".

See you in a moment!

duboise-cnes:
Thanks for the work and convergence ! @rhugonnet @vschaffn and @adebardo !

PR #652, opened 2025-02-13
Implement Tiling for Multiprocessing

### LINKED ISSUE #649: [POC] Tiling for multiprocessing
# Context  

To enable multiprocessing in xDEM for handling large datasets, it is necessary to implement data tiling at the raster level. This is why we are proposing this ticket in GeoUtils.  

# Implementation  

The idea is to create the same API as the `reproject` function at the end of this [raster file](https://github.com/GlacioHack/geoutils/blob/main/geoutils/raster/raster.py):  

```python
tiling_grid, new_shape = raster_sec.compute_tiling(raster_ref)
```
and integrate the code developed in the internally proposed proof of concept (PoC):  

```python
def generate_tiling_grid(
    row_min: float,
    col_min: float,
    row_max: float,
    col_max: float,
    row_split: int,
    col_split: int,
) -> np.ndarray:
    """
    Generate a grid of positions by splitting [row_min, row_max] x
    [col_min, col_max] into segments of size row_split x col_split.
    
    :param row_min: Minimum row index of the bounding box to split
    :param col_min: Minimum column index of the bounding box to split
    :param row_max: Maximum row index of the bounding box to split
    :param col_max: Maximum column index of the bounding box to split
    :param row_split: Height of each split
    :param col_split: Width of each split
    :return: A numpy array grid with splits in two dimensions (0: row, 1: column),
             where each cell contains [row_min, row_max, col_min, col_max].
    """
    nb_col_split = math.ceil((col_max - col_min) / col_split)
    nb_row_split = math.ceil((row_max - row_min) / row_split)
    out_grid = np.ndarray(shape=(nb_row_split, nb_col_split, 4), dtype=float)
    
    for row in range(nb_row_split):
        for col in range(nb_col_split):
            out_grid[row, col, 0] = min(row_max, row_min + row * row_split)
            out_grid[row, col, 1] = min(row_max, row_min + (row + 1) * row_split)
            out_grid[row, col, 2] = min(col_max, col_min + col * col_split)
            out_grid[row, col, 3] = min(col_max, col_min + (col + 1) * col_split)
    
    return out_grid

def compute_tiling(tile_size, dem_ref_path=None, dem_sec_path=None):
    # Get sizes
    shape_ref = None
    shape_sec = None
    shape = None
    
    if dem_ref_path is not None:
        with rio.open(dem_ref_path) as dcr_ref:
            shape_ref = dcr_ref.shape
            shape = shape_ref
            transform_ref = dcr_ref.transform
            
    if dem_sec_path is not None:
        with rio.open(dem_sec_path) as dcr_sec:
            shape_sec = dcr_sec.shape
            shape = shape_sec
    
    if None not in (shape_ref, shape_sec):
        if shape_ref != shape_sec:
            raise Exception("Reference and secondary rasters do not have the same shape")
    elif shape is None:
        raise Exception("Neither reference nor secondary raster provided")
    
    # Generate tiling
    tiling_grid = generate_tiling_grid(0, 0, shape[0], shape[1], tile_size, tile_size)
    
    return tiling_grid, shape
```

## Overlapping Management  

Currently, overlapping is not handled, but it is essential for certain use cases. Since an optimal overlap value has not been determined yet, we will simply add this option for future use.  

## Tests  

- [ ] Implement unit tests

vschaffn:
Resolves #649

## Summary
The tiling grid generation enables splitting large raster datasets into smaller tiles, with an optional overlap parameter, to facilitate multiprocessing and improve performance when handling large

## Features
- Tiling Grid Generation (`generate_tiling_grid`): Generate a grid of positions by splitting [row_min, row_max] x  [col_min, col_max] into tiles of size row_split x col_split with optional overlap.

- Raster Method `compute_tiling`: Compute the raster tiling grid with `generate_tiling_grid`, specifying `tile_size` and `overlap`. Ensure both raster (`self` and `raster_ref`) are compatible with the same tiling.

- Raster Method `plot_tiling` : Plot the raster with its tiling grid.

## Tests
- A test has been implemented on mock and real data, with and without overlapping to ensure the good behavior of `generate_tiling_grid` (expected shape, expected first and last tile, overlap consistency).

## Example (on xDEM)

- 200x200 tiling without overlapping :
image

- 200x200 tiling with 10 pixels overlapping :
image


rhugonnet:
Great! :smile: 

In terms of locations in the package, as these are methods that only array-related, I suggest we move them to either `raster/array.py` or to a new `raster/tiling.py` file (where we could also move the existing `subdivide_array`: https://github.com/GlacioHack/geoutils/blob/6482e77d7cdbca53f2556ebcd63575b85f4ae6d6/geoutils/raster/sampling.py#L152
Preference for the second option.

I don't picture users manipulating these lower-level functions themselves, so we could keep them outside of the `Raster` class as non-public methods that simply work on `Raster.shape`.

Also, I remember that the "stitching" part of the splitting with overlap was quite delicate sometimes (cropping the overlapping chunks with relative indexes). If it can be useful to you, here's how we had it coded in one package ("inv" saves the indexes to stitch the overlapping chunks later on): https://github.com/iamdonovan/pyddem/blob/2fbd54049cf90f77076ad62ba5e1b7cf179cf52e/pyddem/fit_tools.py#L1390

tests/test_raster/test_georeferencing.py
```diff
@@ -209,3 +210,59 @@ def test_coords(self, example: str) -> None:
...
+        tiling_grid_mock = _generate_tiling_grid(0, 0, 100, 100, 50, 50, overlap)
+        if overlap == 0:
+            expected_tiling = np.array([[[0, 50, 0, 50], [0, 50, 50, 100]], [[50, 100, 0, 50], [50, 100, 50, 100]]])
+            assert np.array_equal(tiling_grid_mock, expected_tiling)
+        elif overlap == 5:
+            expected_tiling = np.array(
+                [
+                    [[0, 50, 0, 50], [0, 50, 45, 95], [0, 50, 90, 100]],
+                    [[45, 95, 0, 50], [45, 95, 45, 95], [45, 95, 90, 100]],
+                    [[90, 100, 0, 50], [90, 100, 45, 95], [90, 100, 90, 100]],
```
rhugonnet, line 227: I would have expected the tiling here to yield [0, 55, 0, 55], [0, 55, 45, 100], [45, 100, 0, 55], [45, 100, 45, 100].

The list comprehensions here should normally work, might save you time re-coding this :wink: : https://github.com/iamdonovan/pyddem/blob/main/pyddem/fit_tools.py#L1398

adebardo: Please be careful using pyddem, it's under GPL-3.0 License (same as RichDem), It is therefore not advisable to import it; however, I will look into it regarding "inspiration."

rhugonnet: Yes, I'm not sure where this function came from originally.
From Git blame, I'm the one who added it to Pyddem 5 years ago, but it might have been inspired from a Stackoverflow response or similar.

vschaffn:
@rhugonnet I created a new file `raster/tiling.py` and moved `generate_tiling_grid` and `subdivide_array` as you suggested.
For tiling, if I have understood correctly, the overlap shouldn't be taken into account in the size of the tiles, and should be all around the tiles. That is what I did in the last commit, hope it would be right :smiley:

tests/test_raster/test_georeferencing.py
```diff
@@ -173,7 +173,7 @@ def test_xy2ij(self) -> None:
     @pytest.mark.parametrize("example", [landsat_b4_path, aster_dem_path, landsat_rgb_path])  # type: ignore
     def test_coords(self, example: str) -> None:
 
-        img = gu.Raster(self.landsat_b4_path)
+        img = gu.Raster(example)
```
rhugonnet, line 176: Good catch here!

rhugonnet:
Exactly yes, it's all good now I think! :slightly_smiling_face: 

On the organization again:
As `compute_tiling(Raster.shape)` is mostly meant to be called internally in our future `delayed_multiprocessing` functionalities (and we get the same output by calling `Raster.tiling()` or `tiling(Raster.shape)`), I'm not sure that we need to expose it in the `Raster` class at all, same for `plot_tiling()`. If we do, at least not publicly (Raster._tiling()`).

For next steps: We might have to think about how the user will define the tiling argument into `Raster` for `delayed_multiprocessing` before calling `reproject()`. It might be redundant to have it in every function call, maybe we'll have to define a new attribute `Raster.chunksizes`? (mirroring Dask naming)

vschaffn:
> Exactly yes, it's all good now I think! 🙂
> 
> On the organization again: As `compute_tiling(Raster.shape)` is mostly meant to be called internally in our future `delayed_multiprocessing` functionalities (and we get the same output by calling `Raster.tiling()` or `tiling(Raster.shape)`), I'm not sure that we need to expose it in the `Raster` class at all, same for `plot_tiling()`. If we do, at least not publicly (i.e. `Raster._tiling()`).
> 
> For next steps: We might have to think about how the user will define the tiling argument into `Raster` for `delayed_multiprocessing` before calling `reproject()`. It might be redundant to have it in every function call, maybe we'll have to define a new attribute `Raster.chunksizes`? (mirroring Dask naming)

@rhugonnet I agree that it is maybe better to avoid overloading Raster class with tiling. Then I moved these functions in `tiling.py`.

For next steps, you are probably right, but as tiling is no longer defined in Raster, I find that it makes less sense to define an attribute `Raster.chunksize` there. I think we will see what is best to do about this in future tickets 😃

adebardo:
I am adding the future steps to my thoughts for tiling. Can we merge this PR as it is, Romain?

rhugonnet (approved):
Perfect, yes all good for me!

PR #653, opened 2025-02-13
Temporary fix to CI breaks

rhugonnet:
See #654 
And also: https://github.com/GlacioHack/xdem/issues/639


rhugonnet:
Temporary fix seems to be working, CI passed 5 times in a row.

But now it looks like upload to Coveralls is causing failures...

Issue #648, opened 2025-02-06
[POC] first try of multiprocessing for reprojection 2/2


adebardo:
# Ticket No. 2  
## Context  

The goal of this ticket is to implement tiled reprojection to facilitate the processing of "heavy" datasets, such as 40,000x40,000 DSM CARS.  

## Proposal  

We have observed that this need was anticipated with the creation of the file [[delayed.py](https://github.com/GlacioHack/geoutils/blob/main/geoutils/raster/delayed.py)](https://github.com/GlacioHack/geoutils/blob/main/geoutils/raster/delayed.py). Unfortunately, this file relies on the use of the DASK module, which enables distributed computing. However, based on past experiences, the CS 3D development team does not feel comfortable maintaining such a component for ICC needs, for the following reasons:  

- Memory management issues  
- Significant time lost on debugging and maintenance  
- Dask is effective when initial objects are designed with its philosophy in mind → would require several months of development  

For these reasons, we propose implementing a `delayed_multiproc` module as an alternative to the current delayed implementation.  

## Implementation  (suite)

- [ ] Convert each of the existing functions into a multiprocessing version in `delayed_multiproc.py` (7 functions in total).  
Example:  
```python
## DASK
@dask.delayed  # type: ignore
def _delayed_nb_valids(arr_chunk: NDArrayNum | NDArrayBool) -> NDArrayNum:
    """Count number of valid values per block."""
    if arr_chunk.dtype == "bool":
        return np.array([np.count_nonzero(arr_chunk)]).reshape((1, 1))
    return np.array([np.count_nonzero(np.isfinite(arr_chunk))]).reshape((1, 1))

## MULTIPROCESSING
def _nb_valids(arr_chunk: Union[NDArrayNum, NDArrayBool]) -> NDArrayNum: 
    """Count number of valid values per block.""" 
    if arr_chunk.dtype == "bool": 
        return np.array([np.count_nonzero(arr_chunk)]).reshape((1, 1)) 
    return np.array([np.count_nonzero(np.isfinite(arr_chunk))]).reshape((1, 1)) 

def parallel_nb_valids(chunks: list[Union[NDArrayNum, NDArrayBool]], num_workers: int = 4) -> list[NDArrayNum]: 
    """Parallel execution of _nb_valids using multiprocessing.""" 
    with multiprocessing.Pool(processes=num_workers) as pool: 
        results = pool.map(_nb_valids, chunks) 
    return results
```
  
[...]

PR #655, opened 2025-02-18
Rearrangement of `raster.delayed` to prepare for tiled reprojection implementation with multiprocessing 

### LINKED ISSUE #647: [POC] first try of multiprocessing for reprojection 1/2
# Ticket No. 1  
## Context  

The goal of this ticket is to implement tiled reprojection to facilitate the processing of "heavy" datasets, such as 40,000x40,000 DSM CARS.  

## Proposal  

We have observed that this need was anticipated with the creation of the file [[delayed.py](https://github.com/GlacioHack/geoutils/blob/main/geoutils/raster/delayed.py)](https://github.com/GlacioHack/geoutils/blob/main/geoutils/raster/delayed.py). Unfortunately, this file relies on the use of the DASK module, which enables distributed computing. However, based on past experiences, the CS 3D development team does not feel comfortable maintaining such a component for ICC needs, for the following reasons:  

- Memory management issues  
- Significant time lost on debugging and maintenance  
- Dask is effective when initial objects are designed with its philosophy in mind → would require several months of development  

For these reasons, we propose implementing a `delayed_multiproc` module as an alternative to the current delayed implementation.  

## Implementation  

To achieve this, we propose the following solution:  

- [ ] Research and documentation on Dask / multiprocessing  
- [ ] Update project structure:  
  - In the `raster` directory: create a new `distributed_computing` folder  
  - Inside `raster/calcul_distribue`:  
    - Move and rename `delayed.py` to `delayed_utils.py`  
    - Add `delayed_dask.py`  
    - Add `delayed_multiproc.py`  
- [ ] Move all functions from `delayed.py` that do not depend on Dask to `delayed_utils.py`, as they will be used in both new files  
- [ ] Move all functions dependent on `dask.delayed` to `delayed_dask.py`  

## Tests  

Update test cases involving Dask.  

## Documentation  

Update the documentation accordingly.  

/estimate 5d

vschaffn:
Resolves #647.

## Context
The aim of this ticket is to separate the functions in `raster.delayed` that use dask from those that do not. The functions that do not use `dask` will form a common basis for tiled reprojection with both `dask` and `multiprocessing`, which will be introduced in a future ticket (#648) .

## Changes
- Creation of new repertory `raster.distributed_computing` to organize the new delayed structure of `raster`.
- Creation of new file `raster.distributed_computing.delayed_multiproc` to prepare the next implementations.
- Function from `raster.delayed` that do not use dask have been moved to `raster.distributed_computing.delayed_utils`.
- Function from `raster.delayed` that use dask have been moved to `raster.distributed_computing.delayed_dask`.
- In `_get_block_ids_per_chunk`, `cached_cumsum` has been replaced by `np.cumsum` to avoid using dask, and then keep the function common for both `dask` and `multiprocessing`.

## Tests
- The test structure of  `test_raster.delayed` has been updated in compliance with the new `raster.distributed_computing` directory.

### DIFF
--- a/geoutils/raster/distributed_computing/delayed_multiproc.py
+++ b/geoutils/raster/distributed_computing/delayed_multiproc.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2025 GeoUtils developers
+# Copyright (c) 2025 Centre National d'Etudes Spatiales (CNES)
 #
 # This file is part of the GeoUtils project:
 # https://github.com/glaciohack/geoutils
--- a/geoutils/raster/distributed_computing/delayed_multiproc.py
+++ b/geoutils/raster/distributed_computing/delayed_multiproc.py
@@ -1,0 +1,21 @@
+# Copyright (c) 2025 GeoUtils developers
+#
+# This file is part of the GeoUtils project:
+# https://github.com/glaciohack/geoutils
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+#
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Module for multiprocessing-delayed functions for out-of-memory raster operations.
+"""
--- a/geoutils/raster/distributed_computing/__init__.py
+++ b/geoutils/raster/distributed_computing/__init__.py
@@ -1,0 +1,21 @@
+# Copyright (c) 2025 GeoUtils developers
+#
+# This file is part of the GeoUtils project:
+# https://github.com/glaciohack/geoutils
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+#
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from geoutils.raster.distributed_computing.delayed_dask import *  # noqa
+from geoutils.raster.distributed_computing.delayed_multiproc import *  # noqa
+from geoutils.raster.distributed_computing.delayed_utils import *  # noqa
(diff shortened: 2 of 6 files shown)


rhugonnet:
No comments on this, super! :grin: :rocket: 

**Looking ahead at the next step where we'll have to mirror the main of `delayed_dask`  into `delayed_multiprocessing` (such as  `delayed_reproject()`):** 
I guess it would be ideal to re-use the body of these main functions such as `delayed_reproject()`, as they contain the same over-arching logic in applying the steps sequentially needed for both implementations (and not copy/paste them while adjusting the sub-functions calls?). 
It would simplify maintenance for sure! But I'm unsure how doable that'd be in practice (we could have the same main function for both, one with `backend="dask"` and the other `backend="multiprocessing"` or something: but inputs/outputs may vary...)?

The best is probably to start implementing those main functions in `delayed_multiprocessing`, and once they are operational, if we realize they really mirror closely the ones in `delayed_dask` and could be an easy merge, we can still do that easily later :wink:.

geoutils/raster/distributed_computing/delayed_multiproc.py
```diff
@@ -0,0 +1,21 @@
+# Copyright (c) 2025 GeoUtils developers
```
adebardo, line 1: Please replace with CNES copyright

adebardo (commented):
You will need a __init__.py file in tests/test_raster/test_distributing_computing ---> okay the init file is missing in every folder don't look at my comment ( I can't delete it)

adebardo:
@rhugonnet

Moving these functions could affect the memory usage by altering cache usage or failing to properly clean up memory, but I'm not exactly sure. Do you have any idea?

geoutils/raster/distributed_computing/delayed_multiproc.py
vschaffn: Done

rhugonnet:
@adebardo Not sure I understand, you mean the changes of this PR could negatively affect cache/memory this way?

We have memory testing implemented in `test_delayed`. Those seem to be passing now in this PR's CI.

Sometimes there is an error raised because the cluster does not close properly when using `dem-memusage`. It's an annoying issue we haven't fully solved (for now it is silenced by executing those tests at the end), see here:  https://github.com/GlacioHack/geoutils/issues/545.

rhugonnet:
Haaaaa and that's why the tests are failing in this CI, we need to update the name of the new test module `test_delayed_dask` here to maintain the temporary fix of https://github.com/GlacioHack/geoutils/pull/561: https://github.com/GlacioHack/geoutils/blob/83d9980a28056ddaa071afea9e798f9d0a25dccb/tests/conftest.py#L23

rhugonnet:
@vschaffn @adebardo I went through the code changes again... Hard to grasp why the memory usage would increase. Maybe splitting the code into subfunctions simply creates more links to save in the Dask graph, and this builds up a couple more 10s of MBs over all chunks and loops.

If we check that memory usage before the PR changes was very close to ~100 MBs (the limit of the test, now is slightly above at 120 MBs), then we could simply increase the test threshold by multiplying it by 1.5 or something.
If it was much below 100 MBs, then we should look into it more...

PR #660, opened 2025-03-04
Optimizing `raster.icrop`

### LINKED ISSUE #659: Fix `raster.icrop`
Currently `raster.icrop` uses a bounding box whose inputs are not in the same order as other objects used with `icrop`. Changing the order of the bounding box would be simpler and would avoid errors when using` icrop`.
In addition, `icrop` currently uses crop to cut out the raster. This means that the input bounding box is converted several times (pixel -> coordinates then coordinates-> pixel). There should be a way to make it more optimal.

vschaffn:
Resolves #659 

## Description
- Changes order of `icrop` bounding box to match with Rasterio.BoundingBox.
- Adds a parameter `distance_unit: Literal["georeferenced", "pixel"] = "georeferenced"` in `georeferencing._crop` to match with the use of `icrop`.

## Tests
- Updates crop tests.

## Doc
-  Updates crop section in `raster_class.md`.

### DIFF
--- a/geoutils/raster/georeferencing.py
+++ b/geoutils/raster/georeferencing.py
@@ -173,7 +173,7 @@ def _outside_image(
     """See description of Raster.outside_image."""
 
     if not index:
-        xi, xj = _xy2ij(xi, yj, transform=transform, area_or_point=area_or_point)
+        xi, yj = _xy2ij(xi, yj, transform=transform, area_or_point=area_or_point)
 
     if np.any(np.array((xi, yj)) < 0):
         return True
--- a/geoutils/raster/raster.py
+++ b/geoutils/raster/raster.py
@@ -729,17 +729,16 @@ class Raster:
             and isinstance(new_area_or_point, str)
             and old_area_or_point != new_area_or_point
         ):
-            # The shift below represents +0.5/+0.5 or opposite in indexes (as done in xy2ij), but because
-            # the Y axis is inverted, a minus signs is added to shift the coordinate (even if the unit is in pixel)
+            # The shift below represents +0.5/+0.5 or opposite in indexes (as done in xy2ij)
 
             # If the new one is Point, we shift back by half a pixel
             if new_area_or_point == "Point":
                 xoff = 0.5
-                yoff = -0.5
+                yoff = 0.5
             # Otherwise we shift forward half a pixel
             else:
                 xoff = -0.5
-                yoff = 0.5
+                yoff = -0.5
             # We perform the shift in place
             self.translate(xoff=xoff, yoff=yoff, distance_unit="pixel", inplace=True)
 
--- a/geoutils/raster/raster.py
+++ b/geoutils/raster/raster.py
@@ -219,12 +219,8 @@ def _load_rio(
     * window : to load a cropped version
     * resampling : to set the resampling algorithm
     """
-    # If window is passed, no need to account for transform and shape
-    if kwargs.get("window") is not None:
-        window = kwargs.get("window")
-        kwargs.pop("window")
     # If out_shape is passed, no need to account for transform and shape
-    elif kwargs.get("out_shape") is not None:
+    if kwargs.get("out_shape") is not None:
         window = None
         # If multi-band raster, the out_shape needs to contain the count
         if out_count is not None and out_count > 1:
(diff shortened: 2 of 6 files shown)


adebardo:
Are we sure that the DEM is not completely load before crop ?

geoutils/raster/georeferencing.py
```diff
@@ -173,7 +173,7 @@ def _outside_image(
     """See description of Raster.outside_image."""
 
     if not index:
-        xi, xj = _xy2ij(xi, yj, transform=transform, area_or_point=area_or_point)
+        xi, yj = _xy2ij(xi, yj, transform=transform, area_or_point=area_or_point)
```
rhugonnet, line 176: Good catch! :open_mouth:

geoutils/raster/raster.py
```diff
@@ -2464,15 +2480,29 @@ def icrop(
         if len(bbox) != 4:
             raise ValueError("bbox must be a list or tuple of four integers: (xmin, ymin, xmax, ymax).")
 
-        # Convert pixel coordinates to georeferenced coordinates using the transform
```
rhugonnet, line 2467: The argument description above was ambiguous for me:

>         :param bbox: Pixel-based bounding box as (xmin, ymin, xmax, ymax) in pixel coordinates.

Maybe similar language as in the function description "Bounding box based on indices of the raster array".

We could also add a note in the general description to clarify which this is strikingly different, something like: "Note that, by convention, a raster array has the Y axis as first in the array, and that this Y axis is also flipped (first index corresponds to the highest-value coordinate)."

geoutils/raster/raster.py
```diff
@@ -2464,15 +2480,29 @@ def icrop(
...
-        bottom_left = self.transform * (bbox[0], self.height - bbox[1])
-        top_right = self.transform * (bbox[2], self.height - bbox[3])
+        # Crop the raster data based on the given bounding box (bbox), limit the range to the borders of the raster
+        crop_img = self.data[..., bbox[0] : min(self.height, bbox[2]), bbox[1] : min(bbox[3], self.width)]
+
+        # Create a new affine transform that corresponds to the cropped area
```
rhugonnet, line 2486: Took me a little while to understand why this transform looked like a translation!
When we crop within the extent, the new array size determines the new height/width, so it is only a translation for the affine transform!
In this case, let's use our tested `_translate` function to make it easier to read/debug for future us :wink::
https://github.com/GlacioHack/geoutils/blob/b4da49b8e5ff26f3120085c01cbaedbc1ee1b10b/geoutils/raster/geotransformations.py#L533

tests/test_raster/test_raster.py
```diff
@@ -1818,28 +1818,33 @@ def test_icrop(self, example: str, bbox: list[int] | tuple[int, ...]) -> None:
         raster = gu.Raster(example)
 
         # Call the icrop method with the given bounding box, returning a cropped raster
-        raster_cropped = raster.icrop(bbox=bbox)
+        raster_icropped = raster.icrop(bbox=bbox)
 
         # Compute the expected affine transform for the cropped raster
         transformed_cropped = rio.transform.Affine(
```
rhugonnet, line 1824: Can use `_translate` here too.

rhugonnet (approved):
Perfect, thanks!

geoutils/raster/raster.py
```diff
@@ -2464,15 +2480,29 @@ def icrop(
         if len(bbox) != 4:
             raise ValueError("bbox must be a list or tuple of four integers: (xmin, ymin, xmax, ymax).")
 
-        # Convert pixel coordinates to georeferenced coordinates using the transform
-        bottom_left = self.transform * (bbox[0], self.height - bbox[1])
-        top_right = self.transform * (bbox[2], self.height - bbox[3])
+        # Crop the raster data based on the given bounding box (bbox), limit the range to the borders of the raster
+        crop_img = self.data[..., bbox[0] : min(self.height, bbox[2]), bbox[1] : min(bbox[3], self.width)]
```
rhugonnet, line 2484: Ah sorry I didn't notice this line and @adebardo's comment, good catch.
Indeed, this will load the entire array, then subset it. To circumvent this, we need to use `crop()` as before.

rhugonnet: We should add a test checking if `Raster.is_loaded` is still `False` after calling `icrop()`.

vschaffn:
@rhugonnet @adebardo thanks for your feedback !
I have corrected the `Raster.icrop` method : 
- If the raster is not loaded, I use `_load_rio` with a window to just load the cropped raster
- If the raster is already loaded I kept the same way to do
- I replaced the transform translation by `Raster.translate`

I also fixed a sign error in `_translate`, and added some more tests

The loading check can't be set up in the tests because the raster is always loaded after `icrop`

rhugonnet:
Thanks!

I'm not sure I understand the logic of the part of the PR regarding removing the dependency to `crop()`? To me that's a huge benefit. It allows to have a fully consistent behaviour, with lots of tests already written in `crop()`, without having to duplicate the same code (what is done here in the end, by reading with `rio.Windows`).
And it doesn't affect performance: converting 4 indices to 4 coordinates is instantaneous.

vschaffn:
@adebardo @rhugonnet @adehecq I have changed the implementation of the `icrop` as discussed on 28/03/2025. Does this work for you? 😃

rhugonnet:
Perfect for me! Great idea to make it work through a `distance_unit` argument for consistency with other functions :wink:

adebardo:
Thanks for the work @vschaffn ! :) and @rhugonnet for the review :) I merge

PR #663, opened 2025-03-07
Fix `Raster.get_stats` when an empty `inlier_mask` is passed

### LINKED ISSUE #662: Error in `Raster.get_stats` when an empty `inlier_mask` is passed
As `total_inlier_count` is 0 when an empty `inlier_mask` is passed, it triggered a division by 0 in `Percentage_valid_inlier_points`.

vschaffn:
Resolves #662.

## Changes
- Adds robustness in `Percentage_valid_inlier_ points` when `total_inlier_counts` is 0 to avoid divided by 0.
- Sets all stats to `NaN` when the raster or inlier mask is empty and add a warning in logging.
- Fixes `rmse` (the sum was missing).

## Tests
- Adds a test with an empty inlier mask, ensure returned stats are `NaN`
- Adds a check to ensure that value types in stats are `int` or `float`.

### DIFF
--- a/geoutils/raster/raster.py
+++ b/geoutils/raster/raster.py
@@ -1918,7 +1918,7 @@ class Raster:
             "90th percentile": np.nanpercentile(mdata, 90),
             "LE90": linear_error(mdata, interval=90),
             "NMAD": nmad(data),
-            "RMSE": np.sum(np.sqrt(np.ma.mean(np.square(data)))),
+            "RMSE": np.sqrt(np.ma.mean(np.square(data))),
             "Standard deviation": np.ma.std(data),
             "Valid count": valid_count,
             "Total count": data.size,
--- a/geoutils/raster/raster.py
+++ b/geoutils/raster/raster.py
@@ -1918,7 +1918,7 @@ class Raster:
             "90th percentile": np.nanpercentile(mdata, 90),
             "LE90": linear_error(mdata, interval=90),
             "NMAD": nmad(data),
-            "RMSE": np.sqrt(np.ma.mean(np.square(data))),
+            "RMSE": np.sum(np.sqrt(np.ma.mean(np.square(data)))),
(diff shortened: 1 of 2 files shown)


adebardo:
Can you please look into those -- ? :)

belletva:
> `--` value for most of the statistics and `nan` for 90th percenile and LE90.

I think that we need to be consistent. In my opinion, `nan` could be a good choice for all the statistics.

rhugonnet:
I agree with `nan` for 0 valid values in the mask, more widely supported and understood.

vschaffn:
@adebardo @quinto-clara @rhugonnet @belletva Thanks for your feedback. I set all the stats to `NaN` when the raster is fully masked, and added a warning in the logs indicating that the raster is empty.

geoutils/raster/raster.py
```diff
@@ -1918,7 +1918,7 @@ def _statistics(self, band: int = 1, counts: tuple[int, int] | None = None) -> d
             "90th percentile": np.nanpercentile(mdata, 90),
             "LE90": linear_error(mdata, interval=90),
             "NMAD": nmad(data),
-            "RMSE": np.sqrt(np.ma.mean(np.square(data))),
+            "RMSE": np.sum(np.sqrt(np.ma.mean(np.square(data)))),
```
rhugonnet, line 1921: Actually, reading this again, I think the previous formula was correct.
The sum is included in the mean: `np.mean(X)` = `np.sum(X)/len(X)`, which is the inside term of the root of the RMSE.

RMSE = SQRT ( MSE ) = SQRT ( MEAN ( SQUARE ( ERROR ) ) )

and here ERROR = data.

vschaffn: My bad, you are right, I do not know why I have changed it whereas it was correct. It is changed now 😃

adebardo (approved):

PR #665, opened 2025-03-18
Add handling of failure case for `merge_rasters()` `algo` check

ehusby:
**Bug found with Python 3.12.7**

When a call to `raster/multiraster.merge_rasters()` uses a lambda function as `merge_algorithm` like this:
`merge_algorithm=lambda x: np.logical_and(*x)`
the `TypeError` on unexpected kwarg `axis` is not handled because the error message is different from the hardcoded expected error message structure.

Adds support for this error message structure.

```
  File "/home/ehusby/miniforge3/envs/dem-slice/lib/python3.12/site-packages/geoutils/raster/multiraster.py", line 312, in merge_rasters
    merged_data.append(algo(raster_stack.data, axis=0))
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: my_calling_func.<locals>.<lambda>() got an unexpected keyword argument 'axis'
```

### DIFF
--- a/geoutils/raster/multiraster.py
+++ b/geoutils/raster/multiraster.py
@@ -312,7 +312,10 @@ def merge_rasters(
             merged_data.append(algo(raster_stack.data, axis=0))
         # If that doesn't work, use the slower np.apply_along_axis approach.
         except TypeError as exception:
-            if not ("'axis' is an invalid keyword" in str(exception) or "got an unexpected keyword argument 'axis'" in str(exception)):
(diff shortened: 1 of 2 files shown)


rhugonnet:
Again, good catch! Quite annoying that it prevented the custom function from running.
I've added tests to ensure this doesn't reproduce.

rhugonnet (approved):

PR #666, opened 2025-03-20
Add COG format

belletva:
- Add image structure in tags (`geoutils/geoutils/raster/raster.py`)
- Add tests to save a file in COG format (`tests/test_raster/test_raster.py`) 

Do not check COMPRESSION tags for the following reasons:
The tags of rast and the tags of saved are compared. The file saved has no compression by default (the image is already saved). The tag COMPRESSION is not present in IMAGE_STRUCTURE. The file rast is saved during the test, therefore a compression is applied. The tag COMPRESSION is present in IMAGE_STRUCTURE. Therefore, it is not possible to compare the tag COMPRESSION for both files.

### DIFF
--- a/geoutils/raster/raster.py
+++ b/geoutils/raster/raster.py
@@ -477,7 +477,7 @@ class Raster:
                 converted_tags = decode_sensor_metadata(self.tags)
                 self._tags.update(converted_tags)
                 # Add image structure in tags
-                self._tags.update(ds.tags(ns='IMAGE_STRUCTURE'))
+                self._tags.update(ds.tags(ns="IMAGE_STRUCTURE"))
 
                 self._area_or_point = self.tags.get("AREA_OR_POINT", None)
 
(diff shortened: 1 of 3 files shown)


rhugonnet:
I approved the CI run (won't be needed after the first PR).

Curious that it would fail on the DEFLATE attribute at the start of the test using default parameters (and not COG driver). 
Maybe the new image structure metadata changes the default compression when saving? I'm not sure there.

adebardo (approved):
Do not check COMPRESSION tags -> see PR description

adebardo:
@rhugonnet @adehecq :)

rhugonnet:
OK, good to know! 
Let's also explain the reason in a few words by expanding the comment after "Do not check COMPRESSION tags"  in `test_satimg`, so that next time we bump into this we can understand immediately from the code! :smile:

PR #667, opened 2025-03-24
Add profile to raster metadata and use fork for multiprocessing  

adebardo:
# Small improvements 

- Adding the profile option to the raster metadata
- "fork" seems to be more widely usable than forkserver (especially on the CNES cluster)


tests/test_raster/test_raster.py
```diff
@@ -988,7 +988,7 @@ def test_copy(self, example: str) -> None:
 
         # Check all attributes except name and driver
```
rhugonnet, line 989: Add "profile" this comment

PR #668, opened 2025-03-25
Add `ClassificationLayer` for Raster Classification

### LINKED ISSUE #692: [POC] classif step 1 : Create a base class for classification layers
## Context

To kick off the implementation of classification layers in an object-oriented way, the first step is to define a base `ClassificationLayer` class. This class will act as the foundation for all specific classification types (e.g., `SegmentationClassificationLayer`, `SlopeClassificationLayer`, `FusionClassificationLayer`). The purpose of the base class is to encapsulate the common functionality shared by all classification types while allowing flexibility for subclasses to implement their specific logic.

## Code

In a new python file `classification.py` in xDEM source code, write the `ClassificationLayer` abstract class:

1. **Define Common Attributes**: The base class should define attributes that will be common accross all classification layers, such as:
   * `dem` DEM on which the classification will be applied to.
   * `name` name for the classification layer.
   * `req_stats` list of required statistics to compute (optional, all statistics in `geoutils.Raster` computed by default).
   * `req_stats_classes` list of the classes on which the statistics will be applied (optional, all classes by default).
   * `class_names` dict connecting the class names to the class indexes (set to `None`).
   * `classification` result of the classification, a `geoutils.Mask` object, (set to `None`).
   * `stats_dict` required statistics for required classes in a dict (set to `None`).
2. **Abstract Method**: Since different classification layers will have their own classification logic (e.g., segmentation masks vs. slope ranges), the base class should declare abstract methods that the subclasses are required to implement, such as `apply_classification()`. This method will be used to compute the `classification` attribute, which will be a `geoutils.Mask` object, in which each band will represent one class mask.
3. **Statistics computation**: A common feature across all classification layers is the ability to compute statistics on the classified pixels (e.g., mean, standard deviation). The base class should provide a `get_stats()` method that takes into account the two last attributes. The output, as `stats` attribute, should be a dict, in which the first layer represent the classes, and the second the statistics. This method should use for each required classes the `DEM.set_mask()` to apply the classification mask and the `DEM.get_stats()` method to compute the required statistics. The result will be a dict stored under `stats` attribute.
4. **Saving results**: An other common feature is the ability to save the results. The `save()` method shoud have an `output_dir` in input, and save:
   * The `classification` object with the `Mask.save()` method, under `name.tif`;
   * The `class_name` attribute, that represents the name of each class in a dict, under `name_classes.json`;
   * The `stats` attribute, under `name_stats.json` OR `name_stats.csv`.

## Documentation

We need to start a documentation page on this subject.

vschaffn:
Resolves GlacioHack/xdem#692.

## Context:

This PR introduces the `ClassificationLayer` functionality within the geoutils library. The goal is to provide tools for classifying raster data into discrete classes, whether based on numeric intervals (e.g., binning elevation ranges) or categorical classes (e.g., land cover types). This enhancement improves geospatial workflows by allowing users to easily classify raster data and compute statistics for each class.

## Features:
- `ClassificationLayer` Abstract Base Class:
    - Serves as the foundation for implementing raster classification workflows.
    - Contains methods for applying classification (`apply_classification`), computing statistics (`get_stats`, based on `Raster.get_stats`), and saving results (`save`).

- `Binning` Class:
    - Extends `ClassificationLayer` for classifying continuous raster data using custom binning.
    - Automatically generates class masks based on user-defined bin edges.

- `Segmentation` Class:
    - Classify from pre-defined segmentation masks.

- `Fusion` Class:
   - Combines multiple classification masks into fused classes.

- Save/Export Functionality:
    - Exports classification masks as a multi-band GeoTIFF file.
    - Saves class names as a JSON file.
    - Outputs computed statistics as a CSV file.

## Tests:
Unit tests for Binning to verify:
- Correct creation of class masks based on the given bin ranges.
- Proper computation of statistics for classified raster pixels.
- Coverage for saving/exporting classification results (GeoTIFF, JSON, CSV).

Unit tests for `Segmentation`.
Unit tests for `Fusion`.

## Documentation:
- Adds a new documentation page for classification.

## Example:
```{code-cell} ipython3
# load raster (note, the raster parameter can also be a file path)
raster_file = gu.examples.get_path("exploradores_aster_dem")
raster = gu.Raster(raster_file)

# Define bins
bins = [0, 1000, 2000, 3000, np.inf]

# Create Binning object
binning = gu.raster.Binning(raster, "elevation", bins)

# Apply binning
binning.apply()

# Compute statistics
# if `stats` is None, all stats are calculated by default
# if `classes` is None, stats are computed for all classes by default
binning.get_stats(stats=["mean", "std"], classes=["[0, 1000)", "[3000, inf)"])

# Save results
binning.save("elevation_binning")
```

### DIFF
--- a/doc/source/classification.md
+++ b/doc/source/classification.md
@@ -33,25 +33,21 @@ It provides methods for setting up raster classification, computing statistics f
   - Computed statistics are saved as a `.csv` file.
 
 ## Apply binning
-{class}`~geoutils.raster.RasterBinning` is a specific implementation of {class}`~geoutils.raster.ClassificationLayer`
+{class}`~geoutils.raster.Binning` is a specific implementation of {class}`~geoutils.raster.ClassificationLayer`
 designed for classifying continuous raster data using binning. This class allows you to define ranges (bins) for continuous raster values
-and classify the pixels into discrete classes based on these ranges.
+and classify the pixels into discrete classes based on these ranges. The resulting classification is stored as a multi-band {class}`~geoutils.Mask` object.
 
-### Example Workflow:
-**1. Load Raster:** Load a raster file (e.g., a Digital Elevation Model).
-
-**2. Define Bins:** Set the bin ranges to classify the raster data (e.g., elevation ranges).
-
-**3. Apply Classification:** Classify the raster data based on the bin ranges.
-
-**4. Compute Statistics:** Compute various statistics for each class.
-
-**5. Save Results:** Save the classification and statistics to files.
+### Example workflow:
+1. **Load raster:** Load a raster file.
+2. **Define bins:** Set the bin ranges to classify the raster data (e.g., elevation ranges).
+3. **Apply classification:** Classify the raster data based on the bin ranges.
+4. **Compute statistics:** Compute various statistics for each class.
+5. **Save results:** Save the classification and statistics to files.
 
 ```{code-cell} ipython3
 import numpy as np
 import geoutils as gu
-from geoutils.raster import RasterBinning
+from geoutils.raster import Binning
 
 # load raster (note, the raster parameter can also be a file path)
 raster_file = gu.examples.get_path("exploradores_aster_dem")
@@ -60,8 +56,8 @@ raster = gu.Raster(raster_file)
 # Define bins
 bins = [0, 1000, 2000, 3000, np.inf]
 
-# Create RasterBinning object
-binning= RasterBinning(raster, "elevation", bins)
+# Create Binning object
+binning= Binning(raster, "elevation", bins)
 
 # Apply binning
 binning.apply()
@@ -87,23 +83,15 @@ os.rmdir("elevation_binning")
 print(binning.stats_df)
 ```
 
-### Methods:
-- {func}`~geoutils.raster.RasterBinning.apply` classifies the raster based on the specified bins.
-It assigns each pixel to the appropriate class based on which bin the pixel value falls into.
-The resulting classification is stored as a multi-band {class}`~geoutils.Mask` object.
-
-
 ## Apply segmentation
-
 {class}`~geoutils.raster.Segmentation` is a subclass of {class}`~geoutils.raster.ClassificationLayer` designed for
 categorical classification from pre-defined segmentation masks. It allows assigning class names to each mask and computing related statistics.
 
-### Example Workflow:
-
-1. **Load Raster:** Load a raster (e.g., satellite image).
-2. **Provide Masks:** Supply segmentation masks as a {class}`~geoutils.Mask` or boolean NumPy array.
-3. **Assign Class Names:** Provide a dictionary mapping class IDs to names.
-4. **Apply Classification & Save Results:** Optionally compute stats or export masks.
+### Example workflow:
+1. **Load Raster:** Load a raster.
+2. **Provide masks:** Supply segmentation masks as a {class}`~geoutils.Mask` or boolean NumPy array.
+3. **Assign class names:** Provide a dictionary mapping class IDs to names.
+4. **Apply classification & save results:** Optionally compute stats or export masks.
 
 ```{code-cell} ipython3
 from geoutils.raster import Segmentation
@@ -120,7 +108,7 @@ class_names = {0: "Water", 1: "Land"}
 seg = Segmentation(raster, "water", mask_array, class_names)
 
 # Compute stats
-seg.get_stats()
+seg.get_stats(req_stats=["mean", "median"])
 
 # Save stats
 seg.save("segmentation_output")
@@ -136,4 +124,68 @@ os.rmdir("segmentation_output")
 ```{code-cell} ipython3
 # print stats
 print(seg.stats_df)
+```
+
+## Apply fusion
+{class}`~geoutils.raster.Fusion` is a subclass of {class}`~geoutils.raster.ClassificationLayer` designed to combine multiple classification masks into fused categories.
+It computes all possible combinations across input class layers, generating new composite classes and statistics accordingly.
+
+### Example Workflow:
+1. **Load Raster:** Load a raster.
+2. **Define classifications:** Use classification layers like {class}`~geoutils.raster.Binning` or {class}`~geoutils.raster.Segmentation`.
+3. **Create fusion layer:** Combine classifications nto a single fused output.
+4. **Compute stats & save results:** Analyze and export the fused classification.
+
+```{code-cell} ipython3
+from geoutils.raster import Fusion
+
+# Load raster
+raster = gu.Raster(gu.examples.get_path("exploradores_aster_dem"))
+
+# Define binning thresholds
+bins1 = [0, 1000, 2000, 3000, np.inf]
+bins2 = [0, 1500, 2500, np.inf]
+
+# Create classification layers
+classifier1 = Binning(raster=raster, name="bin1", bins=bins1)
+classifier2 = Binning(raster=raster, name="bin2", bins=bins2)
+classifier1.apply()
+classifier2.apply()
+
+# Create Fusion classification
+fusion = Fusion(
+    raster=raster,
+    name="fusion",
+    layers_to_fuse=[
+        classifier1.classification_masks.data,
+        classifier2.classification_masks.data,
+    ],
+    layers_class_names=[
+        classifier1.class_names,
+        classifier2.class_names,
+    ],
+)
+
+# Apply fusion
+fusion.apply()
+
+# Compute stats
+fusion.get_stats()
+
+# Save results
+fusion.save("fusion_output")
+```
+
+```{code-cell} ipython3
+:tags: [remove-cell]
+import os
+os.remove("fusion_output/fusion.tif")
+os.remove("fusion_output/fusion_classes.json")
+os.remove("fusion_output/fusion_stats.csv")
+os.rmdir("fusion_output")
+```
+
+```{code-cell} ipython3
+# print stats
+print(fusion.stats_df)
 ```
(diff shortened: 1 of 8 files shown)


adebardo:
Thanks, this PR will be used as support for 28/03/2025 and debate with the all team in person :)

doc/source/classification.md
```diff
@@ -0,0 +1,93 @@
...
+bins = [0, 1000, 2000, 3000, np.inf]
+
+# Create RasterClassification object
+classification = RasterClassification(raster, "elevation", bins)
+
+# Apply the classification
+classification.apply_classification()
+
+# Compute statistics
+classification.get_stats(req_stats_classes=["[0, 1000)", "[3000, inf)"])
```
adebardo, line 70: I know we discuss this point but we should think about something more user friendly

geoutils/raster/raster.py
```diff
@@ -3856,12 +3856,12 @@ def __init__(
...
-                    category=UserWarning,
-                    message="Multi-band raster provided to create a Mask, only the first band will be used.",
-                )
-                self._data = self.data[0, :, :]
+            # if self.count > 1:
+            #     warnings.warn(
+            #         category=UserWarning,
+            #         message="Multi-band raster provided to create a Mask, only the first band will be used.",
+            #     )
+            #     self._data = self.data[0, :, :]
```
adebardo, line 3864: careful #

tests/test_raster/test_raster.py
```diff
@@ -2108,7 +2109,7 @@ def test_init(self, example: str) -> None:
         # Check the nodata
         assert mask.nodata is None
         # Check the nbands metadata
-        assert mask.count == 1
+        # assert mask.count == 1
```
adebardo, line 2112: careful #

adebardo (commented):
Its not a real review, I just look a it for a minute, thanks for work again

geoutils/raster/raster.py
adehecq: I would add a quick comment to explain why this was commented, for later reference.

geoutils/raster/classification.py
```diff
@@ -0,0 +1,319 @@
...
+
+        :param raster: Raster on which the classification masks should be applied for computing the stats.
+        :param stats: List of required statistics to compute (optional, default is all statistics in
+            geoutils.Raster.get_stats).
+        :param classes: List of classes on which to compute statistics (optional, default is all classes).
+
+        :raise ValueError: if req_stats_classes are not class names.
+        """
+        if self.classification_masks is None or self.class_names is None:
+            raise ValueError("Classification has not been applied yet. Call apply_classification() first.")
```
adehecq, line 80: Replace "apply_classification" -> apply

adehecq:
Thanks @vschaffn for the implementation. This will certainly be useful for mainy users!
I summarize here my suggestions following the meeting from this morning:
- I find it a bit difficult to grasp the differences between the different classes (ClassificationLayer, Binning, Segmentation, Fusion). I would suggest simplifying to a single class, with different methods/inputs to handle different cases. In short, the Segmentation class could take 1) a single raster with discrete elements (for example a land cover classification containing values 1, 2, 3 etc...) 2) multiple binary rasters (Masks) for each classification 3) a continuous raster (dem, slope...) and a binning strategy for discretizing the values. Then the `get_stats` method would loop through the different values or layers in the input raster.
- in case the previous point is not implemented, I would suggest removing the "shortcut" to geoutils.rasters.classification, to clarify for example that `Fusion` means fusion of classifiers.
- I would add a method `read` that would do the reverse of `save
- I suggest adding a landcover dataset in the example datasets for documenting and testing these features. For example, the [Corine land cover](https://land.copernicus.eu/en/products/corine-land-cover). Unfortunately, I don't think this extends outside Europe, so not possible to have it with the aster_exploradores dataset...
- I would add an issue to remember to remove redundancy of similar functionalities in xdem (e.g., binning)

rhugonnet:
I was about to write my feedback in the issue following previous comments there (https://github.com/GlacioHack/xdem/issues/692#issuecomment-2727557256), but seeing @adehecq's comment I'll follow up here as well.
I already detailed a lot ideas in these previous comments, notably on relying on SciPy or Pandas/Xarray directly, so I refer to those previous comments and I'm only thinking further ahead here.

**Taking stock of existing tools:**
 I did some additional reading on this topic (as things have been evolving in the last years, in Pandas and Xarray), so that we don't re-invent the wheel on this.
I think the **best resource about this topic is the following Xarray page that we should all read**: https://docs.xarray.dev/en/stable/user-guide/groupby.html#groupby-group-and-bin-data. The related pandas doc is here: https://pandas.pydata.org/pandas-docs/stable/user_guide/groupby.html.
In short, the utility `GroupBy` allows to group categorical values with a `UniqueGrouper` (what we call "segmentation" here) or to group continuous values with a `BinGrouper` (what we call "binning" here), and supports multiple indexing (what we call "fusion" here). So it basically includes of the functionalities we need, and is accelerated natively in Xarray and through the recent development of Flox: https://xarray.dev/blog/flox.

**So I think the big question is: What's different in our case that would justify developping a new functionality?**

There are two aspects:
**1. Technical aspect: Can we better optimize this for rasters?** We are not using generic rectilinear N-D labelled arrays, but regular-grid 2D array that are georeferenced. Does this have any impact on the operation? In this case, I don't think so: Masking values based on groups (categorical or bins) is the same no matter the shape of the array, or the nature of the grid. So nothing to optimize here.
**2. Convenience aspect: Can we make it more easy to use for rasters?** Here, there are definitely things to do, because `groupby()` is meant to support any N-D input, that are not necessarily georeferenced, so it can be overly complex to approach. We could wrap it to return easier-to-understand objects. 

So, in my view, it's only 2 that is relevant. But then, I don't think we want to create a completely new API when there is already a good abstraction covering all of our use cases that exists in Pandas/Xarray.

So I'd be in favour of adding a `geoutils/stats/group.py` and a `Raster.groupby()` function that wraps Pandas/Xarray `groupby()`, potentially with additional options specific to rasters (`return_masks` for instance).
This could probably support saving masks and stats fairly easily, such as:
```python
stats, masks = dem.groupby(slope, slope_bins, return_masks=True)
stats.to_csv(out_csv)
masks.save(out_tif)
```
If that API is still deemed too difficult by CNES, we could maintain the classes above to wrap up the operations, but it would duplicate things into a second API.

rhugonnet:
Thinking a bit more about this, a couple additional points:
1. To bin in Pandas, one always need to combine `pd.groupby()` with `pd.cut()`. In Xarray, they simplified that by adding a `groupby_bins` function which repeats all the argument of `pd.cut()`/`BinGrouper` to streamline things for the user (https://docs.xarray.dev/en/stable/generated/xarray.Dataset.groupby_bins.html#xarray.Dataset.groupby_bins). We could do the same for `Raster`. We could also think of accepting `bins=None` to indicate a categorical variable, this way we could combine the API in a single function.
2. A `groupby()` call doesn't return the reduced result directly, but an iterator of (bin/category, groups), which then can be used to apply the desired reducer function. The difference in Xarray is that the coordinates are preserved in the grouping by returning a DataArray with the right dimensions and subset coordinates. For a Raster, this is not possible anyway, so we can leave it aside. 
3. Finally, I realized **our main structural difference with Xarray/Pandas is that we have a single variable being grouped** (if we consider a single-band Raster, which is 95% of the time). Because of this, having the Pandas logic described in point 2 is less useful, because one of the main purposes of the "Grouped" dataframe object of Pandas is being able to then use different reducers on the different variables that were grouped later on, without re-running the grouping.

So, to simplify the API for a Raster and avoid the more complex logic of Pandas, we could immediately have a higher-level function `grouped_statistics` that already calls both `pd.cut()` where necessary, and `.apply()/.agg()` on the grouped output of `self.data`, because that's the only one we can compute anyway. So it would look like this:
```python
class Raster
    def grouped_statistic(self, band=1, vars, bins, apply_funcs):
         df_self = pd.DataFrame([self.data[band-1].ravel(), *(v.ravel() for v in vars)])
         return df_self.groupby([var1, pd.cut(var1, bins1, ...)], [...]).agg(apply_funcs)
```
For this reason (and why it's like that in xDEM's `nd_binning` now that I look back on this), we could use SciPy's `binned_statistic_dd`, we would only need to use a small trick for defining the bins of the categorical data. It should be equivalent. 
Both support binning and multi-indexing otherwise.

rhugonnet:
So the question we need to answer to move forward is: What would be missing in the functionalities of such a `Raster.grouped_statistic()` method to justify building a class wrapper on top?

For getting/saving the georeferenced masks, there could be an optional dictionary output using the same group name logic coded in this PR, with group key copied from the output dataframe, and all could be saved doing:
```python
stats, mask_dict = dem.grouped_statistics(slope, slope_bins, ...)
for name, mask in mask_dict.items():
    mask.save("mask_"+name+".tif")
```

But, if we wanted to have tailored plotting functions (to visualize the stats of 1D/2D groups), it'd be something hard to fit as an output of a function call (and we use it a lot in xDEM), and a class would be really useful.

Finally, having a CLI functionality that wraps this (reading + binning + plotting + saving) all at once might also simplify things for end-users?

These are all my thoughts! :smile:

adehecq:
Just a note on my suggestion of adding a landcover classification. I think we could use the ESA worldcover. An example for accessing the data through the AWS bucket is shown here: https://github.com/ESA-WorldCover/esa-worldcover-datasets.

belletva:
Thank you @rhugonnet and @adehecq for your comments. It is nice to see that the topic is generating a lot of interest. However, I suggest that these ideas should correspond to new issues. It might be interesting to merge if there are no blocking points and propose new issues to be addressed later.

rhugonnet:
Closing with #925

PR #669, opened 2025-03-25
Generic Multiprocessing Functions for Raster Processing

### LINKED ISSUE #670: Add support for generic multiprocessing functions for raster operations in geoutils
## Context
In the geoutils library, there is a growing need to support out-of-memory raster operations, especially when dealing with large datasets.

## Proposal
- Generic Multiprocessing Functions: Introduce a set of generic functions to handle the distribution of raster computations with tiling and multiprocessing.
- Tile Extraction and Padding: Implement utility functions for extracting raster tiles with optional padding to prevent edge effects during computation.
- Map-Reduce Style: Create a map_block functions that allow applying arbitrary operations on each raster tile, handling both in-memory and disk-based rasters.
- Reusability: The functions should be general enough to handle multiple use cases, such as filtering and terrain analysis.

vschaffn:
Resolves #670.

## Context

To enable efficient raster processing on large datasets without exceeding memory limits, this PR introduces generic multiprocessing functions within `geoutils`. These functions allow users to apply any processing function to raster data using a tiling approach with overlap handling. The approach minimizes memory usage by processing tiles separately and writing results directly to disk.

## Features

This PR introduces the following key functions:

- `map_overlap_multiproc_save`: This function divides the input raster into overlapping tiles, processes them in parallel using
    multiprocessing, and writes the processed results to an output raster file.
- `map_multiproc_collect`: This function splits an input raster into overlapping tiles, processes them in parallel, and returns the results as a list. It is intended for cases where `func` does *not* return a `Raster`, but instead returns arbitrary values (e.g., numerical statistics, feature extractions, etc.).
- `apply_func_block`: A helper function that loads a specific raster tile, applies the provided function, and removes padding to avoid edge effects.
- `load_raster_tile`: Loads a specific tile from a raster based on given bounding box coordinates.
- `remove_tile_padding`: Removes extra padding from tiles after processing to mitigate edge artifacts.
- `MultiprocConfig`: Configuration class for handling multiprocessing parameters.

These functions are designed to be generic and reusable for various raster-processing tasks.

## Tests

To ensure the correctness of these functions, tests have been implemented:
- Running `map_multiproc` with a simple function (Raster.copy()) to verify that tiles are processed correctly and raster is not loaded during processing.
- Verifying that `load_raster_tile` load the right tile.
- Verifying that padding removal correctly restores tile boundaries without artifacts.
- Ensuring that multiprocessing runs correctly across different tile sizes and clusters.

## Documentation
A documentation page has been written to explain how to use these generic multiprocessing functions, and they have been added to the API.

## Example Usage

A basic example demonstrating the usage of `map_overlap_multiproc_save` with a simple function (`Raster.copy()`):

```{code-cell} ipython3
# Input raster path
input_raster_path = "path/to/large_raster.tif"

# Define function to apply to the raster
cast_nodata=True
def copy_func(r: RasterType, cast_nodata: bool) -> RasterType:
    return r.copy(cast_nodata=cast_nodata)

# Define config for multiprocessing:
config = MultiprocConfig(
    chunk_size=200, 
    outfile = "path/to/output_raster.tif", 
    cluster=ClusterGenerator("multi", nb_workers=4)
)

# Run multiprocessing with the copy function
map_overlap_multiproc_save(
    copy_func, 
    raster_path, 
    config,
    cast_nodata,
    depth=0
)
```

### DIFF
--- a/geoutils/raster/distributed_computing/multiproc.py
+++ b/geoutils/raster/distributed_computing/multiproc.py
@@ -17,7 +17,7 @@
 # limitations under the License.
 
 """Process out-of-memory calculations"""
-from typing import Any, Callable, overload
+from typing import Any, Callable
 
 import numpy as np
 import rasterio as rio
@@ -36,24 +36,20 @@ class MultiprocConfig:
     Configuration class for handling multiprocessing parameters in raster processing.
 
     This class encapsulates settings related to multiprocessing, allowing users to specify
-    chunk size, depth, output file, and an optional cluster for parallel processing.
+    chunk size, output file, and an optional cluster for parallel processing.
     It is designed to be passed into functions that require multiprocessing capabilities.
     """
 
-    def __init__(
-        self, chunk_size: int, outfile: str | None = None, depth: int = 0, cluster: AbstractCluster | None = None
-    ):
+    def __init__(self, chunk_size: int, outfile: str | None = None, cluster: AbstractCluster | None = None):
         """
         Initialize the MultiprocConfig instance with multiprocessing settings.
 
         :param chunk_size: The size of the chunks for splitting raster data.
         :param outfile: The file path where the output will be written.
-        :param depth: The overlap size between chunks to prevent edge effects (default is 0).
         :param cluster: A cluster object for distributed computing, or None for sequential processing.
         """
         self.chunk_size = chunk_size
         self.outfile = outfile
-        self.depth = depth
         if cluster is None:
             # Initialize a basic multiprocessing cluster if none is provided
             cluster = ClusterGenerator("basic")  # type: ignore
@@ -61,7 +57,7 @@ class MultiprocConfig:
         self.cluster = cluster
 
     def copy(self) -> "MultiprocConfig":
-        return MultiprocConfig(chunk_size=self.chunk_size, outfile=self.outfile, depth=self.depth, cluster=self.cluster)
+        return MultiprocConfig(chunk_size=self.chunk_size, outfile=self.outfile, cluster=self.cluster)
 
 
 def load_raster_tile(raster_unload: RasterType, tile: NDArrayNum) -> RasterType:
@@ -142,36 +138,14 @@ def apply_func_block(
     return result_tile, tile
 
 
-@overload
-def map_overlap_multiproc(  # type: ignore
+def map_overlap_multiproc_save(
     func: Callable[..., RasterType],
     raster_path: str | RasterType,
     config: MultiprocConfig,
     *args: Any,
-    return_tiles: bool = False,
+    depth: int = 0,
     **kwargs: dict[str, Any],
-) -> None: ...
-
-
-@overload
-def map_overlap_multiproc(
-    func: Callable[..., Any],
-    raster_path: str | RasterType,
-    config: MultiprocConfig,
-    *args: Any,
-    return_tiles: bool = False,
-    **kwargs: dict[str, Any],
-) -> list[Any]: ...
-
-
-def map_overlap_multiproc(
-    func: Callable[..., Any],
-    raster_path: str | RasterType,
-    config: MultiprocConfig,
-    *args: Any,
-    return_tiles: bool = False,
-    **kwargs: dict[str, Any],
-) -> None | list[Any]:
+) -> None:
     """
     Multiprocessing function for applying out_of_memory operations on raster.
 
@@ -182,10 +156,9 @@ def map_overlap_multiproc(
 
     :param func: The function to apply to each tile.
     :param raster_path: Path to the input raster or the Raster object itself.
-    :param config: Configuration object containing chunk size, output file, depth, and optional cluster. Output file is
+    :param config: Configuration object containing chunk size, output file, and optional cluster. Output file is
         needed if func returns a Raster or a tuple with a Raster.
     :param args: Additional arguments to pass to the function being applied.
-    :param return_tiles: True to return the tiles coordinates for each tile in the list
     """
     # Load DEM metadata if raster_path is a filepath, otherwise use the Raster object
     if not isinstance(raster_path, Raster):
@@ -194,7 +167,7 @@ def map_overlap_multiproc(
         raster = raster_path
 
     # Generate tiling grid
-    tiling_grid = compute_tiling(config.chunk_size, raster.shape, raster.shape, overlap=config.depth)
+    tiling_grid = compute_tiling(config.chunk_size, raster.shape, raster.shape, overlap=depth)
 
     # Create tasks for multiprocessing
     tasks = []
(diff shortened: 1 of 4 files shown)


rhugonnet:
Great! All good for me on the generic implementation, following our discussion in https://github.com/GlacioHack/xdem/pull/704  :wink:.

I'll just add some line-to-line comments on little things we can adjust to make the namings more inter-changeable with Dask, which will be especially useful once we start documenting things!

geoutils/raster/distributed_computing/multiproc.py
```diff
@@ -0,0 +1,194 @@
...
+    # Apply user-defined function to the tile
+    result_tile = func(raster_tile, *args, **kwargs)
+
+    # Remove padding
+    remove_tile_padding(raster, result_tile, tile, depth)
+
+    return result_tile, tile
+
+
+def map_multiproc(
```
rhugonnet, line 109: Maybe `map_overlap_multiproc()`? To differentiate from `map_overlap` of dask.

geoutils/raster/distributed_computing/multiproc.py
```diff
@@ -0,0 +1,194 @@
...
+    func: Callable[..., Any],
+    raster_path: str | RasterType,
+    tile_size: int,
```
rhugonnet, line 112: I think we should use `chunks` instead of `tile_size` for this main function to keep naming consistent for users (see https://docs.dask.org/en/stable/array-chunks.html#specifying-chunk-shapes).
For internal functions, we can keep "tile" syntax if we want.

geoutils/raster/distributed_computing/multiproc.py
```diff
@@ -0,0 +1,194 @@
...
+        ymin += padding
+    if tile[3] != raster.width:
+        tile[3] -= padding
+        ymax -= padding
+
+    # Apply the new bounding box to crop the tile and remove the padding
+    raster_tile.icrop(bbox=(xmin, ymin, xmax, ymax), inplace=True)
+
+
+def map_block(
```
rhugonnet, line 78: We could name it `apply_func_block` here instead (`map` is usually for the function that distributes the function to all blocks).

geoutils/raster/distributed_computing/multiproc.py
```diff
@@ -0,0 +1,194 @@
...
+    :param tile: The bounding box of the tile as [xmin, xmax, ymin, ymax].
+    :return: The extracted raster tile.
+    """
+    xmin, xmax, ymin, ymax = tile
+    # Crop the raster to extract the tile based on the bounding box coordinates
+    raster_tile = raster_unload.icrop(bbox=(xmin, ymin, xmax, ymax))
+    return raster_tile
+
+
+def remove_tile_padding(raster: RasterType, raster_tile: RasterType, tile: NDArrayNum, padding: int) -> None:
```
rhugonnet, line 48: I think we can pass a `shape` argument instead of `raster` here as it contains height and width, which are the only attributes we need.

tests/test_raster/test_distributing_computing/test_multiproc.py
```diff
@@ -0,0 +1,103 @@
...
+    @pytest.mark.parametrize("example", [aster_dem_path, landsat_rgb_path])  # type: ignore
+    def test_map_block(self, example):
+        """
+        Test applying a function to a raster tile and handling padding removal.
+        """
+        raster = Raster(example)
+        tile = np.array([100, 200, 100, 200])  # [xmin, xmax, ymin, ymax]
+        padding = 10
+
+        cast_nodata = True
```
rhugonnet, line 70: As the `cast_nodata` argument has no effect in this context, we can't be sure it was passed, or if the call will work for several arguments or for keyword arguments.
Maybe instead the custom function can be something like
```python
custom_func(raster, addition, factor, new_type=None):
     return ((raster + addition) * factor).astype(new_type)
```

tests/test_raster/test_distributing_computing/test_multiproc.py
```diff
@@ -0,0 +1,103 @@
...
+    def test_map_multiproc(self, example, tile_size, cluster):
+        """
+        Test the multiprocessing map function with a simple copy operation.
+        """
+        raster = Raster(example)
+        output_file = "output.tif"
+
+        cast_nodata = True
+        # Apply the multiproc map function
+        map_multiproc(_copy_func, raster, tile_size, output_file, (cast_nodata,), depth=10, cluster=cluster)
```
rhugonnet, line 90: Here we could check that the overlap works as intended with a custom function that uses neighbouring cells.
Maybe something simple like a SciPy filter:
```python
def custom_func_overlap(raster, size):
    return scipy.ndimage.maximum_filter(raster.data, size)
```

Where size is smaller than the depth (could even check that the results do differ when it's not!).

vschaffn:
@rhugonnet thanks for your feedback, I have modified the little things you have mentioned in your review :smiley:

tests/test_raster/test_distributing_computing/test_multiproc.py
```diff
@@ -0,0 +1,119 @@
...
+        raster_tile = load_raster_tile(raster, tile)
+        raster_tile_with_padding = load_raster_tile(raster, tile_pad)
+
+        # Remove padding and ensure it's back to the original size
+        remove_tile_padding((raster.height, raster.width), raster_tile_with_padding, tile, padding)
+        assert raster_tile_with_padding.raster_equal(raster_tile)
+
+    @pytest.mark.parametrize("example", [aster_dem_path, landsat_rgb_path])  # type: ignore
+    @pytest.mark.parametrize("padding", [0, 1, 10])  # type: ignore
+    def test_map_block(self, example, padding):
```
rhugonnet, line 73: Update to `test_apply_func_block` here too!

vschaffn:
@rhugonnet I have made a few changes to adapt `map_overlap_multiproc` to functions that don't necessarily return a raster.
I just have a problem with mypy, when overloading `map_overlap_multiproc`, as the Callable[... Any] signature also includes Callable[..., RasterType], I would need to be able to define an Any but not RasterType type, which doesn't seem possible in python. For the moment I've ignored the error. Do you have any ideas?

I have also created a `MultiprocConfig` class, which we can pass as a parameter to functions where we want a multiprocessing equivalent, and if this parameter is not None, activate the multiprocessing version (just like an equivalent of `chunk_size` in Dask, but with additionnal parameters).

rhugonnet:
I see... I think you won't be able to overload here, you might have to `# type: ignore`.

In Xarray, they use `map_overlap/block` only for functions working to smaller Xarray objects (see the "Notes" below the function description: https://docs.xarray.dev/en/stable/generated/xarray.map_blocks.html).
For function working on the array inside of the Xarray objects (not requiring any Xarray metadata), they use `apply_ufunc` instead: https://docs.xarray.dev/en/stable/generated/xarray.apply_ufunc.html#xarray.apply_ufunc.

Potentially, we could also mirror that structure.

rhugonnet:
And great idea for the `MultiprocConfig`, it didn't cross my mind! It will simplify the API a lot for multi-processing calls, and make the behaviour easy to explain :slightly_smiling_face: 

I'm not sure if we need to have `depth` in it?
For instance, if we have `DEM.slope(mp_config=)`, it will be passed to a `slope_multiproc()` function wrapping `map_overlap_multiproc()`. This `slope_multiproc()` should convert the `window` argument of `slope()` into the `depth` needed for the operation.
So I think the user would never have to specify `depth` through a defined function (only if they call directly `map_overlap`, but in that case it is a keyword argument)?

vschaffn:
@rhugonnet @adehecq @adebardo I have changed the implementation of the multiprocessing as discussed on 28/03/2025 :
- `map_overlap_multiproc` has been seperate in two functions:
    - `map_overlap_multiproc_save` write a new raster on disk.
    - `map_multiproc_collect` returns a list of required information by tile.
- `depth` has been removed of `MultiprocConfig`.
- A documentation page has been written, the function has been added to the API.

rhugonnet:
Perfect, thanks!

For the documentation page: 
As this is an early version of the API for these functions that is likely to evolve quickly, I'm not sure we're fully ready to document those yet. We've seen it with the evolving needs in xDEM, and there are still many cases we'll need and haven't covered yet that will trigger changes (loading multiple rasters at once on the same tile, or returning multiple rasters for one operation). 
So I propose to keep `multiprocessing.md` as a draft without rendering it yet. And we wait until showing changes in `api.md` and `index.md` for a bit longer, until we've finalized the various multiprocessing features across GeoUtils-xDEM, and we know the API won't change.

Other small remarks:
- All internal functions not planned to be listed in the API should be non-public and preceded by an underscore, like `_apply_func_block`,
- (I mentioned this in Slack) Our documentation is written in British lower-case capitalization style everywhere. We should keep it consistent, so "Example: Applying a Raster Filter" into "Example: applying a raster filter".

On this last point, we could think of defining a `STYLE.md` or equivalent, to clearly list the rules there and agree on it all together :wink:. And we could agree on those same style rules for the few things that show on the GitHub page of the package (Releases, PR titles).

PR #672, opened 2025-03-26
Fix NumPy integer overflow to get count in `subdivide_array`

### LINKED ISSUE #671: use of `np.prod()` in `subdivide_array()` causes 32bit overflow
### Reproducing 

Fair warning, trying to reproduce this uses like 20GB Memory

```
import geoutils as gu
import numpy as np
array = np.zeros((61812, 43967), dtype=np.uint8)
gu.raster.tiling.subdivide_array(array.shape, 2)
```
File "C:\Users\subora\.conda\envs\xdem\Lib\site-packages\geoutils\raster\tiling.py", line 119, in subdivide_array
    raise ValueError(f"Shape '{shape}' size ({np.prod(shape)}) is smaller than 'count' ({count}).")
ValueError: Shape '(61812, 43967)' size (-1577279092) is smaller than 'count' (2).

### Environment
OS - Windows Server 2019
geoutils installed from mamba v2.0.7
python v3.12.9

### Comments
From my very little research, this seems to be an issue with numpy compiled with MSVC. I believe that MSVC's use of `long` is a 32bit integer. From testing on Linux, I did not receive the error, and the error went away when forcing `dtype = np.int64`

asub-sandwich:
- [x] Resolves #671

### DIFF
--- a/geoutils/raster/tiling.py
+++ b/geoutils/raster/tiling.py
@@ -19,6 +19,7 @@
 """Tiling tools for arrays and rasters."""
 
 import math
+import sys
 
 import numpy as np
 from matplotlib.patches import Rectangle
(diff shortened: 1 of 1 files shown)


rhugonnet:
Oh wow thanks, good catch, never thought `count` could get to such a high value!

For info: If you need to load index-based subdivisions of your raster without having everything in memory, you can use `Raster.icrop()`. Combined with  `compute_tiling()` as is just being done in https://github.com/GlacioHack/geoutils/pull/669, you can map a function to all subdivisions, possibly with overlap, never loading everything at once.

In a couple months we should support this through our Xarray accessor (and thus through Dask's map_block/overlap), or through Multiprocessing as in #669. Both options are well on their way :wink:

asub-sandwich:
Oh, thanks for the info! I will be checking that out. Thanks for all the work that gets put into this :)

rhugonnet (approved):

PR #679, opened 2025-04-12
Update `Vector` class to latest Shapely functions

rhugonnet:
The list of non-supported Shapely functions (returned by warnings in our CI) has been growing, so here's an update.

This will be less of a problem once we mirror the `Vector` and `PointCloud` classes by Pandas accessors.


rhugonnet:
Nothing new here on the functionality, structure or else, simply adding latest wrappers.
So merging, can be reviewed a posteriori if anything doesn't look right.

PR #683, opened 2025-04-17
Handle Masks in `map_overlap_multiproc_save`

vschaffn:
When the output of `map_overlap_multiproc_save` is a `Mask`, it is converted to uint8, with a nodata of 255 so that it can be saved via rasterio.


rhugonnet:
Ah yes it's true we need this for a boolean mask, great addition!
We can also add a short test in `test_multiproc` to ensure the behaviour stays the same in the future :wink:

vschaffn:
@rhugonnet it is done 😄

rhugonnet (approved):

PR #691, opened 2025-04-28
Move `filter.py` from xDEM to GeoUtils and integrate into `raster` class

### LINKED ISSUE #690: Move `filter.py` from xDEM to GeoUtils and integrate into raster class
## Context

The current `filter.py` module in **xDEM** provides a collection of filters that are highly useful for digital elevation models (DEMs). However, these filtering utilities are not intrinsically tied to elevation data—they are generally applicable to any geospatial raster data. Given this, it makes more architectural sense for them to live in **GeoUtils**.

### 1. Migrate Code

- Move the `filter.py` module from `xdem/` to `geoutils/filters.py`.
- Remove `gaussian_filter_cv` to avoid dependence on opencv.
- Move the corresponding unit tests from `tests/test_filter.py` in xDEM to `tests/test_filters.py` in GeoUtils.
- Adjust imports and dependencies accordingly.

### 2. Integrate Filters into Raster Class

Add a new method to the `Raster` class:

```python
 def filter(
        self,
        method: str | Callable[..., NDArrayNum],
        inplace: bool = False,
        **kwargs: dict[str, Any]
    ) -> RasterType | None:
        """
        Apply a filter to the raster data.

        @param method: The filter to apply. Can be a string ("gaussian", "median", "mean", "max", "distance")
                       for built-in filters, or a custom callable that takes a 2D ndarray and returns one.
        @param inplace: If True, modifies the current Raster object. If False, returns a new filtered Raster.

        @return: A new Raster instance with the filtered data if `inplace=False`, otherwise None.

        @raises ValueError: If the filter name is not one of the predefined options.
        @raises TypeError: If `method` is neither a string nor a callable.
        """
```

This method should internally call the appropriate functions from `geoutils.filters`.

### 3. Add Documentation

- Update the raster class documentation to include the `.filter()` method.
- Update the API.

### 4. Testing

- Ensure migrated tests are compatible with GeoUtils.
- Add new tests specifically for the `Raster.filter()` method.

### 5. Deprecate in xDEM

- Add a deprecation warning in `xdem/filter.py` pointing users to `geoutils.filters`.

vschaffn:
Resolves #690.

## Context

The current `filter.py` module in **xDEM** provides a collection of filters that are highly useful for digital elevation models (DEMs). However, these filtering utilities are not intrinsically tied to elevation data—they are generally applicable to any geospatial raster data. Given this, it makes more architectural sense for them to live in **GeoUtils**.

## Changes
- Moved the `filter.py` module from `xdem/` to `geoutils/filters.py`.
- Removed `gaussian_filter_cv` to avoid dependence on opencv.
- Moved `tests/test_filter.py` in xDEM to `tests/test_filters.py` in GeoUtils.
- Added a new function `_nan_safe_filter` common for gaussian, mean and median filter, to deal with arrays containing nan values.
- Added a generic `_filter` function mapping the available filters.
- Added the `Raster.filter` method that extracts the array with Nans, uses the `_filter` function, and rebuild the Raster data.

 ## Tests
- Added new tests for the `Raster.filter` method.

## Documentation
- Added a `filter` section in `raster_class.md`.

## Note
When this ticket is approved, a linked PR to remove filters in xDEM will be opened.

### DIFF
--- a/geoutils/filters.py
+++ b/geoutils/filters.py
@@ -1,0 +1,213 @@
+# Copyright (c) 2025 GeoUtils developers
+# Copyright (c) 2025 Centre National d'Etudes Spatiales (CNES)
+#
+# This file is part of the GeoUtils project:
+# https://github.com/glaciohack/geoutils
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+#
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Filters to remove outliers and reduce noise in rasters."""
+from __future__ import annotations
+
+import warnings
+from collections.abc import Callable
+from typing import Any
+
+import numpy as np
+import scipy
+
+from geoutils._typing import NDArrayNum
+
+
+def _filter(array: NDArrayNum, method: str | Callable[..., NDArrayNum], **kwargs: dict[str, Any]) -> NDArrayNum:
+    """
+    Apply a filter to the array. See description of raster.filter.
+    """
+    if isinstance(method, str):
+        filter_map: dict[str, Callable[..., NDArrayNum]] = {
+            "gaussian": gaussian_filter,
+            "median": median_filter,
+            "mean": mean_filter,
+            "max": max_filter,
+            "min": min_filter,
+            "distance": distance_filter,
+        }
+
+        if method not in filter_map:
+            raise ValueError(f"Unsupported filter method '{method}'. " f"Available methods: {list(filter_map.keys())}")
+
+        filter_func = filter_map[method]
+
+    elif callable(method):
+        filter_func = method
+    else:
+        raise TypeError("`method` must be a string or a callable.")
+
+    # Apply filter
+    filtered_data = filter_func(array, **kwargs)
+    return filtered_data
+
+
+def _nan_safe_filter(array: NDArrayNum, filter_func: Callable[..., NDArrayNum], **kwargs: Any) -> NDArrayNum:
+    """Apply a NaN-safe filter using a weighting trick."""
+    # Check that array dimension is 2 or 3
+    if np.ndim(array) not in [2, 3]:
+        raise ValueError(f"Invalid array shape given: {array.shape}. Expected 2D or 3D array.")
+
+    # In case array does not contain NaNs, use scipy's gaussian filter directly
+    if not np.isnan(array).any():
+        return filter_func(array, **kwargs)
+
+    # If array contain NaNs, need a more sophisticated approach
+    # Inspired by https://stackoverflow.com/a/36307291
+    # Run filter on the array with NaNs set to 0
+    array_filled = np.nan_to_num(array, nan=0.0)
+    weights = (~np.isnan(array)).astype(array_filled.dtype)
+
+    filtered_array_filled = filter_func(array_filled, **kwargs)
+    del array_filled
+
+    weight_sum = filter_func(weights, **kwargs)
+    del weights
+
+    with warnings.catch_warnings():
+        warnings.filterwarnings("ignore", message="invalid value encountered")
+        filtered_array = filtered_array_filled / weight_sum
+
+    return filtered_array
+
+
+def gaussian_filter(array: NDArrayNum, sigma: float) -> NDArrayNum:
+    """
+    Apply a Gaussian filter to a raster that may contain NaNs, using scipy's implementation.
+    gaussian_filter_cv is recommended as it is usually faster, but this depends on the value of sigma.
+
+    N.B: kernel_size is set automatically based on sigma.
+
+    :param array: the input array to be filtered.
+    :param sigma: the sigma of the Gaussian kernel
+
+    :returns: the filtered array (same shape as input)
+    """
+    return _nan_safe_filter(array, scipy.ndimage.gaussian_filter, sigma=sigma)
+
+
+def median_filter(array: NDArrayNum, **kwargs: Any) -> NDArrayNum:
+    """
+    Apply a median filter to a raster that may contain NaNs, using scipy's implementation.
+
+    :param array: the input array to be filtered.
+
+    :returns: the filtered array (same shape as input).
+    """
+    return _nan_safe_filter(array, scipy.ndimage.median_filter, **kwargs)
+
+
+def mean_filter(array: NDArrayNum, kernel_size: int, **kwargs: Any) -> NDArrayNum:
+    """
+    Apply a mean filter to a raster that may contain NaNs.
+
+    :param array: the input array to be filtered.
+    :param kernel_size: the size of the kernel.
+
+    :returns: the filtered array (same shape as input).
+    """
+    # Check that array dimension is 2
+    if np.ndim(array) not in [2]:
+        raise ValueError(f"Invalid array shape given: {array.shape}. Expected 2D array.")
+    kernel = np.ones((kernel_size, kernel_size)) / kernel_size**2
+    return _nan_safe_filter(array, scipy.ndimage.convolve, weights=kernel, **kwargs)
+
+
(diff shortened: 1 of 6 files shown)


geoutils/filters.py
```diff
@@ -0,0 +1,213 @@
...
+        warnings.filterwarnings("ignore", message="invalid value encountered")
+        filtered_array = filtered_array_filled / weight_sum
+
+    return filtered_array
+
+
+def gaussian_filter(array: NDArrayNum, sigma: float) -> NDArrayNum:
+    """
+    Apply a Gaussian filter to a raster that may contain NaNs, using scipy's implementation.
+    gaussian_filter_cv is recommended as it is usually faster, but this depends on the value of sigma.
```
rhugonnet, line 94: We can remove the notes about OpenCV filters!

geoutils/filters.py
```diff
@@ -0,0 +1,213 @@
...
+    return np.where(nans, array, array_nans_replaced_f)
+
+
+def distance_filter(array: NDArrayNum, radius: float, outlier_threshold: float) -> NDArrayNum:
+    """
+    Filter out pixels whose value is distant more than a set threshold from the average value of all neighbor \
+pixels within a given radius.
+    Filtered pixels are set to NaN.
+
+    TO DO: Add an option on how the "average" value should be calculated, i.e. using a Gaussian, median etc filter.
```
rhugonnet, line 179: I like this TO-DO a lot! :wink:

geoutils/filters.py
```diff
@@ -0,0 +1,213 @@
...
+
+def median_filter(array: NDArrayNum, **kwargs: Any) -> NDArrayNum:
+    """
+    Apply a median filter to a raster that may contain NaNs, using scipy's implementation.
+
+    :param array: the input array to be filtered.
+
+    :returns: the filtered array (same shape as input).
+    """
+    return _nan_safe_filter(array, scipy.ndimage.median_filter, **kwargs)
```
rhugonnet, line 114: For the median we need to keep the original implementation that used `np.inf`, which are ignored by `median`.
See the behaviour:
```python
import numpy as np
array = [1, 2, 3]
np.median(array)
Out[4]: np.float64(2.0)
array2 = [1, 2, 3, np.inf, np.inf, np.inf, np.inf]
np.median(array)
Out[6]: np.float64(2.0)
```
I think replacing by zeros in `_nan_safe_filter` won't work:
```python
array = [1, 2, 3, 0, 0, 0, 0]
np.median(array)
Out[8]: np.float64(0.0)
```
Because no matter what you divide the 0 by above, it will never be equal to the real result 2.

rhugonnet: Should we simply call `generic_filter` with `nanmean` and `nanmedian` inside?

rhugonnet (approved):

rhugonnet (changes requested):
Great addition, thanks a lot for all the work! :smile: 

It would be great to add more quantitative tests to ensure the results are as expected with/without NaNs. If I didn't make a mistake in my comment above, the mean/median filters might be erroneous right now.

Maybe we should add a set of test manually checking the operation in a given window, for each filter type, to ensure there is no error?

Example for the mean:
```python
arr = np.random.normal((5, 5))

arr_filtered = _filter(arr, method="mean", size=3)
assert np.nanmean(arr) == arr_filtered[2, 2]
```
They should all be easy to test manually like this, except for the `gaussian` one, that needs to use the logic in the StackOverflow post to correct the division factors.

My last comment: Should we add the `size/radius` parameter in `Raster.filter()`'s parameters directly, to make it more easy to grasp from the user perspective? We also need to specify in the docstring that all other kwargs refer to `scipy.ndimage` filters :wink:.

geoutils/filters.py
adehecq: This would be the simplest solution, but from my experience `generic_filter` is really slow. So we might consider other options in the future. At least, we could compare our implementation to that in the tests.

rhugonnet: Yes that's unfortunately the case, see here (46 times slower to use `generic_filter` for `median`): https://scipy-user.scipy.narkive.com/YyMHNgYg/scipy-ndimage-filters-median-filter-with-nan-values

The other option is to use Numba as default, which should be fast for any custom operation like this.
This would mirror the behaviour of xDEM in terrain attributes, where I'm adding the option of using either a Numba or Scipy engine.

For this PR, we can stick to SciPy only and `generic_filter`.

rhugonnet: There's also a way to write a short C filter to pass to SciPy, which is usually much faster.
See: http://notmatthancock.github.io/2020/04/19/scipy-generic-filter-low-level-callable.html

rhugonnet: Actually my bad, for the `nanmean` we had a working filter already using the trick to replace by the zeros, and then dividing by the sum of valid pixels only, but I don't think this how it behaves exactly in the implementation of this PR.
We actually have an old working duplicate here: https://github.com/GlacioHack/xdem/blob/e1e2065ea52a7db3c8e88045d743aa1f6efd08bb/xdem/spatialstats.py#L2563

In short, to move forward in this PR, we need to remove the `_nan_safe_filter()` function (as this trick only works for the mean, it's not generalizable), re-implement the mean only with this trick.
And, for the median, simply use `generic_filter(np.nanmedian)` (which is quite slow, but there are no other choices I think).

rhugonnet:
I've been working on a very similar problems during the past week in https://github.com/GlacioHack/xdem/pull/486, so while I still know the intricacies of the topic, here's how we could add Numba filters in GeoUtils to dramatically improve the speed for `median` (and we could re-use the same code structure for `nmad`, `nanmax`, and any other `nanfunc`):

```python
from typing import Literal
import numpy as np
from numba import jit, prange
from scipy.ndimage import generic_filter

@jit(parallel=True)
def median_filter_numba(dem: np.ndarray, window_size: int):

    # Get input shapes
    N1, N2 = dem.shape

    # Define ranges to loop through given padding
    row_range = N1 - window_size + 1
    col_range = N2 - window_size + 1

    # Allocate output array
    outputs = np.full((row_range, col_range), fill_value=np.nan, dtype=np.float32)

    # Loop over every pixel concurrently by using prange
    for row in prange(row_range):
        for col in prange(col_range):

            outputs[row, col] = np.nanmedian(dem[row: row+window_size, col:col+window_size])

    return outputs


def median_filter(dem: np.ndarray, window_size: int, engine: Literal["scipy", "numba"]):

    # Using SciPy, options already include edge behaviour
    if engine == "scipy":
        return generic_filter(dem, np.nanmedian, size=window_size, mode="constant", cval=np.nan)
    # Using Numba, need to define edge behaviour ourselves
    elif engine == "numba":
        # We pad the input array with NaNs for half the window size
        hw = int((window_size - 1) / 2)
        dem = np.pad(dem, pad_width=((hw, hw), (hw, hw)), constant_values=np.nan)
        return median_filter_numba(dem, window_size)


############
# TEST SPEED
############

from time import time
import pandas as pd
rnd = np.random.default_rng(42)
nb_it = 10  # Several iterations to be more precise measuring CPU time
window_size = 11
sizes = [50, 100, 500, 1000]  # Array sizes

# Call Numba once in the void to trigger JIT compile
_ = median_filter(np.ones((10, 10)), window_size=3, engine="numba")

list_df = []
for s in sizes:
    dem = rnd.normal(size=(s, s))

    # Get computation times
    t1 = time()
    for i in range(nb_it):
        filt1 = median_filter(dem, window_size=window_size, engine="scipy")
    t2 = time()
    for i in range(nb_it):
        filt2 = median_filter(dem, window_size=window_size, engine="numba")
    t3 = time()

    # Check output is equal
    assert np.allclose(filt1, filt2)

    # Save
    elapsed_scipy = t2 - t1
    elapsed_numba = t3 - t2

    df = pd.DataFrame()
    df = df.assign(elap_numba=[elapsed_numba], elap_scipy=[elapsed_scipy])
    df /= nb_it
    df["s"] = s

    print(f"Size: {s}")
    print(f"Elapsed SciPy: {elapsed_scipy}")
    print(f"Elapsed Numba: {elapsed_numba}")
```

```
Size: 50
Elapsed SciPy: 0.24158883094787598
Elapsed Numba: 0.005011320114135742
Size: 100
Elapsed SciPy: 0.9490830898284912
Elapsed Numba: 0.012121915817260742
Size: 500
Elapsed SciPy: 23.553598880767822
Elapsed Numba: 0.3076012134552002
Size: 1000
Elapsed SciPy: 94.9611337184906
Elapsed Numba: 1.2290620803833008
```

Quite a striking improvement, factor of 100! :wink:

rhugonnet:
Moved to #699

PR #696, opened 2025-04-30
Add input checks for tiling

vschaffn:
- Added error message if the overlap in the tiling is negative/not an integer, added tests.
- Fix error message in `_write_multiproc_result`.


rhugonnet (approved):
Thanks!
The PR title wasn't immediately clear to me, maybe "Add input checks for tiling"? :wink:

PR #699, opened 2025-05-19
Move filter.py from xDEM to GeoUtils and integrate into raster class 

### LINKED ISSUE #690: Move `filter.py` from xDEM to GeoUtils and integrate into raster class

adebardo:
Resolves #690 

I'm picking up @vschaffn's  work on filter management from *geoutils*. I couldn't find any other way than to create a new PR; Valentin's history should be preserved.

I've tried to incorporate the other feedback, with more tests and better handling of NaNs without _nan_safe_filter.

# Previous PR 

## Context
The current filter.py module in xDEM provides a collection of filters that are highly useful for digital elevation models (DEMs). However, these filtering utilities are not intrinsically tied to elevation data—they are generally applicable to any geospatial raster data. Given this, it makes more architectural sense for them to live in GeoUtils.

 ## Changes
- Moved the filter.py module from xdem/ to geoutils/filters.py.
- Removed gaussian_filter_cv to avoid dependence on opencv.
- Moved tests/test_filter.py in xDEM to tests/test_filters.py in GeoUtils.
- Added a new function _nan_safe_filter common for gaussian, mean and median filter, to deal with arrays containing nan values.
- Added a generic _filter function mapping the available filters.
- Added the Raster.filter method that extracts the array with Nans, uses the _filter function, and rebuild the Raster data.
Tests
- Added new tests for the Raster.filter method.

## Documentation
- Added a filter section in raster_class.md.
- 
## Note
When this ticket is approved, a linked PR to remove filters in xDEM will be opened.


rhugonnet:
We could also easily add an NMAD filter based on Numba (just mirroring the median filter code with Numba, with the NMAD calculation instead), which would solve this 4-year old issue! https://github.com/GlacioHack/xdem/issues/69

adebardo:
We could add it to our "project" once created ?

tests/test_filters.py
```diff
@@ -0,0 +1,346 @@
...
+    def test_median_filter_nan_consistency(self) -> None:
+        """Test that different median filter engines return consistent results with NaNs."""
+        arr = np.array([[1, 2, np.nan], [4, np.nan, 6], [7, 8, 9]], dtype=np.float32)
+        filtered_scipy = gu.filters.median_filter(arr, window_size=3, engine="scipy")
+        filtered_numba = gu.filters.median_filter(arr, window_size=3, engine="numba")
+
+        assert filtered_scipy.shape == arr.shape
+        assert filtered_numba.shape == arr.shape
+        mask = np.isfinite(filtered_scipy) & np.isfinite(filtered_numba)
+        np.testing.assert_allclose(filtered_scipy[mask], filtered_numba[mask], rtol=1e-5)
```
rhugonnet: We shouldn't test on the mask but everywhere with `equal_nan=True` here, otherwise they could have different NaNs:
```python
assert np.allclose(filtered_scipy, filtered_numba, equal_nan=True)
```

rhugonnet:
What confusion, can you explain in more details?

environment.yml
```diff
@@ -16,3 +16,4 @@ dependencies:
   - rioxarray=0.*
   - affine
   - shapely
+  - numba
```
rhugonnet: If the dependency is optional (cf the dev-environment), it should be removed from this file

tests/test_filters.py
```diff
@@ -174,14 +244,103 @@ def double_filter(arr: NDArrayNum) -> NDArrayNum:
...
+                        [38.77, 39.21, 41.51, 47.77, 45.75],
+                    ],
+                ),
+                {"sigma": 1},
+            ),
+            (
+                "median",
+                np.array(
+                    [
+                        [39, 42, 52.5, 65.5, 79.5],
```
rhugonnet: I don't think we can verify in our head that those derived array values are correct, so it feels hard to use as a check, no? And we trust NumPy has tests for this (deriving a mean/median/etc), here it's about dimensionality and extra-steps for NaNs.

So, to verify the filtered values, I was thinking more of a simple check against `np.nanfunction` like this:
```python
rng = np.random.default_rng(42)
arr = rng.normal(size=(3, 3))
filtered_arr = gu.filters._filter(array, method="median", window_size=3)
# Central value of the filter should be the same as the filter on the entire array
assert filtered_arr[1, 1] == np.nanmedian(arr)
```
This would work for every filter that is homogeneous in its window (order of values doesn't matter), so all except the Gaussian.

rhugonnet: Actually this rang a bell, we already discussed it here (I had almost forgotten): https://github.com/GlacioHack/geoutils/pull/691#pullrequestreview-2810272027

tests/test_filters.py
```diff
@@ -0,0 +1,294 @@
...
+
+import geoutils as gu
+from geoutils._typing import NDArrayNum
+from geoutils.raster import get_array_and_mask
+
+
+class TestGaussianFilter:
+    """Tests for the Gaussian filter applied to raster data."""
+
+    landsat_dem = gu.Raster(gu.examples.get_path("everest_landsat_b4")).astype(np.float32)
```
adehecq: just a note, this dataset is not a dem, but a Landsat (monoband) image 😉

tests/test_filters.py
```diff
@@ -0,0 +1,294 @@
...
+
+    def test_gauss(self) -> None:
+        """Test Gaussian filter on 2D and 3D arrays, including handling of NaNs and invalid input."""
+        raster_array = get_array_and_mask(self.landsat_dem)[0]
+        raster_sm = gu.filters.gaussian_filter(raster_array, sigma=5)
+        assert np.min(raster_array) <= np.min(raster_sm)
+        assert np.max(raster_array) >= np.max(raster_sm)
+        assert raster_array.shape == raster_sm.shape
+
+        # Test with NaNs
```
adehecq, line 28: should we test with nodata values?

adehecq: Nevermind, the input here is a numpy array, not a geoutils.raster instance.

tests/test_filters.py
```diff
@@ -0,0 +1,294 @@
...
+
+
+class TestDistanceFilter:
+    """Tests for the distance-based outlier filter."""
+
+    landsat_dem = gu.Raster(gu.examples.get_path("everest_landsat_b4")).astype(np.float32)
+
+    def test_dist_filter(self) -> None:
+        """Check that distance filter removes outliers and preserves non-outliers."""
+        ddem = self.landsat_dem.copy()
```
adehecq: why call it "ddem" while it is just an image?

tests/test_filters.py
```diff
@@ -0,0 +1,294 @@
...
+class TestGenericFilter:
+    """Tests for the generic filter API in GeoUtils."""
+
+    landsat_dem = gu.Raster(gu.examples.get_path("everest_landsat_b4")).astype(np.float32)
+
+    def test_generic_filter(self) -> None:
+        """Apply a scipy filter using the generic filter interface."""
+        raster_array = get_array_and_mask(self.landsat_dem)[0]
+        raster_filtered = gu.filters.generic_filter(raster_array, scipy.ndimage.minimum_filter, size=5)
+        assert np.nansum(raster_array) != np.nansum(raster_filtered)
```
adehecq: this is a weak test. Could it be checked against the output of the "min" filter directly?

tests/test_filters.py
```diff
@@ -0,0 +1,294 @@
...
+            ("max", {"size": 3}),
+            ("min", {"size": 3}),
+        ],
+    )
+    def test_raster_filter(self, method: str, kwargs: dict[str, int]) -> None:
+        """Test raster filter methods with standard parameters."""
+        raster = gu.Raster(self.aster_dem_path)
+        filtered_raster = raster.filter(method, inplace=False, **kwargs)
+        assert isinstance(filtered_raster, gu.Raster)
+        assert filtered_raster.shape == raster.shape
```
adehecq: should we check against the output of calling the function directly?

geoutils/filters.py
```diff
@@ -0,0 +1,296 @@
...
+        raise ValueError("Gaussian filter can't be applied to 1D arrays.")
+
+    # Boolean mask: True where NaN
+    mask = np.isnan(array)
+    mask = np.asarray(mask, dtype=bool)
+    # Replace NaNs with 0 for convolution
+    arr_filled = np.where(mask, 0, array)
+
+    # Apply gaussian filter to values and to mask
+    filtered = scipy.ndimage.gaussian_filter(arr_filled, sigma, **kwargs)
```
adehecq, line 110: Unless I miss something, there is an issue here! 
It is not enough to set the nans to 0. The filter must also be applied to the mask array, to propagate the number of valid values. See this [good example](https://stackoverflow.com/questions/18697532/gaussian-filtering-a-image-with-nan-in-python). 
For example, for array [1, 2, nan, 4], the mean is:
- with your approach: (1 + 2 + 4) / 4 = 7/4
- with the correct implementation: (1 + 2 + 4) / 3 = 7/3

geoutils/filters.py
```diff
@@ -0,0 +1,296 @@
...
+
+    :returns: The filtered array (same shape as input).
+    """
+
+    if window_size % 2 == 0:
+        raise ValueError("window_size must be odd")
+
+    if array.ndim == 2:
+        return _apply_median_filter_2d(array, window_size, engine)
+    elif array.ndim == 3:
```
adehecq, line 178: Ok, so the filter is applied only in the "horizontal" dimension. This should be made clear in the description of the filter, as may users may assume that the filter is run in all 3 dimensions.

geoutils/filters.py
```diff
@@ -0,0 +1,296 @@
...
+        return scipy_generic_filter(array, np.nanmedian, size=window_size, mode="constant", cval=np.nan)
+    else:
+        if not _has_numba:
+            raise ValueError("Optional dependency needed. Install 'numba'.")
+        hw = int((window_size - 1) / 2)
+        array = np.pad(array, pad_width=((hw, hw), (hw, hw)), constant_values=np.nan)
+        return median_filter_numba(array, window_size)
+
+
+def mean_filter(array: NDArrayNum, kernel_size: int = 5, **kwargs: Any) -> NDArrayNum:
```
adehecq: is there any reason to set a default value here? I would not make 5 the default but always force the user to set it.

geoutils/filters.py
```diff
@@ -0,0 +1,296 @@
...
+    """
+    Apply a mean filter to a raster that may contain NaNs.
```
adehecq: It should be made clear here (and ideally in the type hinting) that the filter only works for 2D arrays.

geoutils/filters.py
```diff
@@ -0,0 +1,296 @@
...
+    Apply a 2D median filter on an array that may contain NaNs.
+
+    :param array: 2D input array to filter, may contain NaNs.
+    :param window_size: Size of the median filter window (must be odd).
+    :param engine: Filtering engine to use, either "scipy" or "numba".
+    :returns: Filtered array of the same shape as input.
+    """
+
+    if engine == "scipy":
+        return scipy_generic_filter(array, np.nanmedian, size=window_size, mode="constant", cval=np.nan)
```
adehecq: Looking at scipy's documentation, I see that for arrays with Nans, it is suggested to use [vectorized_filter](https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.vectorized_filter.html#scipy.ndimage.vectorized_filter) (see the example with nans near the end). Maybe worth checking the performance for this?

geoutils/filters.py
```diff
@@ -0,0 +1,296 @@
...
+    :param array: The input array to be filtered.
+    :param kernel_size: The size of the kernel.
+
+    :returns: The filtered array (same shape as input).
+    """
+
+    if np.ndim(array) not in [2]:
+        raise ValueError(f"Invalid array shape given: {array.shape}. Expected 2D array.")
+    kernel = np.ones((kernel_size, kernel_size)) / kernel_size**2
+    return generic_filter(array, scipy.ndimage.convolve, **{"weights": kernel, **kwargs})  # type: ignore
```
adehecq: same about using `ndimage.vectorized_filter` here.

geoutils/filters.py
adehecq: Note that this is also different from the scipy's implementation used for example for the min and max filter, which, if I understand correctly, applies the convolution in all dimensions by default.

geoutils/filters.py
```diff
@@ -0,0 +1,296 @@
...
+    # We replace temporarily NaNs by negative infinite values during filtering to avoid spreading NaNs
+    array_nans_replaced = np.where(nans, -np.inf, array)
+    array_nans_replaced_f = scipy.ndimage.maximum_filter(array_nans_replaced, size=size, **kwargs)
+    # In the end we want the filtered array without negative infinite values, so we put back NaNs
+    return np.where(nans, array, array_nans_replaced_f)
+
+
+def distance_filter(array: NDArrayNum, radius: float, outlier_threshold: float) -> NDArrayNum:
+    """
+    Filter out pixels whose value is distantly more than a set threshold from the average value of all neighbor \
```
adehecq: distantly -> distant

geoutils/filters.py
```diff
@@ -0,0 +1,296 @@
...
+pixels within a given radius.
+    Filtered pixels are set to NaN.
+
+    TO DO: Add an option on how the "average" value should be calculated, i.e. using a Gaussian, median etc filter.
```
adehecq: is it planned for this PR? Otherwise, need to add an issue.

adehecq: Also in the meantime, the function used for averaging should be specified in the docstring.

geoutils/filters.py
```diff
@@ -0,0 +1,296 @@
...
+    :param outlier_threshold: the minimum difference abs(array - mean) for a pixel to be considered an outlier.
+
+    :returns: the filtered array (same shape as input)
+    """
+    # Create mask of valid (finite) values
+    valid_mask = np.isfinite(array)
+
+    # Smooth both the data and the valid mask
+    smoothed = gaussian_filter(np.nan_to_num(array, nan=0.0), sigma=radius)
+    normalization = gaussian_filter(valid_mask.astype(float), sigma=radius)
```
adehecq, line 297: Ok, so the fix for taking into account nans is used here. Why not directly in the main `gaussian_filter`?

geoutils/filters.py
adehecq: I just realize the `generic_filter` is a custom function, not scipy's. It would make sense to move the definition of `generic_filter` above that one, so it is defined before it is used.
But I don't understand why this function is needed, and why not directly call ndimage.convolve here...

adehecq: Although the convolution operation is probably fast, can't we use the [uniform_filter](https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.uniform_filter.html#scipy.ndimage.uniform_filter) to calculate the sum, then apply it to the mask to get the number of samples and divide the two to get the mean?

rhugonnet:
Great find on `vectorized_filter` @adehecq! I was curious why we missed out on it before, and the reason is because it's brand new, came out in SciPy 1.16.0 released end-of-July this summer: https://github.com/scipy/scipy/releases/tag/v1.16.0.

This is amazing news, it means we can have performant convolutions/filters with NaNs without requiring Numba, and comforts me into moving Numba into an optional dependency, while having SciPy as default main dependency (both for xDEM and GeoUtils). 

@adebardo If the performance is indeed much better than `generic_filter`, tell us as we need to slightly update the SciPy engine for some terrain attributes in xDEM. :wink:

adebardo:
Hello,
Thank you  @adehecq  for all this feedback, I’ll try to take it into account as quickly as possible.
The use of `vectorized_filter` is very interesting, but this version of SciPy is only compatible with Python 3.11. We need to discuss how to approach this issue.

geoutils/filters.py
adebardo: https://github.com/GlacioHack/geoutils/issues/733

rhugonnet:
> The use of `vectorized_filter` is very interesting, but this version of SciPy is only compatible with Python 3.11. We need to discuss how to approach this issue.

Good catch! For new features, we can do this:
```python
from packaging.version import Version

if Version(scipy.__version__) > Version("1.16.0"):
    generic_filter = scipy.ndimage.vectorized_filter
else:
    generic_filter = scipy.ndimage.generic_filter
```
That should solve it :wink:

rhugonnet (approved):
Perfect! :smile:  
I'll look at fixing the CI this morning.

rhugonnet:
:partying_face:

PR #717, opened 2025-08-01
Add IQR stat  + pytest correction with lazy_fixtures

### LINKED ISSUE #716: Add IQR statistics
## Context

The goal of this ticket is to implement a new metric called the Interquartile Range. `IQR = Q3 - Q1.`
To do this, it needs to be added to all relevant lists in `geoutils`, as statistics are handled at the raster level.

## Code

1. Define an alias: `"IQR"`
   [here](https://github.com/GlacioHack/geoutils/blob/main/geoutils/raster/raster.py#L2014)
2. Add the calculation here (be mindful of the mask):
   [here](https://github.com/GlacioHack/geoutils/blob/main/geoutils/raster/raster.py#L1902)
   Use: `np.percentile(dist, 75) - np.percentile(dist, 25)`
   Exists with scipy : https://docs.scipy.org/doc/scipy-1.16.0/reference/generated/scipy.stats.iqr.html#scipy.stats.iqr
   but do we really want to work with scipy ? 
3. @amaury suggested creating a dedicated `iqr` function and calling it from
   [here:](https://github.com/GlacioHack/geoutils/blob/main/geoutils/stats.py#L28)
   (though this might be overkill for a simpler calculation compared to `nmad`)

## Tests

This new statistic should be tested in as many relevant places as possible.
Example:

* [here](https://github.com/GlacioHack/geoutils/blob/main/tests/test_raster/test_raster.py#L1951)

## Documentation

Remember to update the documentation accordingly:
[here](https://github.com/GlacioHack/geoutils/blob/main/doc/source/raster_class.md)

marinebcht:
This PR proposes to add the IQR (Interquartile Range) to the statistical information (dict) of a raster.

1. SciPy was choosen over Numpy :

- IQR function is already implemented in scipy.stats ([here](https://docs.scipy.org/doc/scipy-1.16.0/reference/generated/scipy.stats.iqr.html))
- This function is faster than recompute it with Numpy 

 ```
# Raster exploradores_aster_dem
Numpy) 0.011367651999535155 sec
iqr = 530.6628723144531
Scipy) 0.0081036680003308 sec
iqr = 530.6628723144531 

# List 1M values
Numpy) 0.14357674500024586 sec
iqr = 499999.5
Scipy) 0.07437301699974341 sec
iqr = 499999.5 
```
2. The computation no need to be outsourced in another fonction (like others np metrics or rmse at worst).

By the way, it replaces the obsolete pytest-lazyfixture with the new (working with Pytest > 8.0.0 and maintained) variant pytest-lazy-fixtures.

 - [x] Resolves #716 
 - [x] Tests: tests/test_raster/test_multiraster.py passes now + iqr test (with scipy) and the presence of the new stat in test_raster.py:test_stats
 - [x] Docs: add IQR to the stats list in raster_class.md
 - [x] Deps: pytest-lazyfixture replaced by pytest-lazyfixtures in `dev-environment.yml` and `setup.cfg`

### DIFF
--- a/doc/source/raster_class.md
+++ b/doc/source/raster_class.md
@@ -404,7 +404,8 @@ Supported statistics are :
 - **Sum:** sum of all data, ignoring masked values.
 - **Sum of squares:** sum of the squares of all data, ignoring masked values.
 - **90th percentile:** point below which 90% of the data falls, ignoring masked values.
-- **LE90 (Linear Error with 90% confidence):** Difference between the 95th and 5th percentiles of a dataset, representing the range within which 90% of the data points lie. Ignore masked values.
+- **IQR (Interquartile Range):** difference between the 75th and 25th percentile of a dataset.
+- **LE90 (Linear Error with 90% confidence):** difference between the 95th and 5th percentiles of a dataset, representing the range within which 90% of the data points lie. Ignore masked values.
 - **NMAD (Normalized Median Absolute Deviation):** robust measure of variability in the data, less sensitive to outliers compared to standard deviation. Ignore masked values.
 - **RMSE (Root Mean Square Error):** commonly used to express the magnitude of errors or variability and can give insight into the spread of the data. Only relevant when the raster represents a difference of two objects. Ignore masked values.
 - **Std (Standard deviation):** measures the spread or dispersion of the data around the mean, ignoring masked values.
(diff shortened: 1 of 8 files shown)


doc/source/raster_class.md
belletva, line 1: For the LE90, the NMAD, the RMSE, etc., we have the information that the masked values are ignored. Could you add this information also for the IQR?

adebardo:
Coveralls seems to be crazy
image

PR #729, opened 2025-09-16
Bug: initialise Raster with user-provided nodata value

### LINKED ISSUE #727: `Raster('filename.tif', nodata=value)` not respected
**Describe the bug**
When loading an image from disk, a manually provided nodata value is ignored - it is overwritten by the image's metadata nodata value.

**To Reproduce**
```python
# The ASTER DEM example contains nodata values of -9999
filename_rast = gu.examples.get_path("exploradores_aster_dem")
# Try to overwrite them with value=0
rast = gu.Raster(filename_rast, nodata=0)
# Examine the result - also useful to compare to the default loading of this image.
# Given the nodata values of -9999 we would expect to see them rendered with cmap minima, but they are still rendered in white, i.e. no data.
rast.plot()
```

**Expected behavior**
The user-supplied no-data value should be respected, but seems instead to be overwritten.

https://github.com/GlacioHack/geoutils/blob/d1790b6c78aa0f2c304e66dc6d77edafe2458f96/geoutils/raster/raster.py#L449-L471

atedstone:
Resolves #727 .

In addition to the test that `self._nodata` gets set, I also visually verified that the masking with the user-provided nodata value works as expected:

```python
im = gu.Raster(gu.examples.get_path('everest_landsat_b4'), nodata=255)
im.plot()
```
All high mountain tops (e.g., snow-covered) get masked.

### DIFF
--- a/geoutils/raster/raster.py
+++ b/geoutils/raster/raster.py
@@ -468,7 +468,11 @@ class Raster:
 
                 self._transform = ds.transform
                 self._crs = ds.crs
-                self._nodata = ds.nodata
+                # Allow user to manually override the nodata value which may be specified in the file.
+                if nodata != None:
+                    self._nodata = nodata
+                else:
+                    self._nodata = ds.nodata
                 self._name = ds.name
                 self._driver = ds.driver
                 self._tags.update(ds.tags())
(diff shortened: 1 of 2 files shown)


rhugonnet (approved):

rhugonnet:
Perfect, thanks @atedstone!

PR #737, opened 2025-10-04
Remove `RasterMask` and `PointCloudMask` classes in favor of boolean dtype in parent class

### LINKED ISSUE #387: Create a parent class `RasterBase` for `Mask` and `Raster`?
Need to think about this a bit, but it'd probably make things more consistent. Maybe also the occasion to bring the core functionalities in one place before writing #383, to test things? 

Right now, `Mask` is a subclass of `Raster`. However, as currently defined, `Raster.data` cannot be a `bool`... So `mypy`'s typing error are ignored for a few functions.
Maybe it would make more sense to have a parent class `RasterBase`, that allows all data types and only contains the core functionalities?

### LINKED ISSUE #706: Potential refactoring of `Mask` classes ahead of accessors
Important discussion on whether to keep the `Mask` class casting, or remove the separate `Mask` classes.

Since #492, we now have two mask classes: `RasterMask` and `PointCloudMask`. This class separation is practical, but do we really need it?

**The problem:** The casting from `Raster` to `RasterMask` implies some method overloading which is unpractical. It's not always the easiest to maintain through inheritance (e.g. https://github.com/GlacioHack/xdem/issues/739). Finally, it would imply requiring separate Xarray/Pandas accessors for these classes.

**Another route:**
Data objects describing rasters (`np.ma.MaskedArray` or `xr.DataArray`) or point clouds (`gpd.GeoDataFrame` support boolean data types in their underlying NumPy arrays.
We could technically simply have `RasterMask` be a `Raster` and `PointCloudMask` a `PointCloud` with an extra attribute `is_bool_mask` checking the data type, which would trigger a slightly different behaviour for some functions like the current `Mask` classes do (saving, loading, plotting, indexing/index assigment, arithmetic, rasterizing, polygonizing, proximity, etc). This might be sufficient to circumvent certain difficulties linked to masks (boolean type not supported on disk or during some operation which requires some conversions).

rhugonnet:
This PR removes the `RasterMask` (very recently `Mask`) and `PointCloudMask` classes, in favor of a boolean `Raster` and `PointCloud` with an attribute `is_mask` to rapidly check if the dtype is indeed boolean.

All mask-specific methods and behaviours were moved directly in `Raster` and `PointCloud`, and documentations merged. As a raster cannot be saved as a boolean on-file, an argument `is_mask` was added to `Raster.__init__()` for users to specific that the files needs to be considered as a mask.

The old `Mask` class is on deprecation until GeoUtils 0.3, and calls `Raster(is_mask=True)`.

Resolves #706
Resolves #387


doc/source/raster_class.md
```diff
@@ -37,6 +37,11 @@ A first category includes georeferencing attributes directly derived from {attr}
 A second category concerns the attributes derived from the raster array shape and type: {attr}`~geoutils.Raster.count`, {attr}`~geoutils.Raster.bands` and
```
belletva, line 37: I think it can be a good idea to add code just after the definition of the four main attributes (such as in file core_composition.md)

doc/source/raster_class.md
```diff
@@ -412,3 +416,144 @@ pc_sub = rast.subsample(500)
...
+
+A raster mask is a boolean {class}`~geoutils.Raster` (True or False).
+
+While boolean data types are typically not supported in raster filetypes or in-memory operations, they are incredibly useful for various logical and
+arithmetical operation in geospatial analysis, so GeoUtils facilitates their manipulation to support these operations natively and implicitly.
+
+```{note}
+There is no {class}`~geoutils.Raster.nodata` value defined in a boolean {class}`~geoutils.Raster`, as it only take binary values. However, the
+{class}`numpy.ma.MaskedArray` still has a {class}`~geoutils.Raster.data.mask` for invalid values.
+```
```
belletva: I think it can be very useful to add code cell with an example

doc/source/raster_class.md
belletva, line 37: It is not very clear for me the first category and the second category, I suggest to add a code cell with some examples.

doc/source/api.md
```diff
@@ -198,7 +199,7 @@ documentation.
 
 And reverse operations.
 
-### Logical operator casting to Mask
+### Logical operators casting to boolean
```
belletva: a boolean raster?

doc/source/core_py_ops.md
```diff
@@ -69,14 +69,13 @@ If an unmasked {class}`~numpy.ndarray` is passed, it will internally be cast int
...
-## Logical comparisons cast to {class}`~geoutils.RasterMask`
+## Logical comparisons cast to a raster mask
 
 Logical comparison operators ({func}`==<operator.eq>`, {func}` != <operator.ne>`, {func}`>=<operator.ge>`, {func}`><operator.gt>`, {func}`<=<operator.le>`,
 {func}`<<operator.lt>`) can be used on a {class}`~geoutils.Raster`, also in combination with any other {class}`~geoutils.Raster`, {class}`~numpy.ndarray` or
 number.
 
-Those operation always return a {class}`~geoutils.RasterMask`, a subclass of {class}`~geoutils.Raster` with a boolean {class}`~numpy.ma.MaskedArray`
-as {class}`~geoutils.Raster.data`.
+Those operation always return a raster mask i.e. a {class}`~geoutils.Raster` with a boolean {class}`~numpy.ma.MaskedArray` as {class}`~geoutils.Raster.data`.
```
belletva, line 78: A suggestion: add `rast.is_mask` to show that the raster is a MaskedArray before and add `mask.is_mask` to show that it is still a raster mask after the operation?

doc/source/core_py_ops.md
```diff
@@ -85,14 +84,14 @@ mask
...
 ```
 
-## Logical bitwise operations on {class}`~geoutils.RasterMask`
+## Logical bitwise operations on raster masks
 
 Logical bitwise operators ({func}`~ <operator.invert>`, {func}`& <operator.and_>`, {func}`| <operator.or_>`, {func}`^ <operator.xor>`) can be used to
-combine a {class}`~geoutils.RasterMask` with another {class}`~geoutils.RasterMask`, and always output a {class}`~geoutils.RasterMask`.
+combine a boolean {class}`~geoutils.Raster` with another boolean {class}`~geoutils.Raster`, and always output a boolean {class}`~geoutils.Raster`.
 
 ```{code-cell} ipython3
```
belletva, line 96: Same thing here for `rast` and `mask`, I suggest to show `is_mask`

doc/source/pointcloud_class.md
belletva, line 1: Additionally, new attributes such as {attr}`~geoutils.PointCloud.point_count` and --> the end of the sentence is missing

doc/source/pointcloud_class.md
```diff
@@ -115,7 +115,7 @@ The operation is applied to the {attr}`~geoutils.PointCloud.data_column` of the
 
 A {class}`~geoutils.PointCloud` can also be applied any pythonic logical comparison operation ({func}`==<operator.eq>`, {func}` != <operator.ne>`,
 {func}`>=<operator.ge>`, {func}`><operator.gt>`, {func}`<=<operator.le>`, {func}`<<operator.lt>`) with another {class}`~geoutils.PointCloud`,
-{class}`~numpy.ndarray` or number. It will cast to a {class}`~geoutils.PointCloudMask`.
+{class}`~numpy.ndarray` or number. It will cast to a boolean {class}`~geoutils.PointCloud`.
```
belletva, line 116: Maybe add in the example : `pc1.is_mask` to show that is really cast to a boolean Point Cloud

atedstone:
a small query on the proposed naming: did we also contemplate `is_bool`? Given that the updated documentation now seems to use the term 'boolean' a lot, I wonder if we should change the flag semantics to match?

belletva:
> a small query on the proposed naming: did we also contemplate `is_bool`? Given that the updated documentation now seems to use the term 'boolean' a lot, I wonder if we should change the flag semantics to match?

It seems a very idea the `is_bool`!

rhugonnet:
Thanks both. After some thought about it when I took on this PR, I was not so convinced by `is_bool`, for several reasons:
- **Wrong target:** "Boolean" describes the data type of the array of the raster. While "Mask" describes directly a boolean array/raster, so the object itself. So it'd have to be `Raster.is_dtype_bool`, not `Raster.is_mask`.
- **Continuity and consistency with other packages:** We've talked a lot about georeferenced "masks" before, and we used this term with reason because it's a widely use term (cf NumPy masked array, for example, or SciPy masked stats). We just need to avoid confusion with the nodata mask (`Raster.data.mask`), but I think that's not too hard to do in the doc.
- **Complexity of "boolean":** I remember that one of the main comments I got before we released the 0.1 (and built most of the documentation) was to remove most occurrences of "boolean" in the documentation (maybe it was you @atedstone?). Because it's quite a technical term, that is not very intuitive to many users. While mask is more easy to understand right away. Right now we're "deep in the forest" so we don't realize that the term can be difficult anymore, but I agreed at the time that it was better to remove it. This is also why I used "Mask" most often in this PR's documentation edits, and only sometimes specified, "The mask, i.e. a boolean Raster (True or False values)" when it wasn't defined yet on the page, for full clarity.

rhugonnet:
@adehecq @adebardo Thoughts?

adehecq:
Seems good to me. I don't have a very strong opinion on whether to use `is_mask` or `is_bool`. Both can be a little bit ambiguous for the user, but as long as it's clear in the documentation, it should be fine. Just scrolled the doc quickly and it seemed ok.
So I would stick to `is_mask` if that is your preference and because it remains close to the original Mask class.

doc/source/core_py_ops.md
rhugonnet: The boolean masked array renders in the documentation with True/False values, so I don't think we need to add `is_mask` here!

doc/source/core_py_ops.md
rhugonnet: Same as above

doc/source/pointcloud_class.md
rhugonnet: Good catch!

doc/source/pointcloud_class.md
rhugonnet: Same as for rasters: the point cloud is printed in the doc with values "True"/"False", so I think this is implicit, we don't need to show `is_mask`.

doc/source/raster_class.md
rhugonnet: Would have to re-shuffle the page, because no file has been opened yet (this comes after).

We need to update the doc soon, there's old pages to update. Keeping this in mind for later.

rhugonnet: Same as above.

doc/source/raster_class.md
rhugonnet: Actually this is outdated, because we don't override the nodata anymore. Replaced with the "important" note below.

Issue #713, opened 2025-07-17
Remove conda test for deps


adebardo:
# Context

The purpose of this ticket would be to implement a pip-only release. To do this, we need to remove or move the tests dependent on conda.

@rhugonnet  I am pointing you to the new ticket; if you could explain it to me in more detail, that would be great.

rhugonnet:
Ah and, because we'll be testing the "import geoutils" through pip in `python-tests.yml`, and then running the tests directly after, we won't need `.github/workflows/pip-checks.yml` anymore.

PR #738, opened 2025-10-04
Improve test speed and structure

### LINKED ISSUE #442: test_doc.py fails on Windows
`test_doc.py` fails on Windows during the reading of the CLI.md file. See failed test [here](https://github.com/GlacioHack/geoutils/actions/runs/7572571733/job/20622721379) during PR #436.
This seems to be related to the fact that the script `geoviewer.py` is not found in the path.

### LINKED ISSUE #731: Improve tests for missing optional dependencies
Opening after https://github.com/GlacioHack/xdem/pull/775 and https://github.com/GlacioHack/geoutils/pull/699#issuecomment-3314374976

Linked to updating CI, so this issue: https://github.com/GlacioHack/geoutils/issues/713

Right now, in CI, we first install the base `environment.yml`, check that the package imports, then update to the `dev-environment.yml` with all optional dependencies included, then run all tests.
This worked OK when we didn't have any transverse optional dependencies that can affect multiple function during runtime, but only optional dependencies specific to certain features.

Now, with the addition of the profiling module that is both optional and transverse to many functions (https://github.com/GlacioHack/xdem/pull/775), as well as running through Numba for various features (only filters for now, but IDW interpolation soon as well), optional dependencies could potentially affect a lot of functions.
And we need to be 100% sure that those run properly without optional dependencies installed.

@adebardo @belletva Do you have some practice on this from other work?
I did some reading on best practices, and it seems that updating the CI environment (like we currently do) is too time-consuming (and too complex to run for tests, you'd have to run tests specific to an environment; or run duplicate tests as many times as you have environments). 
Instead, it seems that people use strategies to "mock missing dependencies", where you make the system believe that a module is not installed to check that code indeed runs without it. One can do this manually using directly `sys.modules` (e.g., https://stackoverflow.com/questions/37916040/how-to-write-unittests-for-an-optional-dependency-in-a-python-package), or we could use this minimal pytest extension: https://pytest-missing-modules.eertmans.be/latest/.

### LINKED ISSUE #739: Test collection significantly slowed down by online file for vector tests
Since recently I found the GeoUtils tests very long to collect (before the tests run, which will run almost every time... So really need to be kept short). Normally, tests are collected in a few seconds. Here they take more than a minute, sometimes more!

After a lot of digging, this comes from the online vector file here: https://github.com/GlacioHack/geoutils/blob/260dde060fa508577bc192d1fe1207d1cb811452/tests/test_vector/test_vector.py#L23

It's a bad practice to have any network dependency during test collection (that is not version-controlled, like our GitHub repo for test files), there's actually a Pytest extension for this: [pytest-socket](https://pypi.org/project/pytest-socket/)

I'll try to address this in #738

### LINKED ISSUE #740: Test of `Vector.freschet_distance()` extremely slow
Since the latest GeoPandas/Shapely.
We can simply reduce the size of real test data, should solve it.

rhugonnet:
This PR improves GeoUtils's CI test speed by reducing the size of test files and optimizing several tests, by moving profiling tests (RAM) in a separate benchmark/ folder to be run manually, and by separating the documentation build in CI.

Resolves #731
Resolves #442
Resolves #739
Resolves #740

Here's an awesome list of tricks to improve speed of Pytest: https://github.com/zupo/awesome-pytest-speedup

TO-DO:
- [x] Move documentation build into its own GitHub Actions run,
- [x] Move profiling tests in a new benchmark/ folder, with a GitHub Actions workflow that can be dispatched manually,
- [x] Make test files smaller to naturally improve tests relying on example files (multiraster, reprojection, etc),
- [x] Improve speed of test collection (why is it so slow???): Issue was #739
- [x] Improve speed of vector tests: Issue was #740 

**Note on Dask (that takes the most time per individual test now, might become limiting in future):** 
All `test_dask` tests take minimum 0.2s each, despite using tiny data. I attributed this to overhead (setting up cluster, tearing down), and so thought it could be optimized. Over a lot of loops (hundreds), this starts to be a lot of time.
I tried to optimize this by using a cluster fixture shared for all tests (cf Dask issues and doc, as recommended), or by removing the `pytest.mark.parametrize()` (that Dask doesn't like, restarts the cluster even if it's a fixture) by moving the parameters within simple `for` loops. I dug quite a bit in various issues/doc. None of what I tried improved test speed, it actually made it worse (by a factor 4). 
So I kept our original setup... Not sure if it can really be improved.


rhugonnet:
@adebardo @belletva @adehecq This PR shows (through the new documentation build GitHub workflows that fails on Pip trying to install GDAL + Rasterio, as we use GDAL in tests) that we cannot yet remove Conda in the tests (#713).

We would have to do the same as in xDEM, and move the GDAL scripts to `geoutils-data`, to ensure Rasterio can build in tests.

adehecq:
Nice job!!
Regarding the test data, I would also save them in the geoutils-data repo so they don't change too regularly. 
In practice, how would they be distinguished from the doc data? Should we add a "test_" prefix to all the subsampled files?
Would it be worth keeping a few "iconic" tests on the larger dataset, just to make sure the documentation examples run without bug? Remember when we had xDEM documentation on Nuth & Kaab showing an awfully coregistered DEM 😆

Issue #693, opened 2025-04-29
Optimize performance of `Raster.get_stats()`


rhugonnet:
While working on https://github.com/GlacioHack/geoutils/pull/492, I noticed a couple issues with `Raster.get_stats()`:
1. The function always computes all statistics even when asked for a single one, here: https://github.com/GlacioHack/geoutils/blob/8a88c8467c53e3f280fa0b18a3558c192f9fc215/geoutils/raster/raster.py#L2009 This makes it ~15 times slower on CPU when asking for a single stat.
2. The function duplicates the data in RAM here: https://github.com/GlacioHack/geoutils/blob/8a88c8467c53e3f280fa0b18a3558c192f9fc215/geoutils/raster/raster.py#L1918  Normally we can skip this by calling function that work on masked arrays directly. For `np.ma.percentile` that doesn't exist in NumPy, we can call the SciPy module for masked array instead, using `mquantiles`: https://docs.scipy.org/doc/scipy/reference/stats.mstats.html#statistical-functions-for-masked-arrays-scipy-stats-mstats

I've had to re-structure these functions outside of the `Raster` class inside a `stats/` module to re-use them on `PointCloud` objects, so better to wait until that PR is merged to tackle these changes.

PR #747, opened 2025-10-14
Optimization of get_stats()

marinebcht:
Resolves https://github.com/GlacioHack/geoutils/issues/693

PR to reduce time cost for the get_stats() functions that compute all stats event if on is asked.

### Resolution

Statistics are only calculated if we want them through a dictionary full of [partial functions].(https://docs.python.org/fr/3.9/library/functools.html) (the calculation is initiated by dict[stat] call).

In case of a **inler_mask**, all these stats are calculated as their are not more complexe than a simple division. 
In case of **one stats_name given**, _statistics is executed with (stats_name=[stats_name]).    
In case of **stats_name is a callable**, get_stats returns directly (after application of the inler_mask if it exists) stats_name(data).   

Several optimizations were performed to calculate all the necessary values ​​only once.
No statistics are calculated if the data is empty.

### Tests  
- tests_get_stats() =>  tests_get_stats_raster() 
- tests_get_stats_pontcloud() to verify all the outputs stats presence for each situation (like in tests_get_stats)
-  test_raster_get_stats_values() : non reg with values
-  test_pointcloud_get_stats_values() : non reg with values

### DIFF
--- a/dev-environment.yml
+++ b/dev-environment.yml
@@ -6,7 +6,7 @@ dependencies:
   - geopandas>=0.12.0
   - matplotlib=3.*
   - pyproj=3.*
-  - rasterio=1.4.3
+  - rasterio>=1.3,<1.4.4
   - pandas>=1,<3
   - numpy>=1,<3
   - scipy=1.*
--- a/environment.yml
+++ b/environment.yml
@@ -6,7 +6,7 @@ dependencies:
   - geopandas>=0.12.0
   - matplotlib=3.*
   - pyproj=3.*
-  - rasterio=1.4.3
+  - rasterio>=1.3,<1.4.4
   - pandas>=1,<3
   - numpy>=1,<3
   - scipy=1.*
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,7 @@
 geopandas>=0.12.0
 matplotlib==3.*
 pyproj==3.*
-rasterio==1.4.3
+rasterio>=1.3,<1.4.4
 pandas>=1,<3
 numpy>=1,<3
 scipy==1.*


geoutils/raster/raster.py
```diff
@@ -2028,37 +2028,24 @@ def get_stats(
...
-        if counts is None:
-            ignore_aliases = [
-                "validinliercount",
-                "totalinliercount",
-                "percentagevalidinlierpoints",
-                "percentageinlierpoints",
-            ]
-            stats_aliases = {k: _STATS_ALIASES[k] for k in _STATS_ALIASES.keys() if k not in ignore_aliases}
+        # Pre-computing depending on nature of array
+        # TODO: Array is duplicated into filled array with NaN at every call, doubling memory usage
```
adebardo: It's better to open issues than adding TODO in code

geoutils/stats/estimators.py
```diff
@@ -69,3 +69,46 @@ def linear_error(data: NDArrayNum, interval: float = 90) -> np.floating[Any]:
         mdata = data
     le = np.nanpercentile(mdata, 50 + interval / 2) - np.nanpercentile(mdata, 50 - interval / 2)
     return le
+
```
adebardo, line 73: I don't think this is the place for such comment x)

geoutils/stats/stats.py
```diff
@@ -48,93 +50,134 @@
...
+    "Percentage valid inlier points",
+]
+
+
+def _statistics(
+    data: NDArrayNum,
+    stats_name: list[str | Callable[[NDArrayNum], np.floating[Any]]] | None = None,
+    counts: tuple[int, int] | None = None,
+    mask_count_nonzero: int = None,
+) -> dict[str, float]:
```
adebardo, line 106: can you add a docstring ?

geoutils/stats/stats.py
```diff
@@ -145,35 +155,34 @@ def _statistics(
...
-            "Percentage valid inlier points",
-        ]
-
         # If inlier mask was passed
-        if counts is not None and (stats_name is None or list(set(list_counts_stats).intersection(stats_name))):
-            valid_inlier_count = np.count_nonzero(mask)
+        if counts is not None and (stats_name is None or list(set(STATS_LIST_MASK).intersection(stats_name))):
             dict_c = {
-                "Valid inlier count": valid_inlier_count,
+                "Valid inlier count": mask_count_nonzero,
```
adebardo: Can you please resume why you changes this name ? We had lot of trouble finding it the first time haha x)

marinebcht:
@rhugonnet 

Before : 

```
def get_stats(...):
    stats_dict = _statistics(.....)
    if stats_name is None:
        return stats_dict


def _statistics(...): 

    # SAME PART
    if np.ma.isMaskedArray(data):
        mask = ~np.ma.getmaskarray(data)
        mdata = np.ma.filled(data.astype(float), np.nan)
    else:
        mask = np.isfinite(data)
        mdata = data
    # Valid count
    valid_count = np.count_nonzero(mask) if counts is None else counts[0]

    # DIFF PART 
     stats_dict = {
        "Mean": np.ma.mean(data),
        "Median": np.ma.median(data),
        "Max": np.ma.max(data),
        "Min": np.ma.min(data),
        "Sum": np.ma.sum(data),
        "Sum of squares": np.ma.sum(np.square(data)),
        "90th percentile": np.nanpercentile(mdata, 90),
        "LE90": linear_error(mdata, interval=90),
        "IQR": iqr(mdata, nan_policy="omit"),  # ignore masked value (nan),
        "NMAD": nmad(data),
        "RMSE": np.sqrt(np.ma.mean(np.square(data))),
        "Standard deviation": np.ma.std(data),
        "Valid count": valid_count,
        "Total count": data.size,
        "Percentage valid points": (valid_count / data.size) * 100,
      }

      + 2x if 
```


Now :

```
def get_stats(...):
    if stats_name is None or  ....:
        return _statistics(...)


def _statistics(...): 

    # SAME PART
    if np.ma.isMaskedArray(data):
        mask = ~np.ma.getmaskarray(data)
        mdata = np.ma.filled(data.astype(float), np.nan)
    else:
        mask = np.isfinite(data)
        mdata = data

   # DIFF PART
   if non vide : 
     valid_count = final_count_nonzero if counts is None else counts[0]
        stats_dict = {
            "Mean": partial(np.ma.mean, data),
            "Median": partial(np.ma.median, data),
            "Max": partial(np.ma.max, data),
            "Min": partial(np.ma.min, data),
            "Sum": partial(np.ma.sum, data),
            "Sum of squares": partial(lambda x: np.ma.sum(np.square(x)), data),
            "90th percentile": partial(lambda x: np.nanpercentile(x, 90), mdata),
            "LE90": partial(linear_error, mdata, interval=90),
            "IQR": partial(iqr, mdata, nan_policy="omit"),  # ignore masked value (nan),
            "NMAD": partial(nmad, data),
            "RMSE": partial(lambda x: np.sqrt(np.ma.mean(np.square(x))), data),
            "Standard deviation": partial(np.ma.std, data),
            "Valid count": partial(lambda x: x, valid_count),
            "Total count": partial(lambda x: x.size, data),
            "Percentage valid points": partial(lambda x: (valid_count / x.size) * 100, data),
        }

        res_dict = {}
        if stats_name is None:
            res_dict = {k: stats_dict[k]() for k in stats_dict.keys()}  # I had in excess before

   + 1x if 
```

`partial` applies function `argv[0]` with` argv[1:]` arguments juste when the `stats_dict[key]` is called 


But the fact that the time difference needed increases with the size on the image, I don't really get it

marinebcht:
As the mixed solutions was chosen, here are the compute stats difference if data is a masked array of not : 

| np.ma.isMaskedArray(data)  | True          | False |
| :--------------- |:---------------:| -----:|
| mean, median, min, max, std, sum  |  np.ma.XX       |   np.nanXX |
| 90th percentile  | partial(lambda x: mquantiles(x, prob=0.9)[0]) | partial(np.nanpercentile, q=90)   |
| rmse | np.sqrt(np.ma.mean(np.square(data)))     |  np.sqrt(np.nanmean(np.square(data))) |
| sum square  |  np.ma.sum(np.square(data)) | np.nanma.sum(np.square(data)) |
| IQR |   partial(iqr, nan_policy="omit") |   partial(iqr, nan_policy="omit") | 
| LE90 | | 
| NMAD | nfact * np.ma.median(np.abs(data - np.ma.median(data))) |  nfact * np.nanmedian(np.abs(data - np.nanmedian(data))) |
| valid count if counts is None | np.count_nonzero(~np.ma.getmaskarray(data)) | np.count_nonzero(np.isfinite(data)) |
| total count | data.size | data.size | 
| valid points % | (valid_count / data.size)*100 if data.size else nan |  (valid_count / data.size)*100 if data.size else nan |

marinebcht:
I change several computations in case of a nodata/masked input data array. So these stats are now always computed : 
- Valid count
- Total count
- Percentage valid points
- Valid inlier count (if input inlier mask)
- Total inlier count (if input inlier mask)
- Percentage inlier points (if input inlier mask)
- Percentage valid inlier points (if input inlier mask)

Plus, I added a warning log in case of a non string stat name

rhugonnet:
Perfect! :smile:

Issue #769, opened 2025-11-19
Reprojection error in case of a "Mask"


marinebcht:
A new error was introduced when we need to reproject a "Mask" into a DEM that don't fully intersect themself  : 

```
import geoutils as gu
from geoutils import examples

filename_rast = gu.examples.get_path("everest_landsat_b4")
reference_dem = gu.Raster(filename_rast)
filename_vect = gu.examples.get_path("everest_rgi_outlines")
vect = gu.Vector(filename_vect)
inlier_mask = ~vect.create_mask(reference_dem)

# dem to reproject 
reference_dem_crop = reference_dem.icrop((0, 0, 88, 88))
inlier_mask_crop = inlier_mask.icrop((0, 0, 88, 88))

# ref
reference_dem = reference_dem.icrop((100, 100, 188, 188))
```

No intersection between reference_dem_crop/reference_dem and inlier_mask_crop/reference_dem.

#### `reference_dem_2.reproject(reference_dem, resampling="nearest")` : OK

```
res = [[-- -- -- ... -- -- --] .... ]
stats_res = {'Mean': nan, 'Median': nan, 'Max': nan, 'Min': nan, 'Sum': nan, 'Sum of squares': nan, '90th percentile': nan, 'LE90': nan, 'IQR': nan, 'NMAD': nan, 'RMSE': nan, 'Standard deviation': nan, 'Valid count': nan, 'Total count': nan, 'Percentage valid points': nan}
```

#### `inlier_mask_crop.reproject(reference_dem, resampling="nearest")` : OK

```
res = [[-- -- -- ... -- -- --] .... ]
stats_res = {'Mean': nan, 'Median': nan, 'Max': nan, 'Min': nan, 'Sum': nan, 'Sum of squares': nan, '90th percentile': nan, 'LE90': nan, 'IQR': nan, 'NMAD': nan, 'RMSE': nan, 'Standard deviation': nan, 'Valid count': nan, 'Total count': nan, 'Percentage valid points': nan}
```

#### `reference_dem_2.reproject(reference_dem, resampling="nearest")` : OK
 
```
res = [[-- -- -- ... -- -- --] .... ]
[...]

PR #770, opened 2025-11-19
Reprojection error in case of a "Mask" 

marinebcht:
Resolves https://github.com/GlacioHack/geoutils/issues/769 

A new error was introduced when we need to reproject a "Mask" into a DEM that don't fully intersect themself  : 

```
import geoutils as gu
from geoutils import examples

filename_rast = gu.examples.get_path("everest_landsat_b4")
reference_dem = gu.Raster(filename_rast)
filename_vect = gu.examples.get_path("everest_rgi_outlines")
vect = gu.Vector(filename_vect)
inlier_mask = ~vect.create_mask(reference_dem)

# dem to reproject 
reference_dem_crop = reference_dem.icrop((0, 0, 88, 88))
inlier_mask_crop = inlier_mask.icrop((0, 0, 88, 88))

# ref
reference_dem = reference_dem.icrop((100, 100, 188, 188))
```

No intersection between reference_dem_crop/reference_dem and inlier_mask_crop/reference_dem.

### geoutils 0.1.17 `inlier_mask_crop.reproject(reference_dem, resampling="nearest")` : OK

```
res = [[-- -- -- ... -- -- --] .... ]
stats_res = {'Mean': nan, 'Median': nan, 'Max': nan, 'Min': nan, 'Sum': nan, 'Sum of squares': nan, '90th percentile': nan, 'LE90': nan, 'IQR': nan, 'NMAD': nan, 'RMSE': nan, 'Standard deviation': nan, 'Valid count': nan, 'Total count': nan, 'Percentage valid points': nan}
```

### geoutils 0.2.0 `inlier_mask_crop.reproject(reference_dem, resampling="nearest")` : KO

```
res = [[True True True ... True True True] .... ]
stats_res = {'Mean': np.float64(1.0), 'Median': np.float64(1.0), 'Max': np.True_, 'Min': np.True_, 'Sum': np.int64(7744), 'Sum of squares': np.int64(7744), '90th percentile': np.float64(1.0), 'LE90': np.float64(0.0), 'IQR': np.float64(0.0), 'NMAD': np.float64(0.0), 'RMSE': np.float64(1.0), 'Standard deviation': np.float64(0.0), 'Valid count': np.int64(7744), 'Total count': 7744, 'Percentage valid points': np.float64(100.0)}
```

# 
In case of a boolean raster to reproject, `src_arr.dtype` is bool but the projection process will work on uint8 (source and dest) in case of a resampling == "nearest", and float32 otherwise.

**Exemple  resampling == "nearest"  :** 

| Type | SRC boolean  | SRC uint8  | / |  DST uint8 | DST boolean |
| :--------------- | :--------------- |:---------------:| -----:| -----:| -----:|
| True |  True  | 1  | / |  1 | True |
| False | False | 0   | / |  0 |  False |
| No data | TODO | TODO | /  | 255  | **True** |

### DIFF
--- a/tests/test_raster/test_geotransformations.py
+++ b/tests/test_raster/test_geotransformations.py
@@ -797,14 +797,39 @@ class TestMaskGeotransformations:
         with pytest.warns(UserWarning, match="Reprojecting a raster mask .*"):
             mask.reproject(res=50, resampling="bilinear", force_source_nodata=2)
 
-        # Test 3: No intersection lead to a full nan mask
-        raster_boolean = gu.Raster.from_array(
-            np.random.randint(2, size=(5, 5), dtype=bool), transform=rio.transform.from_origin(0, 5, 1, 1), crs=4326
-        )
-        ref_dem = gu.Raster.from_array(
-            np.random.randint(100, size=(5, 5), dtype="uint8"),
-            transform=rio.transform.from_origin(10, 20, 1, 1),
+        # Test 3: Nan Handling
+
+        dem_bool = gu.Raster.from_array(
+            np.random.randint(2, size=(100, 100), dtype=bool),
+            transform=rio.transform.from_origin(0, 100, 1, 1),
             crs=4326,
         )
-        with pytest.warns(RuntimeWarning, match="All-NaN slice encountered"):
-            raster_boolean.reproject(ref_dem, resampling="nearest")
+        ref_dem = gu.Raster.from_array(
+            np.random.randint(100, size=(100, 100), dtype="uint8"),
+            transform=rio.transform.from_origin(0, 100, 1, 1),
+            crs=4326,
+        )
+
+        # non masked dem with no intersection
+        dem_bool_crop = dem_bool.icrop((0, 0, 20, 20))
+        ref_dem_crop = ref_dem.icrop((40, 40, 60, 60))
+
+        assert not dem_bool_crop.get_footprint_projected(ref_dem_crop.crs).intersects(
+            ref_dem_crop.get_footprint_projected(ref_dem_crop.crs)
+        )[0]
+
+        res = dem_bool_crop.reproject(ref_dem_crop, resampling="nearest")
+        assert isinstance(res, gu.raster.raster.Raster)
+        assert res.get_stats(stats_name="Valid count") is np.nan
+
+        # non masked dem with no intersection
+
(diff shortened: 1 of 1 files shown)


geoutils/raster/_geotransformations.py
```diff
@@ -385,8 +383,4 @@ def _rio_reproject(src_arr: NDArrayNum, reproj_kwargs: dict[str, Any]) -> NDArra
     # Run reprojection
     _ = rio.warp.reproject(src_arr, dst_arr, **reproj_kwargs)
 
-    # If output needs to be converted back to boolean
```
rhugonnet, line 388: We need to keep this conversion somewhere, otherwise reprojecting a `boolean` raster will yield a `uint8` raster, and we lose consistency in `dtype` while `reproject()` is supposed to conserve it.

geoutils/raster/geotransformations.py
```diff
@@ -124,10 +124,11 @@ def _reproject(
         src_arr = source_raster.data.filled(src_nodata)
         # All masked values must be set to a nodata value for rasterio's reproject to work properly
         dst_arr = _rio_reproject(src_arr, reproj_kwargs=reproj_kwargs)
-        # Set mask
-        dst_arr = np.ma.masked_array(dst_arr.astype(dtype), fill_value=nodata)
+        # Save mask
         if nodata is not None:
-            dst_arr.mask = dst_arr == nodata
+            mask = dst_arr == nodata
```
rhugonnet, line 129: The problem having a logic outside of `_rio_reproject` here, is that it won't work on `_multiproc_reproject` or `_dask_reproject` (called right above), so results will be inconsistent between Multiprocessing and normal calls to this function.
We need to find a way to make the fix either within `_rio_reproject`, or at least mirrored in the other `_xxx_reproject` functions. :wink:

geoutils/raster/_geotransformations.py
rhugonnet: I see two solutions here to avoid the problems (this one and the one I mention further below):
1. We change the `dst_arr` output in a `dst_arr, dst_mask`, to specify where nodata values are even for `bool` types,
2. We return `dst_arr` as a masked array instead of a classic array, which supports having a mask for any `dtype`.

geoutils/raster/geotransformations.py
rhugonnet: See my comment above on potential solutions for a fix there.

Issue #772, opened 2025-11-26
BIGTIFF in multiproc


adebardo:
# Context 

When multiprocessing attempts to save data larger than 4GB, the process crashes. 

# Solution 

It has been suggested to add support for the GDAL BigTIFF argument: https://gdal.org/en/stable/drivers/raster/gtiff.html#config-BIGTIFF_OVERVIEW  
I propose using the IF_NEEDED option. 

# Code 

I propose adding the argument directly here: https://github.com/GlacioHack/geoutils/blob/main/geoutils/raster/distributed_computing/multiproc.py#L233  
with a log message to notify the user.  

# Unit tests 

1/ Check that the log is emitted  
2/ Investigate whether the metadata contains the information, and if so, test for it  

# Documentation

Add a warning in the multiproc documentation

PR #779, opened 2025-12-12
Dealing with BIGTIFF in multiproc 

marinebcht:
Resolves https://github.com/GlacioHack/geoutils/issues/772

### Explanations
Added the possibility to save BigTiff output with multiprocess that which allows dealing image larger than 4 GB : https://gdal.org/en/stable/drivers/raster/gtiff.html#config-BIGTIFF_OVERVIEW

The option BIGTIFF=​[IF_NEEDED](https://github.com/marinebcht/geoutils/blob/2ff0d17ff7164d6412534c7ad724970abfd87100/geoutils/raster/distributed_computing/multiproc.py#L243)  was choosen to only create a BigTIFF if it is clearly needed as explained in the GDAL documentation.

After the process of writing every patch in the output image, a warning is appear in that case : "Due to the size of the output raster, it has been saved in BigTIFF format." ([here](https://github.com/marinebcht/geoutils/blob/2ff0d17ff7164d6412534c7ad724970abfd87100/geoutils/raster/distributed_computing/multiproc.py#L227))

The test to tell if the output image is a bigTiff was implemented in [Raster.isBigTiff](https://github.com/marinebcht/geoutils/blob/2ff0d17ff7164d6412534c7ad724970abfd87100/geoutils/raster/raster.py#L3910) : [In the file header, BigTIFF is declared as 0x002B (43) at offset 2 bytes as compared with the TIFF version of 0x002A (42)](https://www.loc.gov/preservation/digital/formats/fdd/fdd000328.shtml)

### Tests

As dealing with image larger than 4 GB will be verytedious on a daily basis, only the isBigTiff with normal image was [tested](https://github.com/marinebcht/geoutils/blob/2ff0d17ff7164d6412534c7ad724970abfd87100/tests/test_raster/test_raster.py#L1910)... 

A issue is now open https://github.com/GlacioHack/geoutils/issues/784 to be able to work on this kind of data ([here](https://github.com/marinebcht/geoutils/blob/2ff0d17ff7164d6412534c7ad724970abfd87100/tests/test_raster/test_distributing_computing/test_multiproc.py#L257)) and with a future scaling up of uses.

### DIFF
--- a/tests/test_raster/test_raster.py
+++ b/tests/test_raster/test_raster.py
@@ -1907,11 +1907,14 @@ class TestRaster:
             red_c.data.data.squeeze().astype("float32"), img.data.data[0, :, :].astype("float32"), equal_nan=True
         )
 
-    def test__is_bigtiff(self) -> None:
-        """Test _is_bigtiff function"""
+    @pytest.mark.skip()  # type: ignore
+    def test__is_bigtiff_true(self) -> None:
+        """Test _is_bigtiff function for BigTIFF"""
+
+    def test__is_bigtiff_false(self) -> None:
+        """Test _is_bigtiff function for classic TIFF"""
         img = gu.Raster(self.landsat_rgb_path)
         assert img._is_bigtiff() is False
-        # TODO: test with BigTIFF image
 
 
 class TestMask:


geoutils/raster/raster.py
```diff
@@ -3906,6 +3907,23 @@ def subsample(
             array=self.data, subsample=subsample, return_indices=return_indices, random_state=random_state
         )
 
+    def isBigTiff(self) -> bool:
```
rhugonnet: We should name this attribute `_is_bigtiff`? (non-public + no camel-case as it is reserved for classes).

marinebcht: you write.. did not think about this !

marinebcht:
Changed the isBigTiff to _is_bigtiff :) 

About your comment "The general outline looks good to me, **to finalize the tests**" :

I opened a new issue https://github.com/GlacioHack/geoutils/issues/784 to deal with data > 4 GB. We can speak about this Friday. But do you see other tests to add right now ?

rhugonnet:
No I think this is fine, I said this because I saw your skipped test. 
This is an option in Rasterio (already tested there), so we don't have much to test, except the basic behaviour.

geoutils/raster/raster.py
```diff
@@ -3906,6 +3907,23 @@ def subsample(
             array=self.data, subsample=subsample, return_indices=return_indices, random_state=random_state
         )
 
+    def _is_bigtiff(self) -> bool:
+        """
+        Test is the raster file name is a BigTiff or a normal Tiff.
```
belletva: Replace "BigTiff" by "BigTIFF"
Replace "normal Tiff" by "classif TIFF"

geoutils/raster/raster.py
```diff
@@ -3906,6 +3907,23 @@ def subsample(
...
+        In the file header, BigTIFF is declared as 0x002B (43) at offset 2 bytes as compared with the TIFF
+        version of 0x002A (42)"
+
+        :return: if the file name is a BigTiff or not
```
belletva: BigTIFF

Issue #651, opened 2025-02-12
Conflict between arguments `co_opts` and `compress`/`tiled` of Raster.save


adehecq:
In Raster.save, there are two options to handle compression and tiling of the output file, either through the arguments `compress` and `tiled`, or through `co_opts`.
Because the first ones have default values, if one tries to set them through `co_opts`, one gets a rasterio that is not so helpful.
I think we should detect this conflict beforehand and raise a warning instead.

Here is a MWE:
```
import geoutils as gu
img = gu.Raster(gu.examples.get_path("everest_landsat_b4"))
img.save("tmp.tif", co_opts={"tiled":"yes"})
```
-> raises a "TypeError: rasterio.open() got multiple values for keyword argument 'tiled'"

PR #796, opened 2026-01-02
Fix conflict between arguments co_opts and compress/tiled of Raster.to_file() (previously .save())

marinebcht:
Resolves https://github.com/GlacioHack/geoutils/issues/651 

"In Raster.save, there are two options to handle compression and tiling of the output file, either through the arguments compress and tiled, or through co_opts. Because the first ones have default values, if one tries to set them through co_opts, one gets a rasterio that is not so helpful."

To avoid possible conflicts, the two underlying parameters have been removed and can now only be redefined by the co_opts dictionary, by modifying their values:

The default values are : 
- `COMPRESS`=`DEFLATE`
- `BIGTIFF`=`IF_SHARED`, saving output image as BigTIFF format in case of a data > 4Go. `IF_SAFER` was chosen over `IF_NEEDED`, because `IF_NEEDED` only "work" when no compression is used
- other params here: https://gdal.org/en/stable/drivers/raster/gtiff.html#creation-issues), like `TILED`=`NO`

### Example

```
import geoutils as gu
img = gu.Raster(gu.examples.get_path("everest_landsat_b4"))
img.to_file("tmp.tif", co_opts={"tiled":"yes"})
```

The output image is  classical TIFF tiled and compressed with the `DEFLATE`value.

### Test

The co_opts parameters were taken into account is test_to_file : 
- no co_opts 
- one value in the dict : {"COMPRESS": "NONE"} to verify if the output tiff is not tiled (>1)
- several values in the dict : {"COMPRESS": "LZW", "BIGTIFF": "YES"} to verify is the compression has been changed and file is now a BigTIFF.

### DIFF
--- a/geoutils/raster/raster.py
+++ b/geoutils/raster/raster.py
@@ -2778,6 +2778,8 @@ class Raster:
         Tiled default value is set to 'NO' as the GDAL default value.
         Raster is saved as a BigTIFF if the output file might exceed 4GB and as classical TIFF otherwise.
 
+        Example: dem.to_file(to_file, co_opts={'TILED':'YES', 'COMPRESS':'LZW'})
+
         :param filename: Filename to write the file to.
         :param driver: Driver to write file with.
         :param dtype: Data type to write the image as (defaults to dtype of image data).


geoutils/raster/raster.py
```diff
@@ -2795,6 +2793,11 @@ def to_file(
 
         if co_opts is None:
             co_opts = {}
+
+        # Set compression default value to DEFLATE
+        if "COMPRESS" not in map(str.upper, co_opts.keys()):
```
belletva, line 2802: what about "TILED", this case is not replace ?

marinebcht: Tiled keeps its default value from GDAL, so no need to re-set it
Compress, on the other hand, takes "deflate" value by default in xdem (and None in GDAL).
But I will write an example to be more explicit.

marinebcht:
Added BIGTIFF = IF_SAFER in co_opts, saving output image as BigTIFF format in case of a data > 4Go. IF_SAFER was chosen over IF_NEEDED, because  IF_NEEDED only "work" when no compression is used (https://gdal.org/en/stable/drivers/raster/gtiff.html), which is not the case for us (compress=deflate by default). 

We can also change this parameter to IF_NEEDED if compression is set to NONE, but I am not convinced of the advantage compared to the complexity of the configuration.

PR #812, opened 2026-01-08
Downsample <1 unallowed

### LINKED ISSUE #778: Downsample as <1
# Context 

It appears that we are allowing users to downsample with a value < 1, for example 0.5, and therefore perform upsampling. We would also like to add protection against negative values. 

# Code 

Adapt the code here: https://github.com/GlacioHack/geoutils/blob/main/geoutils/raster/raster.py#L513

# Unit tests 

- add / to be updated

adebardo:
Resolves #778 

# Context

We add a security for the downsampling argument.
The user can't instanciate a downsample strictly inferior to 1

# Code

We used a raise ValueError

# Test

We used pytest.raises


rhugonnet (approved):
OK for me too

PR #815, opened 2026-01-08
Add `grouped_stats` for categorical or continuous binning

adebardo:
Resolves #774

# Context

The purpose of this PR is to offer users a simplified API for implementing grouped statistics for their raster. 

To do this, we use pandas' capabilities to work on 1D arrays and their associated classification. 

We implemented the function directly in the base.py file as well as the TUs.
 
# Code

1. Numerous safeguards have been put in place to ensure that users submit their entries in the correct format, as pandas is not capable of covering everything.
2. Special case: If a user wishes to enter a single number (such as a threshold), the interval created is between -inf and +inf. 
3. Boolean masks can be used, and multimodal masks (e.g. landcover) can also be managed. 

# Tests

For testing purposes, we provide a fake raster to ensure that field truths can be calculated by hand. 
 We have implemented one test per bin type. 

We also introduce panda dataframe equality into this file and the necessary tests.

# Documentation

https://adebardo-geoutils.readthedocs.io/en/774-grouped-stats/stats.html

We have added documentation to the statistics tab. Currently, tests are being carried out on raster altitude assumptions against the same raster, as well as the binarisation of the glacier mask. We have opened a ticket to propose data additions.


adebardo:
- saving generated masked for bins
- rename to zonal_stats

doc/source/stats.md
```diff
@@ -93,6 +93,90 @@ inlier_mask = rast > 1500
...
+
+A dictionary containing the masks that have been created during the computation will also be returned by the function.
+Using geoUtils functions makes it very easy to visualise them.
+
+```{code-cell} ipython3
+from geoutils.stats import grouped_stats
+import math
+import matplotlib.pyplot as plt
+
+group_by = {"rast": rast}
```
belletva: I propose to replace `group_by ` by `bins_criteria` and `to_aggregate` by `compute_stats_on`. What do you think @adehecq @rhugonnet ?

rhugonnet: Hmm... 
Well, before thinking names, the `to_aggregate` argument didn't exist in my original implementation (and I had explained why we probably wouldn't need it here: https://github.com/GlacioHack/geoutils/pull/668#issuecomment-2836954792; but maybe I was wrong).
So I think we first need to understand why it's been added and its exact purpose, which is not clear from this doc page (where it is always exactly equal to `group_by`, so serves no purpose), or the PR description (currently empty). I let @adebardo explain the logic :wink:

For the `group_by` argument, I would simply use `groups` (same name as Xarray's `groupby_bins` [here](https://docs.xarray.dev/en/latest/generated/xarray.DataArray.groupby_bins.html), which is the closest function to what we do here).

geoutils/stats/grouped_stats.py
```diff
@@ -0,0 +1,171 @@
...
+
+def from_raster_to_flattened(dict_raster: dict[str, Any]) -> dict[str, Any]:
+    """
+    Transform every raster in flattened data to be processed by pandas
+    :param: Dictionary containing binning name associated to be grouped by and a raster
+    """
+    dict_arrays = dict_raster.copy()
+
+    for key, raster in dict_raster.items():
+        dict_arrays[key] = raster.data.data.flatten()
```
rhugonnet: Careful, we lose the masked array by calling `.data` twice here, so masked values will be gone. Need to call it only once.

Also, `flatten()` creates a copy, duplicating RAM usage, so we need to use `ravel()` instead.

geoutils/stats/grouped_stats.py
```diff
@@ -0,0 +1,171 @@
...
+
+    return dict_arrays
+
+
+def is_interval(bin_value: RasterType | list[int] | object) -> bool:
```
rhugonnet: Do we really need this to add this check ourselves? Can't we let Pandas do the job (feed this to `pd.cut()` directly, like done below which will raise exceptions on its own). Then get the output intervals, and we can use those to derive all next steps (including our own raster masks).

It would be a massive hassle to recode all the pandas logic here (they support `int` for instance too, to support a number of bins).

Also, I don't see when "bins" could be a `RasterType`, do you have an example?

geoutils/stats/grouped_stats.py
```diff
@@ -0,0 +1,171 @@
...
+    bins: dict[str, RasterType | list[int] | object],
+    aggregated_vars: dict[str, NDArrayNum],
+    statistics: list[str],
+) -> tuple[Any, dict[str, dict[str, NDArrayNum] | None]]:
+    """
+    Get statistics grouped (=binned) by other variables, whether categorical or continuous.
+    :param groupby_vars: Dictionary containing Raster to group by.
+    :param bins: Bins to use. Can be a list of interval, binary mask or segmentation map.
+    :param aggregated_vars: Values to group, can be a dictionary .
+    :param statistics: List or dict of statistics to compute, e.g. ["mean", "std"].
```
rhugonnet: Once we've converged on the API, we'll need to explain the arguments in more details!

geoutils/stats/grouped_stats.py
```diff
@@ -0,0 +1,171 @@
...
+            returned_masks[group_col] = {}
+
+        # ---------- Interval bins ----------
+        if is_interval(bins_array):
+            cut = pd.cut(df[groupby_key], bins=bins_array)
+            df[group_col] = cut
+
+            if returned_masks[group_col] is not None:
+                for interval in cut.cat.categories:
+                    mask_flat = (cut == interval).to_numpy()
```
rhugonnet: Can you comment what the code does here?

geoutils/stats/grouped_stats.py
```diff
@@ -0,0 +1,171 @@
...
+        # ---------- Raster (mask or segmentation) ----------
+        elif isinstance(bins_array, Raster):
+            data = np.asarray(bins_array.data.data).flatten()
+
+            if len(data) != len(df):
+                raise ValueError(f"Mask has invalid length {len(data)}, expected {len(df)}")
+
+            # Binary mask
+            if bins_array.is_mask:
+                df[group_col] = pd.Categorical(data, categories=[False, True])
```
rhugonnet: OK, I understand a bit more here, but not completely. 🤔 

Why would the raster be needed in the `bins` category? If it's a binary categorical variable, it would be passed in `groupby_vars` (along with the other rasters used for grouping). Then, by checking its `dtype`, we would know it is `boolean` and that we thus categorize per `True`/`False`, so we always know what its associated `bins` value (the user can simply pass `None` in this case).

Or did I miss something?

geoutils/stats/grouped_stats.py
```diff
@@ -0,0 +1,171 @@
...
+
+                if returned_masks[group_col] is not None:
+                    mask_to_raster = Raster.from_array(data.reshape(shape).astype("uint8"), transform, crs)
+
+                    returned_masks[group_col][f"{groupby_key}_mask"] = mask_to_raster
+
+            # Segmentation
+            else:
+                df[group_col] = pd.Categorical(data)
```
rhugonnet: Same here: If the user doesn't pass anything in `bins` for an `integer` array, we know they want to use all the categories. I don't think we need a `Raster` as input of `bins`?

rhugonnet:
@adebardo Thanks for all this work! Nice that it's taking shape, I think it's getting close, and we should be able to add zonal stats shortly after :slightly_smiling_face:. (I'm actually working on separate code that should allow to add Dask/Multiprocessing support fairly easily there).
I started a quick review (I realized I forgot after the meeting, sorry! thankfully @belletva's comment reminded me).

On the technical aspects of the implementation: To be honest, I had a lot of trouble understanding them. I hope I didn't miss anything obvious :sweat_smile:
I think this is a comment we made before: The PR description really needs to **detail the logic behind the changes, how things work and why the design is chosen a certain way**. It's usually 5min to write it up when you work on the PR. And otherwise it's 5 times more work for me (and any other reader) to understand what was done just from the code (I just spent 30min failing at understanding it), and we often end up going in circles... :confused:

doc/source/stats.md
adebardo: “to_aggregate” is one of the names I used to replace the values that did not seem obvious in the first proposed API. The second name is the equivalent in the PR description and is aggregate_vars.

doc/source/stats.md
```diff
@@ -93,6 +93,51 @@ inlier_mask = rast > 1500
...
+across classes, bins, or segmentation derived from the data itself.
+
+
+### Example with altitude intervals
+In this example, we will create different altitude classes from a chosen interval [400, 1000, 2000, 3000, >3000].
+Once these bins are created, we reapply them and compute the mean, minimum, and maximum values of the same raster
+for each sub-interval. It is also possible to use a reference other than the raster itself for the group_by.
+
+A dictionary containing the masks that have been created during the computation will also be returned by the function.
+Using geoUtils functions makes it very easy to visualise them.
```
adehecq: geoUtils -> GeoUtils

doc/source/stats.md
```diff
@@ -93,6 +93,51 @@ inlier_mask = rast > 1500
...
+
+group_by = {"slope": rast.slope()}
+bins = {"slope": [0, 5, 20]}
+statistics = ["mean", "min", "max"]
+
+df = rast.grouped_stats(group_by, bins, statistics)
+df
+```
+
+### Example with altitude masks
```
adehecq: it does not have to be specifically "altitude" masks? So it could be generic "Example with raster mask"?

doc/source/stats.md
```diff
@@ -93,6 +93,51 @@ inlier_mask = rast > 1500
...
+
+In this example, we will create a mask such as altitude is more than 2000 meters.
```
adehecq, line 122: The example is not very informative, as one could do the same with 
```
rast_masked = rast[rast>2000]
rast_masked.get_stats()
```
no?
Maybe we can use a vector file instead (we have the associated glacier shapefile I believe) and rasterize it to create a glacier mask? This would be very much an example for zonal statistics then.

doc/source/stats.md
```diff
@@ -93,6 +93,51 @@ inlier_mask = rast > 1500
...
+for each sub-interval. It is also possible to use a reference other than the raster itself for the group_by.
+
+A dictionary containing the masks that have been created during the computation will also be returned by the function.
+Using geoUtils functions makes it very easy to visualise them.
+
+```{code-cell} ipython3
+import math
+import matplotlib.pyplot as plt
+
+group_by = {"slope": rast.slope()}
```
adehecq: Ok I see the problem of dependency with xDEM...
Maybe we should add an additional dataset for the examples? Could we add a landcover raster dataset?

doc/source/stats.md
```diff
@@ -93,6 +93,56 @@ inlier_mask = rast > 1500
...
 
+## Grouped statistics
+
+GeoUtils provides support for grouped statistics, allowing statistics to be computed independently over subsets of data
+defined by one or more grouping bins. This is particularly useful when analyzing how statistical properties vary
+across classes, bins, or segmentation derived from the data itself.
+
+
+### Example with altitude intervals
+In this example, we will create different altitude classes from a chosen interval [400, 1000, 2000, 3000, >3000].
```
belletva: not the same interval between the text and the code

doc/source/stats.md
```diff
@@ -93,6 +93,56 @@ inlier_mask = rast > 1500
...
+Once these bins are created, we reapply them and compute the mean, minimum, and maximum values of the same raster
+for each sub-interval. It is also possible to use a reference other than the raster itself for the group_by.
+
+A dictionary containing the masks that have been created during the computation will also be returned by the function.
+Using GeoUtils functions makes it very easy to visualise them.
+
+```{code-cell} ipython3
+group_by = {"raster": rast.data}
```
belletva, line 112: Why there is no file linked to rast ?

doc/source/stats.md
```diff
@@ -93,6 +93,56 @@ inlier_mask = rast > 1500
...
+In this example, we will create a mask such as altitude is more than 2000 meters.
+Once these masks are created as a raster, we reapply them and compute the mean, minimum, and maximum values of the same raster
+for masks = True. It is also possible to use a reference other than the raster itself for the group_by.
+
+```{code-cell} ipython3
+import matplotlib.pyplot as plt
+
+filename_rast = gu.examples.get_path("everest_landsat_b4")
+filename_vect = gu.examples.get_path("everest_rgi_outlines")
+rast = gu.Raster(filename_rast)
```
belletva: rast =  gu.Raster(gu.examples.get_path("everest_landsat_b4")) 
vect = gu.Vector(gu.examples.get_path("everest_rgi_outlines"))

doc/source/stats.md
adebardo: Because we cover only NDArrayNum here, do you want to be more inclusive ?

belletva: It is in case the user wants to reproduce the example. It is not possible with this example.

adebardo: I'm not sure to understand the problem, but if you look at the beginning of stats.md file you'll see a file attached to raster

rhugonnet:
Closing with #925

PR #826, opened 2026-01-16
Add test for is_bigtiff() function

### LINKED ISSUE #784: Manage data > 4Go (BIGTIIF)
As dealing with images larger than 4 GB – corresponding to BIGTIFF format file – will be very tedious on a daily basis, no test was implemented in the current CI after this PR https://github.com/GlacioHack/geoutils/pull/779.

The idea is to find a solution (additional CI used only on request ?) to deal with this data or other specific processes which require much more time.

adebardo:
resolves #784 

We discuss about the fact to use Mock to create fake data, but it seems complicate

Indicate Bigtiff="Yes" seems to work even if the data is small

### DIFF
--- a/tests/test_raster/test_raster.py
+++ b/tests/test_raster/test_raster.py
@@ -1937,9 +1937,26 @@ class TestRaster:
             red_c.data.data.squeeze().astype("float32"), img.data.data[0, :, :].astype("float32"), equal_nan=True
         )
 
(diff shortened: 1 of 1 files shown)


rhugonnet (approved):
Perfect, thanks! I didn't know about the `tmp_path` fixture, very practical! :smile: 

We could consider using it everywhere

PR #855, opened 2026-01-30
Improve typing, input checks and import management for package scalability

### LINKED ISSUE #728: Remove Mypy plugin for NumPy that is deprecated
See: https://numpy.org/devdocs/release/2.3.0-notes.html#deprecations

Need to remove the `mypy.ini` line. The backwards-comp on this does not seem great, so we'd have to do it  soon, even if running NumPy lower than 2.3.

In the meantime, I've been having some trouble setting up Mypy locally because of it:
- We need to pin an earlier NumPy version than 2.3,
- Yet I still get new linting hints I didn't have before...

### LINKED ISSUE #827: Add duck typing to avoid circular import risks
Right now, we have several risks for circular imports across the package, for instance:

1/ `geoutils/raster/raster.py` imports `geoutils/raster/geotransformations.py` (because we want to use `_reproject` in the `Raster class`
2/ `geoutils/raster/geotransformations.py` also imports `geoutils/raster/raster.py`, because we want to accept `Raster` as an input argument of `_reproject()`.

Having 1/ is normal, but 2/ is not good practice (feature should not depend on the object), and currently only works because of import order and partial initialization, but could break easily in the future, and become hard to maintain.

Thankfully, we can fix this easily while respecting static type checking using "duck typing" (or "structured typing" with declared types) by defining what methods are expected to be called (without declaring explicitly we want a Raster), which will also be useful to accept `xr.DataArray` inputs.

For example, if we need to have `icrop()` supported by our feature;
```
from typing import Protocol, TypeVar

# We specify that the function below needs an input with icrop method
class RasterLike(Protocol):
    def icrop(): ...

# We specify input/output of the functions will be the same, no matter their exact type
RasterType = TypeVar("RasterType", RasterLike)

def _reproject(rast: RasterType, ...) -> RasterType:
    rast.icrop()
```

### LINKED ISSUE #850: Add consistent input check function and tests for dual match-reference with Xarray/Raster
Right now several functions do this separately, this would make input check + error raising consistent.

Then, we should add tests for all relevant functions (`reproject()`, `crop()`, `create_mask()`, `rasterize()`, `proximity()`).

### LINKED ISSUE #856: Allow untyped decorators to avoid repeated type ignores
We are always ignoring decorator types, especially in tests for `@pytest.mark.parametrize` where we have hundreds of those.
Given that our source code does not rely on decorators significantly, we can ignore them consistently to have cleaner code everywhere (and save time when resolving Mypy errors).

Fixed this in #855

### LINKED ISSUE #857: Argument `input_latlon` behaves differently for `interp_points` and `reduce_points`
So one of them has a wrong result. Maybe a remnant of the old `always_xy=False` default behaviour of Pyproj?

Will fix in #855.

### LINKED ISSUE #858: Max coordinate check inverted in `_outside_image`
Old function that isn't used much, the row/col are inverted in one of the two checks.

rhugonnet:
This PR restructures imports and type management for robustness (notably to avoid any circularity issues) while preserving typing and function behaviour, using the following (also explained with examples [in this new Wiki page](https://github.com/GlacioHack/xdem/wiki/Good-practices-for-managing-dependent-objects-and-features-(import,-typing))):

1. **Non-runtime type checking:** We use `typing`'s `TYPE_CHECKING` at top of file to isolate relevant typing imports, 
2. **Duck typing:** This means doing `hasattr(obj, "crs")` instead of `isinstance(obj, Raster)`, thereby removing the need to import `Raster` for runtime in every feature class (i.e. remove circularity, as features are already imported by the objects). We add custom helper functions in  in `_dispatch.py` that check both normal object attributes `obj.attr` and Xarray/Pandas accessor attributes `obj.rst.attr` (making it consistent for any object type),
3. **In-method imports:** When a cross-import is absolutely necessary (e.g., creating a PointCloud in `Raster.interp()`, or a `Raster` in `PointCloud.grid()`, without circularity issue), it is done within the function (i.e. at runtime), to avoid any circularity issue.

These three methods allow us to remove all of the `import geoutils as gu` we had at the top of nearly every file (bad practice that we had to use `gu.Raster` for typing, user-input checks, etc), which created partially initalized modules everywhere and big risks of package breakage through circular imports.

This PR also:

- Re-locates the Multiprocessing module (previously in `raster/`), moving it to the root, and separating functions specific to the raster functionalities (e.g. `multiproc_reproject()`), and ones generic to its block-behaviour. This also allows to reduce some circularity in imports.
- Removes the need to type decorators (by adding a MyPy exception in `pyproject.toml`), thereby removing hundreds of `# type: ignore` we were accumulating (especially in the tests next to `@pytest.mark.parametrize`, but also in source code). This would have technically fit better in a separate PR... But well, I had already started on it :sweat_smile:.
- Merges consistent input checks for functions requiring similar inputs in the new `_dispatch` module, with a series of tests in `test_dispatch.py`, and raise clearly defined exceptions defined in a new `exceptions.py` module (needs to be public for users to catch errors if desired).

Finally, we now have to separate typing into 3 categories for each object (only for raster for now, but will be true for all when we add Pandas accessors).
Those are:
- `RasterBase` (parent class of `Raster` and `RasterAccessor`): Used internally. Should be used to type any function that receives a `Raster` or `RasterAccessor` itself (usually only the first argument of a `_feature()` functions that was passed `self` from the `RasterBase` class).
- `RasterLike = Union[Raster, xr.DataArray]`: The generic raster input type. This is to cover also Xarray objects as input (Note: the `RasterAccessor` class **does not cover this**, as it is just a "temporary" extension class that applies methods to the `xr.DataArray`, but does not shares its type; so this type needs to be different).
- `RasterType = TypeVar("RasterType", "RasterBase")`: Situational. This is to indicate functions that are passed a subtype of `RasterBase` object (for example, a `DEM`) and return the exact same subtype.

Resolves #827
Resolves #728 
Resolves #850
Resolves #857 
Resolves #858 
Resolves #856 

TO-DO:

- [x] Fix typing of projtools, interp_points and reduce_points,
- [x] Consolidate single function for input checks of "reproject match-ref", "crop match-ref", "interpolation match-ref", etc?
- [x] Link new check_match_grid with existing functions (reproject, rasterize, grid)

### DIFF
--- a/geoutils/_dispatch.py
+++ b/geoutils/_dispatch.py
@@ -133,8 +133,9 @@ def _check_bounds(
 
         for k, v in bbox.items():
             if not isinstance(v, (int, float)):
-                raise InvalidBoundsError(f"Bounding box dictionary value for {k!r} must be numeric, got"
-                                         f" {type(v).__name__}.")
+                raise InvalidBoundsError(
+                    f"Bounding box dictionary value for {k!r} must be numeric, got" f" {type(v).__name__}."
+                )
         xmin, ymin, xmax, ymax = bbox["left"], bbox["bottom"], bbox["right"], bbox["top"]
 
     # If bbox is an iterable with 4 coordinates (excluding strings and bytes)
--- a/geoutils/vector/geotransformations.py
+++ b/geoutils/vector/geotransformations.py
@@ -29,7 +29,7 @@ from geoutils._dispatch import get_geo_attr, has_geo_attr
 
 if TYPE_CHECKING:
     from geoutils.raster.base import RasterLike
-    from geoutils.vector.vector import VectorLike, Vector
+    from geoutils.vector.vector import Vector, VectorLike
 
 
 def _reproject(
--- a/geoutils/vector/vector.py
+++ b/geoutils/vector/vector.py
@@ -22,9 +22,8 @@ Module for Vector class.
 
 from __future__ import annotations
 
-import warnings
-
 import pathlib
+import warnings
 from collections import abc
 from os import PathLike
 from typing import (
@@ -51,7 +50,7 @@ from pyproj import CRS
 from shapely.geometry.base import BaseGeometry
 
 from geoutils import profiler
-from geoutils._dispatch import _check_match_bbox, _check_crs, get_geo_attr, has_geo_attr
+from geoutils._dispatch import _check_match_bbox, get_geo_attr, has_geo_attr
 from geoutils._misc import copy_doc, deprecate, import_optional
 from geoutils._typing import NDArrayBool, NDArrayNum, Number
 from geoutils.interface.distance import _proximity_from_vector_or_raster
@@ -1576,8 +1575,9 @@ class Vector:
 
         # Deprecate old xres and yres
         if "xres" in kwargs.keys() or "yres" in kwargs.keys():
-            warnings.warn(message="Input 'xres' and 'yres' are deprecrated in favour of 'res'.",
-                          category=DeprecationWarning)
+            warnings.warn(
+                message="Input 'xres' and 'yres' are deprecrated in favour of 'res'.", category=DeprecationWarning
+            )
         xres = kwargs.get("xres", None)
         yres = kwargs.get("yres", None)
         if xres is not None:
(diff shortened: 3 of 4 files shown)


rhugonnet:
Oh, and I forgot to mention the obvious big thing to summarize: This PR basically allows to remove all of the `import geoutils as gu` we had at the top of nearly every file (to use `gu.Raster` for typing, user-input checks, etc), which created partially-initalized modules everywhere and big risks of package breakage with circular imports.

Now this is all solved thanks to the above.

PR #862, opened 2026-02-08
Restructure, homogenize and link chunked Dask/Multiprocessing operations

### LINKED ISSUE #258: Improve Vector.rasterize() test coverage
There is very little test coverage of the Vector.rasterize() operation. PR #254 has introduced two lightweight tests but these are not sufficient as they only provide partial tests of `crs` and `xres` inputs. We need to implement tests for the other input arguments too.

### LINKED ISSUE #720: Update uninformative error message in `subsample`
Relic of old code, the message literally says: "Something is wrong with array dimension, check input data and shape" 😅 
See:
https://github.com/GlacioHack/geoutils/blob/e79a1ff8948709fda445835f30cf7dae649e046b/geoutils/raster/sampling.py#L100

### LINKED ISSUE #745: Match performance of Dask reprojection with Multiprocessing
Our Dask reprojection functions well, about 40s to reproject a raster of 50,000x50,000 out-of-memory on 2,500x2,500 chunks. (and keeping memory low according to chunk size, 1.2GB max)

However, the Multiprocessing one (not public yet in the doc), despite using most of the same underlying code, is not. It (sometimes?) loads the full dataset
Here's a reproducible example, to add in `test_delayed_reproject__memusage` at the bottom of our old Dask memory tests now in "benchmark" [here](https://github.com/GlacioHack/geoutils/blob/8554fa410c647a3bf58f53a777439ab4d2cbc3ee/benchmark/test_raster_dask.py#L371).

```python
          # Test multiprocessing
          from geoutils.raster.distributed_computing.multiproc import _multiproc_reproject, MultiprocConfig
          mp_config = MultiprocConfig(chunk_size=2500, outfile=fn_tmp_out.replace(".nc", "_reproj_mp.nc"))
          from geoutils import Raster
          rst = Raster(fn)
          _multiproc_reproject(
              rst=rst,
              config=mp_config,
              src_transform=src_transform,
              src_crs=src_crs,
              dst_transform=dst_transform,
              dst_crs=dst_crs,
              dst_shape=dst_shape,
              src_nodata=src_nodata,
              dst_nodata=dst_nodata,
              resampling=resampling,
              dtype=rst.dtype,
          )
```

One can adjust the array shape at the start of the test class, and chunksizes in the test parameters.

Likely culprits for Multiprocessing:
- `Raster.icrop()` issue,
- Driver issue or chunk definition issue when reading from Rasterio instead of Xarray/Rioxarray.

### LINKED ISSUE #830: Re-organize multiprocessing and chunked modules
We need to update the architecture of chunked operations to ensure we have a clear and robust structure for chunked features. (Right now it works without circularity, but it's hard to grasp).
Here's the logic:
- Our objects methods `Raster.feature()` calls a main function `feature()`,
- The parent function `feature()`  dispatches towards either a `_feature()` (core function, unchunked), or a `_feature_dask` (Dask backend), or `_feature_multiproc` (Multiproc backend),
- The `_feature_dask` and `_feature_multiproc` rely on a similar block/chunk logic, so both call a `_feature_chunk`, which itself calls `_feature()` underneath.

This logic is inevitable to be consistent in our functions, and requires us to have an adapted structure.
In terms of layers, the mental model we have is this:
```
Domain objects (RasterBase / VectorBase / PointCloudBase; inherited by accessors and main objects)
↓
Feature API (parent functions for reproject, proximity, interpolate, grid, for any backend)
↓
Execution backends (Dask / Multiprocessing / Direct core)
↓
Chunked function logic (only for Dask/MP)
↓
Core functions
```

So we need a structure like this, either for each feature or for a group of features:
```
geoutils/
├── feature/                 # feature subsystem
│   ├── __init__.py
│   ├── api.py
│   ├── core.py
│   ├── chunked.py
│   ├── backends/
│   │   ├── __init__.py
│   │   ├── dask.py
│   │   └── multiprocessing.py
```

Also, currently, the `multiproc` module is unnecessarily deep within the `raster/` module. The `cluster` module and most of the `multiproc` could live in a `multiproc/` folder at the root, to provide an fixed API (mirroring Dask's) as additional backend.

### LINKED ISSUE #845: Wrapping of base `interp_points` within Dask function
The first implementation just wrapped `scipy` directly, we should now wrap `interp_points`, the same way `delayed_reproject` wraps `_rio_reproject`.
Explained in #830

### LINKED ISSUE #849: Add support for multi-band rasters in Dask/Multiprocessing reprojection
A small change to add, especially relevant for the Xarray accessor, right now it is not supported and fails.

### LINKED ISSUE #853: Adapt `reduce_points()` to in-memory array
The method currently works only on an array linked to a filename on disk... This fails for a in-memory-only Raster, or with the Xarray accessor...

### LINKED ISSUE #864: Core filter functions are currently ignored through `Raster` API
The `generalized_filter` logic here is making the `Raster.filter()` calls skip all of the core functions of the `filter` module (that run fast on older SciPy + support other engines such as Numba): https://github.com/GlacioHack/geoutils/blob/8d73547d1a171323ddcdea30342b1db218aaf272/geoutils/filters.py#L69

We need to move `generalized_filter` out of here, and apply `generalized_filter` within each individual core function and ONLY if SciPy version is too old for the vectorized version.

This also means that most tests are not covering the right behaviour right now (they test subfunctions directly, not the main API). And so we have erroneous results for edge behaviour (mode="constant" and cval=np.nan).

I'm fixing all of this in #862...

### LINKED ISSUE #865: Polygonize and rasterize API inconsistent and incomplete
We weren't covering all arguments of the subfunctions from Rasterio. Now we'll be after #862 is merged.

### LINKED ISSUE #866: Resolution argument inconsistent between `rasterize` and `create_mask`
One was using xres/yres, the other only res.

Now will be fixed through #862 thanks to #855

### LINKED ISSUE #867: Deprecate `inplace` argument of raster functions to ensure Dask serialization
We had `inplace` for reproject, crop, filter, astype... This is actually not good practice, as it prevents Dask from serializing correctly when chaining ops.

Deprecating this in #862

### LINKED ISSUE #868: Add chunk-invariant Dask subsampling of valid values
Previous implementation was chunk-dependent, now a chunk-invariant one is added in #862

### LINKED ISSUE #869: Add Dask support for `polygonize`
Had an old draft on this, now finalized for #868

### LINKED ISSUE #870: Add Dask support for `rasterize`
Easy enough for the raster part, done in #862

rhugonnet:
This PR restructures existing modules to link chunked functionalities (Dask/Multiprocessing) to the main API consistently, or add/mirror Dask/Multiprocessing support when they didn't exist, and homogenizes the behaviour by making use of #855 for consistent input checks.

In particular, the PR:
- Links existing `subsample` and `interp_points` functionalities to their Dask implementation, and mirrors it with a Multiproc implementation re-using the same logic,
- Adds a new chunked `subsample` strategy that yields a random sample that is chunk-invariant (previous one was chunk-dependent), which is more costly but ensures 100% reproducibility and consistency,
- Adds a Dask/Multiproc implementation for `filter`, `rasterize`, `create_mask` (all of those being fairly easy, wrapping `map_overlap` with some border/depth logic for `filter()`, and wrapping `map_block` with some logic to minimize polygon comparisons for `rasterize`) and `polygonize` (that last one was quite a bit of work, which I had started earlier on).
- Adjusts the chunked implementation for `reproject()` to work on multi-band rasters.

And a lot of tests to ensure those yield **the exact same output**, and **trigger the exact same loading/laziness behaviour**!

Basically, almost all of our functionalities now have consistent chunked support! :smile: 

**Note:** Some methods (right now only `subsample` and `polygonize`) don't really have a "single-best" strategy on how to deal with chunks: it depends how the data is distributed, or how deterministic the result needs to be (order, random seed, etc). So I include all strategies I thought of in the implementation. We can compare their performance in more details in the future benchmarking using the work of @marinebcht :wink:. 

**FYI, to come soon:** I also have another branch which has several chunked/Dask strategies for **pairwise sampling** (for a future spatial stats `variogram()` with optional dependency to SciKit-GStat/others, which we have long planned to move from xDEM to here). It doesn't always make sense to support the sampling logic directly in the home package (as optimizations are very specific to a data type, here regular grids for instance, while variogram functions of these geostats packages except a tuple of coordinates often irregular). So GeoUtils is likely a good place for it, but I'll discuss it with the lead of SciKit-GStat and maybe others.

## Details on restructuration

Most the line changes is moving old Dask/Multiprocessing code into feature modules, and often also restructuring the code of the core function so it can be used within the chunked implementations of Dask/Multiprocessing for full consistency.

Detailed discussion started in #830. 
Initially, the idea was to separate into submodule for each backend for readability.
But in the end, since the re-org of Multiprocessing in #855, code lines are decently short per backend (whether Dask or Multiprocessing, it's 200 max, slightly more for complex code like reprojection/interpolation), so it is better to fit base + chunked code + backend all together inside a single `feature` module (otherwise we need at minima 3 sub-files to keep the logic working, **and the directory tree would explode very fast**...).
So I think it's all OK to keep in a single file for now, as they are often shorter than 1000 lines (a couple exceptions with reprojection, polygonize, and subsampling). If some modules keep growing, we'll just have to turn them into folders with submodules in it.

## Some notable smaller changes

This PR moves the helper classes `GeoGrid` and `ChunkedGeoGrid` (previously used only for reprojection) out of the `transformation` module and into the shared `multiproc/chunked.py` module, where they can be used for multiple implementations. In this PR, we re-use those helper classes for `polygonize` and `rasterize`.

This PR fixes the `reduce_points()` behaviour to work for in-memory and on multi-band rasters.

## Renaming

I made naming more consistent for some modules (only internal anyway): no verb, no plural, instead singular name everywhere. So `interpolation` instead of `interpolate`, `rasterization`/`vectorization` are split from the old `raster_vector` module. Also `transformation` instead of `geotransformations`, `referencing` instead of `georeferencing`.

## Issues solved

Resolves #258
Resolves #720
Resolves #745
Resolves #830
Resolves #845
Resolves #849
Resolves #853
Resolves #864
Resolves #865 
Resolves #866 
Resolves #867 
Resolves #868 
Resolves #869 
Resolves #870 

TO-DOs:

- [x] Link chunked `subsample` and `interp_points` to their own base function,
- [x] Mirror chunked `subsample` and `interp_points` with Multiprocessing,
- [x] Add new strategy for `subsample` that is chunk-invariant,
- [x] Add support for multi-band raster consistently across Dask/Multiprocessing implementations,
- [x] Adjust test structure accordingly...
- [x] Improve speed of chunk mapping for reproject,
- [x] Add non-lazy Xarray dataset to consistency tests,
- [x] Add `band` argument to `subsample`?
- [x] Link easy `map_overlap` or `map_block` operations: `filters` (overlap), `rasterize` and `create_mask` (block).
- [x] Add support for `polygonize`?

### DIFF
--- a/geoutils/stats/sampling.py
+++ b/geoutils/stats/sampling.py
@@ -118,17 +118,17 @@ def _splitmix64(x: np.typing.NDArray[np.uint64]) -> NDArrayNum:
 
     # First mixing step: XOR-shift to spread high bits into low bits, then multiply by a chosen odd constant
     z = x
-    z = (z ^ (z >> 30)) * np.uint64(0xBF58476D1CE4E5B9)
+    z = (z ^ (z >> 30)) * np.uint64(0xBF58476D1CE4E5B9)  # type: ignore[assignment]
     z &= mask
 
     # Second mixing step: Another XOR-shift followed by multiplication with a different constant
     # The constants were empirically chosen to achieve strong avalanche properties (each input bit affects
     # many output bits)
-    z = (z ^ (z >> 27)) * np.uint64(0x94D049BB133111EB)
+    z = (z ^ (z >> 27)) * np.uint64(0x94D049BB133111EB)  # type: ignore[assignment]
     z &= mask
 
     # Final XOR-shift to finish diffusion
-    z = z ^ (z >> 31)
+    z = z ^ (z >> 31)  # type: ignore[assignment]
 
     return z.astype(np.uint64, copy=False)
 


rhugonnet:
FYI @remi-braun @guillaumeeb: Final technical step done, I did a bit more than planned so that we really support most features through Dask.  :slightly_smiling_face:  
Now it's just drafting the documentation, will tag you for review soon!

rhugonnet:
P.S @remi-braun I remember you mentioning an issue with upsampling/downsampling in chunked reprojection in odc-geo: I can confirm we don't have it here. I tested upsampling/downsampling in similar or different CRS, with different resampling, and the output is always 100% the same as in-memory (see [here](https://github.com/GlacioHack/geoutils/blob/ce021f19d50cf16149622effb6fb9db279dbd158/tests/test_raster/test_transformations_raster.py#L805)).

remi-braun:
🤩

guillaumeeb:
Congrats and many thanks!!!

Issue #860, opened 2026-02-03
Contrast enhancement options for raster plotting


atedstone:
**Is your feature request related to a problem? Please describe.**
Our support for contrast enhancement of rasters to be plotted is, if I understand correctly, rudimentary. 

1. It is only possible to supply `vmin` and `vmax` to `Raster.plot()`, and these keywords if supplied are typed only as `float` or `int`. Their values have to be set manually by the user. 
2. Further, it is not possible to apply different vmin and vmax values to different bands of a multi-band image - they are applied to all bands.

**Describe the solution you'd like**
1) In the simplest case, we could allow `vmin` and `vmax` to also take lists/tuples ordered by band, solving the multi-band issue.
2) I would further like to implement some simple contrast enhancement options, e.g. percentile-based min/max stretching much as applied by QGIS.

**Describe alternatives you've considered**
At the moment the best alternative I'm aware of is to manually modify the `.data` contents on the user side before calling `Raster.plot()`, especially in the case of wanting to custom-stretch multi-band images.

**Additional context**
In the case of Sentinel-2 optical TOA imagery it is usually essential to apply stretches to each band separately.

rhugonnet:
Hmmm...
For the second point: I think it would belong as its own `Raster` functionality only if there is usefulness in having a constrast-enhanced raster outside of plotting (needed for some other analysis). Is that the case?

For the first: If we want to be generic, we could support different categories of constrast enhancement, it looks like there typically is:
- **Linear/log/gamma constrast stretching:** Normalization over a range `img_stretched = (img - img.min()) * 255 / (img.max() - img.min())`, I think `skimage` covers those here: https://scikit-image.org/docs/stable/auto_examples/color_exposure/plot_log_gamma.html,
- **Percentile clipping:** Clip to a symmetric percentile of the data which removes outliers and thus enhances the constrast, like Xarray's 2/98th percentile in `robust` mentioned by @remi-braun,
- **Histogram equalization**: Spreads out intensity values according to histogram (better for non-symmetrical distributions I think), as here: https://scikit-image.org/docs/stable/auto_examples/color_exposure/plot_equalize.html

We could wrap `skimage` as an optional dependency if relevant.
I don't use many of these methods myself, so I'd follow your opinion on what level is appropriate here @atedstone :wink:
Do you use any of these often @remi-braun?

atedstone:
[...]

PR #872, opened 2026-02-18
Prevent automatic scalar mapping when plotting RGB images

atedstone:
Relates to prep. work for #860 . I found a bug where we are passing a `cmap` value to `plt.imshow()` when using `Raster.plot()` to plot a multi-band image (e.g. RGB). This results in `imshow()` triggering some kind of scalar remapping of the supplied data. This is not a desired automatic behaviour as we are then not in control of this remapping. In my tests I found that it brings a danger of introducing a colour contrast in some extreme images (e.g. accumulation zone of the Greenland Ice Sheet).

By passing `cmap=None` we disable this behaviour.

### DIFF
--- a/geoutils/raster/raster.py
+++ b/geoutils/raster/raster.py
@@ -2161,10 +2161,10 @@ class Raster(RasterBase):
 
         # Create colorbar
         # Use rcParam default
-        if cmap is None and len(bands) == 1:
+        if cmap is None and isinstance(bands, int):
             # ONLY set a cmap arg for single band images
             cmap = plt.get_cmap(plt.rcParams["image.cmap"])
-        elif cmap is None and len(bands) > 1:
+        elif cmap is None and isinstance(bands, tuple):
             # Leave cmap as None for multi-band image, because if a cmap
             # is passed then imshow treats this as an instruction to apply scalar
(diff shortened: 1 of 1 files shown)


belletva (dismissed):

marinebcht:
Hello @atedstone , I'm trying to reproduce the problem you identified :smiley: 

I might be mistaken, but I tried this (with imshow directely):

```
plt.imshow(data, cmap="gray")
plt.show()
plt.imshow(data, cmap=None)
plt.show()

```

both plots are the same. In the imshow [doc](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.imshow.html), `cmap` "is ignored if X is of type RGB(A)".

Could you show me the code you used to get different results depending on the `cmap` parameter?
Thanks !

atedstone:
Hi @marinebcht , thanks for following up on this. I just tried to reproduce my side, using existing code that I believe I used when preparing this PR. However, that code now shows no difference depending on whether or not `cmap` is provided - i.e. I now get the same result as you. Maybe I did something in error, or perhaps something in my environment changed. Anyway, this can be closed now; apologies for taking your time on this one.

PR #882, opened 2026-04-14
Fix: documentation problems

marinebcht:


belletva (approved):
Good for me

rhugonnet (approved):
Good catch!

PR #888, opened 2026-06-03
Add some parameters as global parameters

### LINKED ISSUE #884: Add resampling and nodata_propagation as global parameters
Following https://github.com/GlacioHack/xdem/issues/854 point 4 and 5

> Some thoughts of global parameters that could be useful for users to access:
> 
> The resampling during reprojection, or during point-grid interpolation (bilinear, cubic, etc). We could use the [global GeoUtils parameters](https://geoutils.readthedocs.io/en/stable/config.html) for this.
> The nodata_propagation behaviour during reprojection/interpolation.

marinebcht:
Resolves https://github.com/GlacioHack/geoutils/issues/884

The idea was to move some recurring parameters as global parameters  [see documentation](https://geoutils.readthedocs.io/en/stable/config.html) to allow for greater consistency.

So, now we have, in addition to `shift_area_or_point` and `warn_area_or_point`: 

```

[resampling]

# Default resampling used during reprojection
reprojection_method = bilinear

# Default resampling used during interpolation
interpolation_method = linear

# Default distance of nodata spreading during interpolation
interpolation_dist_nodata_spread = half_order_up
```

**Tests**

- validate resampling_method/interpolation_method/interpolation_dist_nodata_spread values + error message if not 
- validate the fact that the value is well setted from the global parameters 
     - resampling_method: reproject/_reproject/merge_rasters/stack_rasters
     - interpolation_method + interpolation_dist_nodata_spread: interp_points, interpolation._interp_points


marinebcht:
hey @belletva, I think it is ok :)

geoutils/interface/interpolation.py
```diff
@@ -91,12 +92,12 @@ def _interpn_interpolator(
...
-    dist_nodata_spread: Literal["half_order_up", "half_order_down"] | int = "half_order_up",
-    method: Literal["nearest", "linear", "cubic", "quintic", "slinear", "pchip", "splinef2d"] = "linear",
+    dist_nodata_spread: Literal["half_order_up", "half_order_down"] | int | None = None,
+    method: Literal["nearest", "linear", "cubic", "quintic", "slinear", "pchip", "splinef2d"] = None,
 ) -> Callable[[tuple[NDArrayNum, NDArrayNum]], NDArrayNum]:
     """
-    Create SciPy interpolator with nodata spreading. Default is spreading at distance of half the method order
-    rounded up (i.e., linear spreads 1 nodata in each direction, cubic spreads 2, quintic 3).
+    Create SciPy interpolator with nodata spreading. Default method is linear.
+    Can be configured with the global setting geoutils.config["interpolation_method"].
```
belletva: maybe add the file : `config.ini`

geoutils/raster/base.py
```diff
@@ -1232,7 +1232,8 @@ def reproject(
         :param nodata: Destination nodata value. If set to ``None``, will use the same as source. If source does
             not exist, will use GDAL's default.
         :param dtype: Destination data type of array.
-        :param resampling: A Rasterio resampling method, can be passed as a string.
+        :param resampling: A Rasterio resampling method, can be passed as a string. Defaults to bilinear.
+            Can be configured with the global setting geoutils.config["reprojection_method"].
```
belletva, line 1236: in file `config.ini`

geoutils/raster/base.py
```diff
@@ -1558,10 +1563,11 @@ def interp_points(
             If points fall outside of image, value returned is nan.
         :param method: Interpolation method, one of 'nearest', 'linear', 'cubic', 'quintic', 'slinear', 'pchip' or
             'splinef2d'. For more information, see scipy.ndimage.map_coordinates and scipy.interpolate.interpn.
-            Default is linear.
+            Default is linear. Can be configured with the global setting geoutils.config["interpolation_method"].
```
belletva, line 1566: in file `config.ini`

geoutils/raster/multiraster.py
```diff
@@ -150,15 +151,19 @@ def stack_rasters(
     :param rasters: List of rasters to be stacked.
     :param reference: Index of reference raster in the list or separate reference raster.
         Defaults to the first raster in the list.
-    :param resampling_method: Resampling method for reprojection.
+    :param resampling_method: Resampling method for reprojection. Defaults to bilinear.
+        Can be configured with the global setting geoutils.config["reprojection_method"].
```
belletva: in file `config.ini`

geoutils/raster/multiraster.py
```diff
@@ -272,7 +277,8 @@ def merge_rasters(
         Defaults to the first raster in the list.
     :param merge_algorithm: Reductor function (or list of functions) to merge the rasters with. Defaults to the mean.
         If several algorithms are provided, each result is returned as a separate band.
-    :param resampling_method: Resampling method for reprojection.
+    :param resampling_method: Resampling method for reprojection. Defaults to bilinear.
+        Can be configured with the global setting geoutils.config["reprojection_method"].
```
belletva: in file `config.ini`

rhugonnet:
@marinebcht @belletva After searching again in my notes, it looks like there is no valid solution to have the function signature keep the default and having a global configuration.

The only options are:
```python
myfunc(arg: bool | None = None)
``` 

like we have, or to be clearer on the fact that "None" in this case represents defaulting to a global config (through a literal or a small dataclass):
```python
myfunc(arg: bool | Literal["config"] ="config")
``` 
```python
myfunc(arg: bool | _UseConfig = USE_CONFIG)
``` 
We can move this bit as an issue, I think?

**In any case, all these options require clarifying the default separately (in the docstring):**
I think we should do this in this PR:
- Describe the default in the argument docstring of every affected function,
- Mention the link to the global config for this argument.

marinebcht:
The recipe I applied for each parameter passed in global config, with the example of the `interpolation_method` : 

Before :  
```python
myfunc(method: Literal["nearest", ....] = "linear", ...)
```

After : 

```python
myfunc(method: Literal["nearest", ....] = None, ...) : 
   """
   ....
   :param method: {old definition} Defaults to "linear" (or Default to "linear"). Can be configured with the global setting 
      geoutils.config["interpolation_method"] (in config.ini).
   """
   # If interpolation method undefined, default to the global system config
   if method is None:
      method = config["interpolation_method"]
```

So if I understood well, we have the description + the link with the config file :) 

For the signature problem, as you want @belletva

rhugonnet:
Perfect. I think I prefer `Defaults to "linear"`. 
I think we can remove the `(in config.ini)` note, as this is just one way (and probably the least common) to use the configuration. Most users will simply set it calling `geoutils.config` at the top of their script, without touching the .ini :wink:.

belletva (approved):
Good for me!

belletva:
@rhugonnet , @adehecq : is it ok for you ? can we merge ?

geoutils/interface/interpolation.py
```diff
@@ -91,12 +92,12 @@ def _interpn_interpolator(
...
     bounds_error: bool = False,
-    dist_nodata_spread: Literal["half_order_up", "half_order_down"] | int = "half_order_up",
-    method: Literal["nearest", "linear", "cubic", "quintic", "slinear", "pchip", "splinef2d"] = "linear",
+    dist_nodata_spread: Literal["half_order_up", "half_order_down"] | int | None = None,
+    method: Literal["nearest", "linear", "cubic", "quintic", "slinear", "pchip", "splinef2d"] = None,
 ) -> Callable[[tuple[NDArrayNum, NDArrayNum]], NDArrayNum]:
     """
-    Create SciPy interpolator with nodata spreading. Default is spreading at distance of half the method order
-    rounded up (i.e., linear spreads 1 nodata in each direction, cubic spreads 2, quintic 3).
+    Create SciPy interpolator with nodata spreading. Default method is linear.
```
rhugonnet: Need to add back the default nodata spread here!

geoutils/interface/interpolation.py
```diff
@@ -91,12 +92,12 @@ def _interpn_interpolator(
...
+    Can be configured with the global setting geoutils.config["interpolation_method"] (or in config.ini).
```
rhugonnet: Here too, it's not just the resampling method, also the nodata spread.

geoutils/interface/interpolation.py
rhugonnet: (This function is not user-facing so not as important, but for us developers to remember too :wink:)

geoutils/raster/base.py
```diff
@@ -1558,22 +1567,34 @@ def interp_points(
             If points fall outside of image, value returned is nan.
         :param method: Interpolation method, one of 'nearest', 'linear', 'cubic', 'quintic', 'slinear', 'pchip' or
             'splinef2d'. For more information, see scipy.ndimage.map_coordinates and scipy.interpolate.interpn.
-            Default is linear.
+            Default is linear. Can be configured with the global setting geoutils.config["interpolation_method"]
+            (or in config.ini).
         :param dist_nodata_spread: Distance of nodata spreading during interpolation, either half-interpolation order
             rounded up (default; equivalent to 0 for nearest, 1 for linear methods, 2 for cubic methods and 3 for
-            quintic method), or rounded down, or a fixed integer.
+            quintic method), or rounded down, or a fixed integer. Defaults to "half_order_up".
```
rhugonnet: "Default" was already written in the line above here, need to remove the redundancy.

geoutils/raster/multiraster.py
```diff
@@ -284,6 +290,10 @@ def merge_rasters(
             merge_algorithm,
         ]
 
+    # If resampling method undefined, default to the global system config
+    if resampling_method is None:
```
rhugonnet: Not sure this is needed here, as it is passed to `reproject` further down anyway?

geoutils/raster/multiraster.py
```diff
@@ -150,15 +151,19 @@ def stack_rasters(
...
     :returns: The merged raster with same CRS and resolution (and optionally bounds) as the reference.
     """
-    # Check resampling method
-    if isinstance(resampling_method, str):
+
+    # If resampling method undefined, default to the global system config
+    if resampling_method is None:
+        resampling_method = _resampling_method_from_str(config["reprojection_method"])
+    elif isinstance(resampling_method, str):
         resampling_method = _resampling_method_from_str(resampling_method)
```
rhugonnet: Same here as below, probably `reproject()` takes care of it already?

rhugonnet (commented):
Added a few comments. I also reiterate that we should remove the "(in config.ini)" for each parameter description: It's not the recommended way of using this in an installed package, it's really an edge case (I checked against how it's done in Matplotlib, Dask, etc).
We should remove it from the documentation page, and only recommend manipulating the `geoutils.config` instead. (It's my fault, I wrote that initially, but didn't know that much about creating `config` files)

marinebcht:
Hello @rhugonnet, thank you for you review :) You convinced Valentine ^^ 
I made all the modifications !

geoutils/interface/interpolation.py
```diff
@@ -91,12 +92,14 @@ def _interpn_interpolator(
...
-    method: Literal["nearest", "linear", "cubic", "quintic", "slinear", "pchip", "splinef2d"] = "linear",
+    dist_nodata_spread: Literal["half_order_up", "half_order_down"] | int | None = None,
+    method: Literal["nearest", "linear", "cubic", "quintic", "slinear", "pchip", "splinef2d"] = None,
 ) -> Callable[[tuple[NDArrayNum, NDArrayNum]], NDArrayNum]:
     """
-    Create SciPy interpolator with nodata spreading. Default is spreading at distance of half the method order
-    rounded up (i.e., linear spreads 1 nodata in each direction, cubic spreads 2, quintic 3).
+    Create SciPy interpolator with nodata spreading. Default method is linear and default spreading is at distance of
+    half the method order rounded up (i.e., linear spreads 1 nodata in each direction, cubic spreads 2, quintic 3).
+    Can be configured with the global respectively setting geoutils.config["interpolation_method"] and
```
rhugonnet: Typo/order issue here maybe? "global respectively setting"?

rhugonnet (approved):
Perfect! Just noticed a typo, then we can merge 🙂

geoutils/interface/interpolation.py
marinebcht: Done :) Thanks @rhugonnet 

But I have this error https://github.com/GlacioHack/geoutils/actions/runs/28858842584/job/85815375415?pr=888 I don't understand why and what has changed since the last run?

rhugonnet: I think this has nothing to do with your PR, looks like an issue with latest versions of Sphinx or plugins to generate the CLI auto-summary.

PR #902, opened 2026-08-18
Add option to Raster.coords to provide reprojected coordinates

### LINKED ISSUE #886: Add option to Raster.coords to provide reprojected coordinates
`Raster.coords` is a convenient tool to directly get the coordinates of each pixel. Sometimes it is also convenient to get the outptut directly in another CRS. We could add a simple option to return the output after reprojection?

marinebcht:
Resolves https://github.com/GlacioHack/geoutils/issues/886

I added the possibility to define an other target CRS `dst_crs` when you want to get coordinates (x,y) of all pixels in the raster using `Raster.coords()` function.

So now, here the function signature : 

image

### DIFF
--- a/geoutils/raster/base.py
+++ b/geoutils/raster/base.py
@@ -1448,7 +1448,7 @@ class RasterBase(ABC):
         grid: bool = True,
         shift_area_or_point: bool | None = None,
         force_offset: str | None = None,
-        dst_crs: CRS | int | None = None,
+        crs: CRS | int | None = None,
     ) -> tuple[NDArrayNum, NDArrayNum]:
         """
         Get coordinates (x,y) of all pixels in the raster.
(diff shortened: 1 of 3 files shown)


geoutils/raster/base.py
```diff
@@ -1448,7 +1444,11 @@ def ij2xy(
         )
 
     def coords(
-        self, grid: bool = True, shift_area_or_point: bool | None = None, force_offset: str | None = None
+        self,
+        grid: bool = True,
+        shift_area_or_point: bool | None = None,
+        force_offset: str | None = None,
+        dst_crs: CRS | int | None = None,
```
rhugonnet: I don't think we need to call it `dst_crs`, because there is no `src_crs` (it is implicitly that of the raster called by `Raster.coords` when working on data object).
So we should simply name this parameter `crs`!

geoutils/raster/base.py
```diff
@@ -1459,12 +1459,18 @@ def coords(
             Defaults to True. Can be configured with the global setting geoutils.config["shift_area_or_point"].
         :param force_offset: Ignore pixel interpretation and force coordinate to a certain offset: "center" of pixel, or
             any corner (upper-left "ul", "ur", "ll", lr"). Default coordinate of a raster is upper-left.
+        :param dst_crs: Coordinate reference system of the coordinates wanted.
```
rhugonnet: "Coordinate reference system in which the coordinates are provided. Defaults to CRS of the input raster."

geoutils/raster/base.py
marinebcht: oh yes ok !

belletva (approved):

PR #878, opened 2026-03-06
Add documentation for Dask/Multiprocessing support and Xarray accessor

rhugonnet:
This PR finally adds the documentation for the Xarray accessor `rst`, as well as the Dask and Multiprocessing support through chunked implementations that we have been steadily developing for the last 2-3 years! :partying_face: 

Thanks in particular to @vschaffn @ameliefroessl for their big contributions at various stages!

Link to the new doc (landing on the Scalability page with schematics that was a bit of work!): https://geoutils-rhugonnet.readthedocs.io/en/add_accessor_daskmp_doc/scalability_logic.html

## Details

The documentation changes are the following:
- Added a whole new **"Scalability"** section in **"Features"** with 4 sub-pages **"Usage and good practices"** (intro with summary and short examples; for those who want to get directly into it), **"Concept definition"** (explanation of scalability concepts with small examples; for those who want to learn), **"Supported operations"** (detailed table of scalability support across methods/object types; for those who want to know scalability exactly), and **"Implementation strategies"** with diagrams and details of implementation (for interested users debugging their memory or who want to learn, or developers/contributors to get a grasp of the code logic),
- Re-worked a page **"Feature and scalability overview"** in **"Getting started"** that is very high-level (no code run, just tables and summaries),
- Added a new page **"Cheatsheet: From GDAL"** + **"Ecosystem"** pages in a new **"Resources"** area (not so happy with the ecosystem page yet, need to buff it up a bit more),
- Edited page **"The georeferenced raster"** to work for both Raster and accessor,
- Removed page **"Implicit lazy loading"** in **"Fundamentals"** (as now covered in new **"Scalability/Concept definition"** page),
-  Updated the **"API reference"** to explain the mirrored API + to have a custom template showing `Raster.method() or ds.rst.method()` for each method linked through a single `RasterBase` call with `RasterBase` itself not being visible (for users using the search button). Also added full `Raster` + `RasterAccessor` full autoclass summary for those also curious to search through the class details. But those two pages are "hidden" in the table-of-content structure to avoid duplicating the main API page (users can only land there through search or clicking the class object name); and the page starts with a link pointing back to the main API, if that's not where they wanted to land.

I haven't yet updated the **"Quick start"**, but I think we **really need a better example there, and we should use the Xarray/Pandas accessor directly**.
Same for the **"Fundamentals"** section, there's some editing to do there so that's it's not too "GeoUtils object-focused", but balanced whether it's about accessors or GeoUtils objects.
[...]

PR #912, opened 2026-08-22
Add Pandas accessors with Dask/MP support, benchmarking, and optimizations

### LINKED ISSUE #491: Lazy loading of vector dataframes?
For a lot of operations (rasterizing, creating mask, point interpolation, and all geometrical operations inherited from Shapely), we mostly require a `Vector`'s geometry columns, but not all of the tabular data associated to it. The data could be loaded implicitly when called?

This could be the same for a `PointCloud`'s auxiliary data from point cloud files, at the exception of coordinates X/Y/Z.

### LINKED ISSUE #712: Add rectangle tiles in multiproc
# Context

Currently, when using the multiproc class, only one dimension is passed. This means that we are only working with square tiles, so this needs to be corrected.

This ticket is under consideration.

### LINKED ISSUE #777: Profiling for Multiprocessing or Dask enabled functions.
Study the behavior of the current Geoutils profiling from https://github.com/GlacioHack/geoutils/pull/751

### LINKED ISSUE #881: Reproject error with Multi and Dask
Here's a small reprojection test I performed, which yields different results in Multi/Dask mode compared to normal mode.

We have a 10x10 DEM with 4 blocks (chunk = 5). The problem appears on the bottom left block (gives the top left block and swaps its lines).

(I commented the check [here](https://github.com/GlacioHack/geoutils/blob/1be54421812a9ecb3cb7fe14a0eb0cfe2c3eb151/geoutils/raster/transformation.py#L784) to pass the `dem = same_dem.reproj()`)

```
src_arrs = np.linspace(0, 99, 100).reshape(10,10)
transform = rio.transform.from_origin(0, 5, 1, 1)
chunk_size = 5
dem = gu.Raster.from_array(src_arrs, transform=transform, crs=4326, nodata=200)

 # Test Normal
dem_reproj = dem.reproject(res=(1,1)) #, mp_config=multiproc_config)
print ("Res normal: \n", dem_reproj.data)
```

Image

```
# Test Multi
multiproc_config=MultiprocConfig(chunk_size=chunk_size,  outfile="tmp/test.tif")
dem_reproj = dem.reproject(res=(1,1), mp_config=multiproc_config)
print ("Res Multiprocess: \n", dem_reproj.data)
```

Image

```
# Test Dask
ds = open_raster("input.tif", chunks={"band": 1, "x": chunk_size, "y": chunk_size})
dem_reproj = ds.rst.reproject(res=(1,1))
dem_reproj = dem_reproj.compute()
print ("Res Dask: \n", dem_reproj.rst.data)
```

Image

### LINKED ISSUE #893: Errors: case of a xarray
From https://github.com/GlacioHack/geoutils/pull/878 : 

## Compute stats 

Some statistics computation give errors : 

```
ds = gu.open_raster(gu.examples.get_path("exploradores_aster_dem"), chunks={"band":1, "x":100, "y": 100})
# or ds = gu.open_raster(gu.examples.get_path("exploradores_aster_dem"))
ds.rst.get_stats("median")  # like 90thpercentile, le90 and nmad
```

=> NotImplementedError: The da.nanmedian function only works along an axis or a subset of axes.  The full algorithm is difficult to do in parallel

```
ds = gu.open_raster(gu.examples.get_path("exploradores_aster_dem"), chunks={"band":1, "x":100, "y": 100})
ds.rst.get_stats("iqr")
```

=> NotImplementedError: Use of `nan_policy` and `keepdims` is incompatible with lazy arrays.

## Create mask

```
ds2 = open_raster(gu.examples.get_path("everest_landsat_b4"), chunks={"band": 1, "x": 100, "y": 100})
inlier_mask = ~Vector(everest_outlines_path).create_mask(ds2.rst)
```

```
  File "/home/mbouchet/Documents/xDem_project/geoutils/bug/test_xarray.py", line 140, in <module>
    inlier_mask = ~Vector(everest_outlines_path).create_mask(ds2.rst)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mbouchet/Documents/xDem_project/geoutils/geoutils/vector/vector.py", line 1557, in create_mask
    return _create_mask(
           ^^^^^^^^^^^^^
  File "/home/mbouchet/Documents/xDem_project/geoutils/geoutils/interface/rasterization.py", line 670, in _create_mask
    return _create_mask_raster(
           ^^^^^^^^^^^^^^^^^^^^
  File "/home/mbouchet/Documents/xDem_project/geoutils/geoutils/interface/rasterization.py", line 571, in _create_mask_raster
    rst01 = _rasterize(
            ^^^^^^^^^^^
  File "/home/mbouchet/Documents/xDem_project/geoutils/geoutils/interface/rasterization.py", line 472, in _rasterize
    dst_chunks = _chunks2d_from_chunksizes_shape(chunksizes=chunksizes, shape=out_shape)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mbouchet/Documents/xDem_project/geoutils/geoutils/multiproc/chunked.py", line 229, in _chunks2d_from_chunksizes_shape
    for i in range(int(np.ceil(shape[0] / chunksizes[0])))
                               ~~~~~~~~~^~~~~~~~~~~~~~~
TypeError: unsupported operand type(s) for /: 'int' and 'tuple'
```

Same error here : 

```
filename_rast = gu.examples.get_path("exploradores_aster_dem")
rast = open_raster(filename_rast, chunks={"band": 1, "x": 100, "y": 100})
rast = rast.rst.crop([rast.rst.bounds.left, rast.rst.bounds.bottom, rast.rst.bounds.left + 5000, rast.rst.bounds.bottom + 5000])
rast_polygonized = rast.rst.polygonize((2500, 3000))
rast.rst.polygonize((2500, 3000)).rasterize(rast.rst)
```

### Reprojection changes dimension (order)

```
rast = open_raster("../../new_xdem/xdem/xdem/example_data/Longyearbyen/data/DEM_1990.tif")
rast2 = open_raster("../../new_xdem/xdem/xdem/example_data/Longyearbyen/data/DEM_2009_ref.tif")
rast_reproj = rast.rst.reproject(crs=rast.rst.get_metric_crs()) => error get_stats()
rast2_reproj = rast2.rst.reproject(crs=rast.rst.get_metric_crs()) => error get_stats()

out_res = (rast.rst.res[0] * 2, rast.rst.res[1] / 2)
rast_reproj = rast.rst.reproject(res=out_res,resampling="bilinear") => error get_stats()
```

Error : 
```
  File "/home/mbouchet/Documents/xDem_project/geoutils/bug/test_xarray.py", line 189, in <module>
    rast_reproj.rst.get_stats()
  File "/home/mbouchet/Documents/xDem_project/geoutils/geoutils/profiler.py", line 328, in wrapper_profile
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/mbouchet/Documents/xDem_project/geoutils/geoutils/raster/base.py", line 773, in get_stats
    data = self.data[band - 1, :, :] if self.count > 1 else self.data
                                        ^^^^^^^^^^
  File "/home/mbouchet/Documents/xDem_project/geoutils/geoutils/raster/xr_accessor.py", line 185, in count
    return self._obj.rio.count
           ^^^^^^^^^^^^^^^^^^^
  File "/home/mbouchet/Documents/xDem_project/geoutils/venvGeoutils/lib/python3.11/site-packages/rioxarray/rioxarray.py", line 996, in count
    extra_dim = self._check_dimensions()
                ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mbouchet/Documents/xDem_project/geoutils/venvGeoutils/lib/python3.11/site-packages/rioxarray/rioxarray.py", line 983, in _check_dimensions
    raise InvalidDimensionOrder(
rioxarray.exceptions.InvalidDimensionOrder: Invalid dimension order. Expected order: ('y', 'x'). You can use `DataArray.transpose('y', 'x')` to reorder your dimensions.
```

### LINKED ISSUE #913: Make profiler relevant to Dask
Something I had been looking at: Our current profiler using `psutil` cannot necessarily capture memory accurately for a Dask cluster. We can simply detect when the cluster is from Dask, use Dask tools for profiling, and return a result for both cases.

Tackling this in #912

### LINKED ISSUE #914: Add ASV for benchmarking
Mentioned in a meeting in a few months ago, this is the norm for benchmarking our tool every few weeks to capture regressions/improvements: https://github.com/airspeed-velocity/asv
Its a minimal setup, I'll propose a small integration in #912 along with benchmarking through our profile tool.

### LINKED ISSUE #915: Make Multiprocessing module self-syntax consistent with Dask
To avoid complexifying things for our users and ourselves, basically an extension of #677 which was a closed a bit early...

### LINKED ISSUE #916: Add Pandas accessors for point cloud and vector
Somehow we forgot to open a formal issue on this despite all the content/discussion in others. Just to keep track. Linked to #383. Working on it in #912. It is much less effort than for rasters, given how close our `Vector` and `PointCloud` objects were from a `gpd.GeoDataFrame` already.

### LINKED ISSUE #917: Add `sieve` functionality for rasters
As discussed in #878

### LINKED ISSUE #918: Add `fillnodata` functionality for rasters
As discussed in #878

### LINKED ISSUE #919: Add support for more `grid()` methods
Adding IDW and circular statistics: I had already been doing this in parallel anyway (I need it for a study), I refactored it with added chunked support in #912 

Kriging will come later after we move variography from xDEM.

rhugonnet:
This PR is a bit big (sorry!), as it implements a lot of co-dependent changes that were more practical to deal with together, notably: new accessors and their Dask support, benchmarking across distributed backends and changes in the backends themselves (Dask/MP), and the performance comparison to GDAL CLI and adding new GDAL-like functionalities (grid options, and sieve/fillnodata).

At its core, this PR adds Pandas accessors `vct` and `pc` mirroring the `Vector` and `PointCloud` classes, homogenizes the Multiprocessing module to have similar name/support than Dask, and updates the Profiling module to also support Dask clusters.

Dask-GeoPandas is added as a new optional dependency to support vector and point cloud accessors, and we extend the LAS reading/writing functionalities to support chunks using existing support in LasPy.
The Vector and PointCloud class are moved to implicit loading with Multiprocessing support (as the Raster class). ?Note that the above depends on file types (some don't support reading only metadata, or reading in chunks).

The PR also adds benchmarking tests in a `benchmarks/` folder that defines [ASV](https://github.com/airspeed-velocity/asv) benchmarks to track performance over package commits, contains a manual CI fail/pass test for large data to check we respect out-of-core execution.
ASV is good to plot commit tag VS single function performance, but not so good to track the dependency of an operation with its parameters. So the `benchmark/` folder also generates a small custom HTML report to visualize the performance with varying parameter ranges (e.g., raster size, interpolation sample size, etc).

The benchmarking results and comparison will be done in a next PR (easier to keep track of main commit once ASV setup, and to share the GitHub page rendering that can only be done once this is merged into `main`!).

As this ties together with the accessor, this PR also updates the point-related operations to support chunked execution. For basic ops, this is trivial and directly supported by Dask/Dask-GeoPandas and slightly modified version of our Multiprocessing module (e.g. reproject is independent between points).
Those are the ones that required some effort:
- Gridding points to a raster (previously not supported),
- Interpolating a raster at points (previously supported Dask only on the raster side, now both!).

The `grid()` function was also refactored to support more algorithms than just Delauney triangulation, adding IDW and circular statistics. And a couple of basic GDAL raster functionalities directly available in Rasterio were wrapped: `sieve` and `fillnodata` (might add our own implementation in the future). We also add `edit` to mirror GDAL CLI, as discussed in #878.

More details below. :smile: 

Resolves #919
Resolves #918
Resolves #917
Resolves #916
Resolves #915 
Resolves #777
Resolves #712 
Resolves #914 
Resolves #913 
Resolves #712 
Resolves #881 
Resolves #893 
Resolves #491 

I also made a quick Wiki page for developers on benchmarking/profiling, so we keep track of the different components!
https://github.com/GlacioHack/xdem/wiki/Things-to-know-on-profiling-and-benchmarking

### New Pandas accessors

This addition is pretty straightforward: Same as for Xarray accessors, just with Pandas. We split modules into a `base.py` with all abstract methods new to the objects, then inherit for both the GeoUtils object (`Vector` or `PointCloud`) into their old module, and inherit for the new accessor in a `pd_accessor.py` module. 
We were composing a `gpd.GeoDataFrame` before, so we can simply apply functions to the same object now. Overall, little changes, mostly restructuring, and mirroring tests that we did for rasters to ensure Dask and loading behaviour are preserved.

*Note: default behaviour between accessors*

For now, a GeoUtils object always return a GeoUtils object, and accessed object returns another accessed object.
For example, for `to_pointcloud()`: a `Raster` returns a `PointCloud`, and a `xr.DataArray` returns a `gpd.GeoDataFrame`. Pretty intuitive, I think.

### Loading behaviour of  `Vector`/`PointCloud` classes

We add a `load()` function to both of these classes, to mirror the behaviour of `Raster`, with no data loading by default (only metadata) if the file format allows (some don't for vectors). This also allows to naturally use Multiprocessing for chunked functions, mirroring what is done in the Raster class.

### Multiprocessing API homogenization with Dask

The functions/arguments of functions in the `multiproc/` module are renamed to match exactly those of Dask, with the caveat that there is no laziness so the behaviour is not always an exact match but still pretty close. The objective is to avoid unnecessary confusion and complexity; for example: `chunks/chunksizes` were mixed, `get_res/compute`, `map_overlap/block`, and more...

### Profiler support for Dask

I consolidated the profiler to support Dask profiling. I first searched and found that `psutil` did not encourage the checking of children tasks for scenarios such as Dask clusters, as it could drop memory usage profiling depending on how those evolved. So we instead use Dask's `MemorySampler` instead, if we detect that a `dask.distributed` client is running.

### Benchmarking tests

We add ASV for long-term, practical benchmarking into a new `benchmarks/` folder. The workflows run all operations for a given set of parameters.
We also add a workflow that tests core computation-intensive functions (reproject, grid, interp_points, etc) across a range of parameters that influence the performance (e.g. raster or point size).
Because ASV only create plots of "Commit-number/Time" (X axis) vs "Performance metric" (Y axis), that aren't super easy to read/interpret for users (remember, we weren't convinced during our dev meeting), I made the effort to add a custom renderer to add a "Summary" page, an "Options" page (different backends/methods) and a "Scaling" page that varies raster/point/vector size and generates plots of "Parameter range" (X axis) vs "Performance metric" (Y axis). They all use the same ASV benchmark result, simply plot them in more interpretable ways for users, and I kept the "Benchmark history" ASV native page for developers.
Finally, a new test module `test_large_data` that reuses the same setup as the ASV definitions is run in the CI to test that Dask/MP functionalities indeeds respects bounded memory.

### LasPy chunked support

I moved the LasPy-specific routines out of the `pointcloud` module into a new `las` module, and refined them to use chunk reading/writing support from LasPy. We can now manipulate those point clouds lazily and out-of-memory! For MP usage, it uses the read/write helpers. For Dask, it converts into a lazy dataset.

### New options and operations

This PR adds new `grid` options using SciPy and Numba, and the `edit`, `sieve` and `fillnodata` functions we discussed in #878. These last ones were little effort, as for now they directly wrap Rasterio.
For gridding, I had been working on implementing those separately for a study on DEM coregistration, so I brought those in this PR directly to compare to GDAL with the existing workflow.
The PR adds IDW interpolation and circular-neighborhood statistics following GDAL’s `gdal_grid` definitions [here](https://gdal.org/en/stable/programs/gdal_grid.html). SciPy implementations uses existing spatial trees to find neighbouring points, while Numba uses some simple point-to-grid accumulation loops. 
These are the simplest implementation, we don't add Numba support for linear/cubic Delauney interpolation (equivalent to SciPy's `griddata`), as this would be a big beast to tackle! We'll also add kriging later, once we migrate variograms from xDEM!
As always, we check Numba vs SciPy to ensure the output result is the same (as for xDEM's terrain attributes), and we now also compare their performance in the new benchmarks! (hence the addition directly in this PR)


doc/source/multiprocessing.md
```diff
@@ -99,30 +99,30 @@ This function applies a function to raster tiles and **returns a list** of extra
...
 ---
 
 ## Choosing the right function
 
-| Use case                                      | Function                                                               |
-|-----------------------------------------------|------------------------------------------------------------------------|
-| Apply processing and save results as a raster | {func}`~geoutils.multiproc.map_overlap_multiproc_save`                 |
-| Extract statistics or features into a list    | {func}`~geoutils.multiproc.map_multiproc_collect`                         |
-| Track tile locations with extracted data      | {func}`~geoutils.multiproc.map_multiproc_collect` with `return_tile=True` |
+| Use case                                      | Function                                                                     |
```
marinebcht, line 124: I do not find any explations (or sig link in this page) of the map_overlap / map_blocks  functions

geoutils/raster/base.py
```diff
@@ -463,6 +505,52 @@ def set_area_or_point(
...
+        :param nodata: New nodata metadata value without changing pixel values or their mask.
+        :param tags: Metadata tags to update, or ``None`` to clear all tags.
+        :param area_or_point: New pixel interpretation without shifting the geotransform.
+
+        :return: Raster copy with updated metadata.
+        """
+
+        # A shallow copy keeps the unchanged pixel storage unloaded and separates metadata edits
+        raster_copy = self.copy(deep=False)
+        edited = raster_copy.rst if isinstance(raster_copy, xr.DataArray) else raster_copy
```
marinebcht, line 535: Maybe we could do this in some other functions ? I still forget the .rst too often ^^

benchmarks/README.md
```diff
@@ -0,0 +1,92 @@
+# GeoUtils benchmarks
+
+This directory contains repeatable performance measurements and pass/fail large data tests.
+
+## Organization
+
+- `workflows/` defines deterministic inputs, operation methods, calculation engines, chunk strategies, execution modes
```
marinebcht, line 7: @belletva we need to do the same thing for the accuracy/topo benchmarking

doc/source/benchmarking_index.md
```diff
@@ -0,0 +1,12 @@
+(benchmarking)=
+# Benchmarking
+
+The following sections present GeoUtils' **performance benchmarks** and its **profiling tool** available
+to measure time and memory on your own workflows.
```
marinebcht, line 5: "measure execution time and memory usage" for the lambda people 
tools ?

doc/source/benchmarking_performance.md
```diff
@@ -0,0 +1,53 @@
...
+Dask and Multiprocessing, track **improvements or regressions over time**, and compare different **execution modes** and
+**calculation engines** against the [**GDAL CLI**](https://gdal.org/en/stable/programs/index.html). Chunked operations
+also compare strategies used to coordinate selections or reconcile results across chunks.
+
+The benchmarks use GeoUtils' profiling tool described in {ref}`profiling`, which relies on [**psutil**](https://psutil.readthedocs.io/) and [**Dask's worker memory diagnostics**](https://distributed.dask.org/en/stable/worker-memory.html).
+
+## Improvements or regressions over time
+
+The [**GeoUtils benchmark webpage**](https://glaciohack.github.io/geoutils/) provides performance of core functionalities and their changes with commit history.
+It relies on [Airspeed Velocity (ASV)](https://asv.readthedocs.io/) to record fixed performance measurements for each commit and publish them.
```
marinebcht, line 14: reproducible more that fixed ?

doc/source/benchmarking_performance.md
```diff
@@ -0,0 +1,53 @@
+(benchmarking-performance)=
+# Performance
+
+GeoUtils benchmarks its functionalities to measure **RAM and wall time**, check **lazy and out-of-core behaviour** with
```
marinebcht, line 4: RAM usage

doc/source/benchmarking_performance.md
```diff
@@ -0,0 +1,53 @@
...
+
+Two reference graphics summarize the core results:
+
+- **End-to-end time relative to GDAL** for each comparable operation
+- **Peak RAM as raster size increases**, including the full process and its workers
+
+:::{figure} https://glaciohack.github.io/geoutils/documentation/time_relative_to_gdal.svg
+:alt: End-to-end GeoUtils execution-mode time relative to GDAL for four raster operations
+
+End-to-end time on the largest raster size shared by every execution mode and the GDAL CLI. GDAL is the reference at one.
```
marinebcht, line 34: "the reference at one" ? as the baseline

doc/source/benchmarking_performance.md
```diff
@@ -0,0 +1,53 @@
...
+It relies on [Airspeed Velocity (ASV)](https://asv.readthedocs.io/) to record fixed performance measurements for each commit and publish them.
+
+## Comparison across execution modes, engines and GDAL CLI
+
+GeoUtils compares **eager, Dask and Multiprocessing** execution using the same deterministic inputs. Reprojection,
+polygonization, rasterization and gridding additionally use equivalent **GDAL CLI commands** with matching
+output grids, data types and storage options where possible. Every result identifies its operation, numerical method,
+calculation engine, chunk strategy and execution mode separately. Engine comparisons use eager execution, while
+execution-mode and chunk-strategy comparisons fix the other dimensions. Fixed-size Numba checks additionally exercise
+each compiled gridding kernel in Dask and Multiprocessing workers. The Rasterio/GDAL calculation engine used inside
```
marinebcht, line 23: Do we we capitalize the first letter of Multiprocessing ? (ex: not done here benchmarks/test_large_data.py)

doc/source/benchmarking_performance.md
```diff
@@ -0,0 +1,53 @@
...
+End-to-end time on the largest raster size shared by every execution mode and the GDAL CLI. GDAL is the reference at one.
+:::
+
+:::{figure} https://glaciohack.github.io/geoutils/documentation/peak_ram_by_raster_size.svg
+:alt: Peak process-tree RAM by raster size for GeoUtils execution modes and GDAL
+
+Peak RAM for the benchmark process and all execution-mode workers as raster dimensions increase.
+:::
+
+These graphics show the latest complete CI benchmark and may be newer than this documentation version.
```
marinebcht, line 43: I do not understand why

doc/source/benchmarking_performance.md
```diff
@@ -0,0 +1,53 @@
...
+[Download the exact values and run metadata](https://glaciohack.github.io/geoutils/documentation/benchmark_snapshot.json).
+
+## Test suite for scalable execution and large datasets
+
+Every GeoUtils operation advertised as {ref}`chunked or lazy <scalability-support>` is tested
+on all supported Python versions and operating systems for its respect of Dask laziness, deferred I/O, and loading behaviour,
```
marinebcht, line 49: "for" ? to ensure that ... ?

doc/source/benchmarking_performance.md
```diff
@@ -0,0 +1,53 @@
...
+while yielding **exactly** the same output as in-memory operations.
```
marinebcht, line 50: equivalent

doc/source/benchmarking_performance.md
```diff
@@ -0,0 +1,53 @@
...
+
+In addition, GeoUtils also includes large data tests running on the latest Python and Ubuntu,
```
marinebcht, line 52: Ubuntu release

doc/source/benchmarking_profiling.md
```diff
@@ -0,0 +1,142 @@
...
+The same measurements support the controlled comparisons presented in {ref}`benchmarking-performance`.
+
+```{note}
+The profiling functionalities rely on [psutil](https://psutil.readthedocs.io/en/latest/) and [plotly](https://plotly.com/) as optional dependencies. You can install them manually or with ``pip install geoutils[opt]``
+```
+
+## Profiling a single function call
+
+For a one-off measurement, `profile_call()` runs a function and returns both its output and normalized metrics for
+runtime and memory. The current Python process is measured with psutil. If a distributed Dask client is active,
```
marinebcht, line 26: "execution time and memory usage"
I don't know to what extent that we say "runtime and memory" by "abus de language"

marinebcht: `psutil`

doc/source/benchmarking_performance.md
```diff
@@ -0,0 +1,53 @@
...
+execution-mode and chunk-strategy comparisons fix the other dimensions. Fixed-size Numba checks additionally exercise
+each compiled gridding kernel in Dask and Multiprocessing workers. The Rasterio/GDAL calculation engine used inside
+GeoUtils remains distinct from the external GDAL CLI reference.
+
+Two reference graphics summarize the core results:
+
+- **End-to-end time relative to GDAL** for each comparable operation
+- **Peak RAM as raster size increases**, including the full process and its workers
+
+:::{figure} https://glaciohack.github.io/geoutils/documentation/time_relative_to_gdal.svg
```
marinebcht, line 31: I'm not sure the ---- for GDAL CLI is needed
Was confused at first

doc/source/multiprocessing.md
rhugonnet: It's the same function renamed? I'm not sure I understand

geoutils/raster/base.py
rhugonnet: Unfortunately this is only relevant here because we need a shallow copy that returns an object that is not necessarily a Raster, but could also be a Xarray DataArray. For other functions, we usually don't have this issue because we can use `from_array` at the end (create a new raster from scratch).

doc/source/benchmarking_performance.md
rhugonnet: We link the plot from RTD to the GitHub pages, which are updated at every commit. So they might be newer that the "stable" doc page. We can't really maintain easily a big benchmark run that updates in the doc itself.

doc/source/benchmarking_performance.md
rhugonnet: This is to have a reference to benchmark against.
When we use GDAL under-the-hood through Rasterio, we check that Rasterio wrappers or GeoUtils (through additional logic or Dask/MP support) doesn't add any unreasonable overhead.
We we don't use GDAL (SciPy/Numba instead) for a functional that GDAL supports, we can compare the performance.

doc/source/multiprocessing.md
marinebcht: There is no link for each "{func}`~geoutils.multiproc.xx`"

PR #933, opened 2026-09-12
Documentation before `v0.3` release

rhugonnet:
This PR adds the latest bits of documentation given the addition/updates of features before a `v0.3` (pre-1.0) release, which are:
- Addition of Pandas accessors `pc` and `vct` for point clouds and vector, with Dask support,
- Addition of lazy behaviour for Vector/PointCloud classes, with Multiprocessing support,
- Extension of LAS reading/writing in chunks through LasPy to work with the two above,
- Migration and improvement of features from xDEM: grouped statistics, variography, cosampling, pairsampling,
- Extended support for chunked execution (Dask/MP) for easy cases that derived from other functions (e.g., `clip` from `create_mask`, `to_pointcloud` from `subsample()`, etc).
- Addition of benchmarking of all features.
- Renaming/consolidation of most functions (clip versus crop, bbox versus bounds) and full mirroring of GDAL CLI overhaul!

Resolves #907
Resolves #940 
Resolves #631 
Resolves #568 
Resolves #395 

Ongoing

PR #938, opened 2026-09-13
Improve Dask/MP support in conversion to point cloud, subsampling and pair sampling

### LINKED ISSUE #707: Make output of `subsample()` a point cloud
Whether for a raster or a point cloud, the `subsample` operation always returns data that could be contained in a point cloud.
Should we add this return type? Should we even make it default?

Right now subsample returns either the subsampled `.data`, or the index used to subsample `.data` (2D for rasters, 1D for point cloud).

Maybe it would also be the occasion to remove the `subsample` argument from `Raster.to_pointcloud` and re-use the behaviour there (which uses out-of-memory indexing to get the raster subsample):   
https://github.com/GlacioHack/geoutils/blob/e79a1ff8948709fda445835f30cf7dae649e046b/geoutils/interface/raster_point.py#L116
I remember this issue coming up when finalizing `to_pointcloud`.

### LINKED ISSUE #929: Make `to_pointcloud` conversion work on chunks
All in the title, both for Dask and Multiprocessing.

### LINKED ISSUE #937: Add Multiprocessing support to pair sampling
Only feature where we don't have parity Multiproc/Dask, as I had only worked on the Dask implementation in xDEM.

### LINKED ISSUE #939: Add support for out-of-memory pair sampling for point clouds
Right now only supported on rasters. Point clouds pair sampling implements a similar log-lag sampling, but not the Dask/MP support.
Should not be too much work to add.

rhugonnet:
This PR adds:
- Multiproc support to **raster pairsampling** (Dask was already supported, re-used the logic), mirroring the xDEM implementation migrated in #925 
- Dask/MP support to **subsampling** and **raster to point cloud conversion**, which was the bulk of the work (subsampling already supported reading by chunks, but not the logic to determine output samples in chunks or to write to a point cloud file this way),
- Dask/MP support to **point pairsampling** (before was loaded in memory), which can reuse some of the same logic and subfunctions as the two above.

I started separately and realized all those shared logic, so I ended up grouping them together in this PR.

I also realized that `Raster.subsample()` and `Raster.to_pointcloud()` were doing a lot of things in parallel, so I used the opportunity to clarify the boundary between the two:
- `subsample()` now returns a point cloud by default, or an 1D array of values or 2D indices for `as_array=True` and `return_indices=True`, and is simply wrapped by `to_pointcloud(subsample=)`,
- `to_pointcloud()`  also returns a point cloud file or a MxN array (M=X/Y + number of bands, as it supports a multi-band raster).

They are basically doing the same thing underneath, except that `to_pointcloud()` signature is conversion-oriented, while `subsampling` is subsample-oriented. 

So most of the chunked logic is specific to `subsample()`, and just reused in `to_pointcloud()`. In particular, I added a `skip_nodata` option to `subsample()` (before was only valid values), so that all options are supported consistently.

The most technically challenging part of this PR was to add **output chunk support** for subsampling/conversion to point cloud:
- Input chunks were already supported: we read raster tiles chunk by chunk to gather the subsample.
- But, if the user-requested **a sample size S** (e.g., 10,000,000) **larger than a raster chunk NxN** (e.g., 1000 x 1000 = 1,000,000), then underlying operations still worked on the S-sized sample size (both for choosing which valid indices to subsample, then for returning the final array).

Now, by using an iterative algorithm that can select the key cutoff for the "topk" method chunk by chunk (before required the full array in memory), we can limit the memory to a single raster chunk. 
Then, we use this cutoff to read input raster chunk by chunk and stream the sampled values to a partitioned point cloud file (for Multiproc) or a lazy GeoDataFrame (for Dask), so the output is also chunked!
This simple algorithm is described in the literature, and I added sources in the code (I don't think it's often cited, for example in Dask's topk implementation it is not: https://docs.dask.org/en/latest/generated/dask.array.topk.html).

I used the opportunity to add a comparison to PDAL in benchmarking for the point cloud conversion/subsampling. It's looking good!
I want to further polish the new benchmark module and its rendering in a separate PR, now that almost all functionalities are supported by Dask/MP.

Resolves #929 
Resolves #937 
Resolves #939 
Resolves #707

### DIFF
--- a/geoutils/pointcloud/base.py
+++ b/geoutils/pointcloud/base.py
@@ -741,6 +741,7 @@ class PointCloudBase(VectorBase):
         nn_max_batches: int = 200,
         index_dtype: DTypeLike = np.int32,
         distance_dtype: DTypeLike = np.float32,
+        mp_config: MultiprocConfig | None = None,
     ) -> xr.Dataset:
         """Sample finite point pairs for statistics by distance.
 
@@ -748,7 +749,7 @@ class PointCloudBase(VectorBase):
         and accepts a nearby observed endpoint, which is generally faster for large point clouds.
 
         Strategy controls apply to ``"loglag"``. ``"random_xy"`` uses ``max_rounds`` and ``nn_batch_size``.
-        Dask point tables are loaded because the search requires all coordinates.
+        Dask and Multiprocessing point tables are collected because the search requires all coordinates.
 
         :param n_pairs: Requested number of pairs with two finite values; fewer may be returned if sampling stops early.
         :param sampling: ``"loglag"`` balances short and long distances on a log scale; ``"random_xy"`` draws
@@ -777,6 +778,8 @@ class PointCloudBase(VectorBase):
         :param nn_max_batches: Maximum batches to fill the sample with ``"nn_logvector"``.
         :param index_dtype: Integer NumPy dtype for returned row indexes (e.g. ``"int64"`` for very large point clouds).
         :param distance_dtype: Floating NumPy dtype for returned distances (e.g. ``"float64"`` for greater precision).
+        :param mp_config: Worker and row partition settings for reading an unloaded point cloud. Cannot be combined
+            with Dask inputs. The global point search and returned Xarray Dataset are eager.
         :returns: Xarray Dataset with pair and endpoint dimensions, containing original row indexes, values,
             coordinates, and distances.
         """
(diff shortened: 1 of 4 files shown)


rhugonnet:
@adehecq @belletva @marinebcht This PR is done. We need it to finalize https://github.com/GlacioHack/xdem/pull/759, so I'm merging now. Will highlight changes at meeting + document in #933

adehecq:
Looking all good to me, not much to add! Thanks! 🙌

PR #943, opened 2026-09-18
Add plotting with downsampling for large datasets

### LINKED ISSUE #447: Sampling interval is irregular when loading Raster with downsampling option

### LINKED ISSUE #485: Simplify setting plot limits
Currently, if one plots several layers on top of each other, the limits of the last layer will be used. This can be annoying if one is much larger than the other. Here is an example (taken from xDEM) that has annoyed me a lot...
```
import xdem
import geoutils as gu

ref_dem = xdem.DEM(xdem.examples.get_path("longyearbyen_ref_dem"))
outlines = gu.Vector(xdem.examples.get_path("longyearbyen_glacier_outlines"))

ref_dem.show()
outlines.show(fc="none", ec="k")
plt.show()
```
The final figure looks like this. It's simply impossible to see the raster. To force the limits to the raster, I always reverse the order of the commands and play with matplotlib's zorder to have them in the right order (which is easier than using `ax.set_lim`).
Figure_1

What would be a simpler way to do this?
Should we have a function `set_plot_lim` or something alike that takes either a `Raster` or a `Vector`, even possibly 4 coordinates to set the plot limits? Or we could have an option to set the limit to the intersection and/or union of the bounds?
Were should such a function sit? We don't have a "plot" submodule, but we could create one with a few tools to facilitate plots, maybe interfaced with cartopy...

### LINKED ISSUE #585: Padding of colorbar in `Vector.plot()` varies despite fixed `pad` in `cax.append_axes`
I also tried passing a {"pad": "2%"} value to the `legend_kwgs` that are apparently passed to `colorbar`, without success...
The padding created by: 
```python
divider = make_axes_locatable(ax0)
cax = divider.append_axes("right", size="5%", pad="1%")
```
 seems to generally work, except when `column` is used...

If we can't manage to make it work by passing an argument or defining the `cax` in advance, the other solution would be to re-create the colorbar after the plotting. But then we'd need to extract the vmin/vmax, other `legend_kwds` associated to colorbars (like we already do a bit), and it could rapidly become complex.

### LINKED ISSUE #641: Forced `interpolation="None"` in `plot()` to avoid NaN artefacts causes large vectorized file size

### LINKED ISSUE #711: Plot - memory on disk
# Context

The purpose of this ticket is to implement the creation of PNG plots out of memory during the process.

This ticket is under consideration.

### LINKED ISSUE #792: Add `downsample` to `PointCloud` object to mirror `Raster` functionality
I think this could be a useful functionality, and would help stay consistent for elevation point clouds inputs in the CLI for example.
@adebardo @belletva @adehecq

### LINKED ISSUE #799: Default plotting renders weirdly for geographic CRS
Problem with colorbar spacing and default size...
This is for `PointCloud.plot()`:

Image

### LINKED ISSUE #901: Allow setting aspect in Raster.plot
Currently, Raster.plot forces the aspect to be "equal" (see [here](https://github.com/GlacioHack/geoutils/blob/0d5df8dc46294d9c0405151960db71984b1d0f84/geoutils/raster/raster.py#L2201)). While this makes sense in most cases, one may want to set a different aspect (e.g. when plotting data in lat/lon).
A quick fix is to check whether "aspect" is set in the kwargs, and use that in priority, otherwise revert to the default "equal".
One can also easily overwrite it with `ax.set_aspect()` afterwards, so this is not a big issue.

rhugonnet:
This PR makes plotting scalable for large datasets by allowing downsampling out-of-memory (both for accessors and Raster/Point clouds objects).
Structurally, it moves plotting functionalities to their own `plotting.py` modules, and the `plot()` function to `RasterBase` or `PointCloudBase` (before was implemented separately in accessor and GeoUtils object) for consistency.

In practice, the downsampling uses either `subsample()` (for point clouds) or `reproject()` (for rasters) that already support out-of-memory resampling thanks to our existing implementations, so we simply link to those.
To control it, the `plot()` function has a new `max_points` arguments which defaults to an `"auto"` mode which accounts for the size of the plot canvas (computed from figure size + DPI), or fallbacks to a maximum number of points. 

As the above joins an issue we had on adding a `downsample=` argument for point clouds for consistency with rasters, I added this as well. Also adding it to `open_raster()` for the accessor, so that downsampling is consistently treated/supported everywhere! :slightly_smiling_face: 

I also used the opportunity to fix a list of small issues associated with plotting, as most of them were fairly straightforward, details further below.

Resolves #711 
Resolves #792 
Resolves #901 
Resolves #485 
Resolves #799 
Resolves #641 
Resolves #585 
Resolves #447

## Explanations for small issues

For #901: `Raster.plot()` was forwarding Matplotlib options through kwargs AND passing aspect="equal" internally, causing an error when `aspect` was passed by a user. We now use `kwargs.setdefault("aspect", "equal")`, so aspect="auto" or another value can override it cleanly.

For #485: We extend match-referencing: when receiving a georeferenced object, plotting can now use both that object’s CRS and its bounds for the axis limits. Passing just a CRS still works. So we rename the argument as `ref` (before was `ref_crs`).

For #799: Point cloud colorbars were previously weird because GeoPandas adjusted the axes for a geographic CRS. We now create colorbars through GeoPandas after the map aspect is calculated, fixing its position.

For #641: Fixed thanks to the downsampling above. Matplotlib’s configured interpolation is now respected, but only after the downsampling from us.

For #585: Was the same cause as #799! Now fixed as well.

For #447: Our raster `downsample` previously used `ceil` to round with output dimensions. This made Rasterio spread windows over a grid that did not correspond to exact intervals. We now uses direct intervals to read the exact source window. For example, a 100 x 100 raster downsampled by 6 produces 16 x 6 columns sampled consistently every 6 pixels, which omits the incomplete trailing chunk at the end.


geoutils/pointcloud/base.py
```diff
@@ -203,16 +215,86 @@ def point_count(self) -> int:
...
+        savefig_fname: str | None = None,
+        *,
+        max_points: Literal["auto"] | int | None = "auto",
+        random_state: int | np.random.Generator | None = 0,
+        **kwargs: Any,
+    ) -> None | tuple[matplotlib.axes.Axes, matplotlib.axes.Axes | None]:
+        """
+        Plot the point cloud.
+
+        This method performs automatic subsampling to facilitate the plotting of large datasets
```
adehecq: Just an open question. Should the subsampling be done silently (but many users will be unaware of that feature at first and may be puzzled), or should we print a statement when subsampling is done?
I would be in favor of printing a statement, but I know some people find it annoying.

adehecq:
Thanks! This is a great feature to have and it's nice to have all these small issues fixed!
I don't have many comments apart from the small question above.

Issue #927, opened 2026-09-07
Solidify namings before upcoming release


rhugonnet:
There are a couple names in GeoUtils that are not ideal. We need to decide on those for longevity before the main release.

Those are:

1. `Vector.create_mask()`

The naming is quite unclear: What mask? The word 'create' doesn't really enlighten much?
I think we should rename it to `geometry_mask()`, which is symmetric with Rasterio.
(Note: `mask_geometries` doesn't work, because we're not masking the geometries, but the raster/pc values WITH the geometries)

2. `Raster/PointCloud.get_stats()`

The word "get" is typically use in software for internal method that retrieve object attributes/data. For instance `get_mask` and `get_array` for a NumPy masked array. They are not supposed to be used for the **computation** of a new output, and often not for public methods.

My favorite would be simply `stats()`.
Pandas uses `describe()` also for a stats summary, but the function is not meant to be expanded (simple descriptive stats), and I think the method escapes many users because of that technical name. We could have it duplicated as `describe()`/`stats()`.

Actually, I would even put everything into `stats()`, including currently drafted `grouped_stats(), to have a single consistent API like this:
```python

# Summary for entire raster
rast.stats()
rast.stats("mean")
rast.stats(["mean", "std", "nmad"])

# Three modes of grouped stats
# 1/ Categorical statistics
rast.stats(["mean", "std"], by=landcover)

# 2/ Binned statistics
rast.stats(["mean", "std"], by=elevation, bins=np.arange(0, 5000, 100))

# 3/ Zonal statistics
rast.stats(["mean", "std"], by=glacier_outlines)
```

3. `Raster.get_nanarray()`

This only exists for the `Raster` object, not the accessor (always a NaN-array in the associated `xr.DataArray`).
Same problem as above: We're not getting, but converting a masked array into a nanarray. I think `to_nanarray` might be better.

4. `Raster/Vector.crop(clip=True)`

Right now this is inconsistent with GDAL.
I think we need to keep `crop()`, as it behaves differently from `clip`.
For rasters: `crop()` cuts to the closest pixel for the bounding box, `clip` actually create nodata within the geometry used.
Big difference in terms of execution: crop can be done as deferred I/O (only load specific chunks or geometry partitions), while clip modifies data.
For vectors: `crop()` cuts to intersecting geometries only, `clip` actually clips the geometries.

@belletva @adehecq @atedstone @marinebcht Thoughts?
Any other naming I might have missed?

[...]

PR #949, opened 2026-09-20
Separate `crop()` and `clip()`, and add chunked support

### LINKED ISSUE #570: Add option to `Vector.crop()` for removing all features not contained (instead of intersecting)

### LINKED ISSUE #583: Raster clipping
It would be great to have a function to clip a raster to a vector. Currently, the function `Raster.crop` if passed a vector can crop to the bounds of the vector, but it will not clip (set values to nodata outside the vector outlines). 
`Vector.crop` has an option "clip" to do that. Maybe we could just add a similar "clip" option to `Raster.crop` that works when the input is a vector?

### LINKED ISSUE #852: Make `crop()` function not load output for `Raster`
This is the only remaining difference between Xarray accessor and Raster behaviour.
For a Raster, we don't load the input, but we return a loaded output array, which is unnecessary. We can fix it by adding internal logic in the `Raster` class, for instance defining `_out_window` at the end of `crop()` (which will be later passed to `_load_rio()` when `.data` is called).
This would also allow to use single consistent loading function everywhere (currently different in `crop()`).

### LINKED ISSUE #948: Add chunked support to clipping rasters/vectors/points
Proper clipping with geometries (not cropping, which is already lazy with deferred I/O) as defined in #927

rhugonnet:
This PR separates `crop()` and `clip()` as we discussed here: #927.

It also adds chunked (Dask/MP) support for `clip()`: This is fairly easy, given that it derives from `create_mask()`/`rasterize()` for which we already have chunked support for rasters and point clouds since quite a while.

We also add logic to support deferred I/O for `crop()` in GeoUtils objects consistently (`Raster`, `Vector`, `PointCloud`). Before, we did read only the necessary chunks (without loading the whole array) but didn't defer the operation. To add this is straightforward: We store the cropping window to read for the later implicit `load()` call. The logic is slightly different for vector and point cloud (we save both the window and the type of cropping, see directly below), but very similar.

Finally, we add support for two type of "vector cropping" as this is ambiguous (as noted in #570). As a reminder: cropping does not touch underlying data (geometries for a vector), so it can have several definitions for a vector (while for points/raster, we just remove all data outside the bounding box, easy). Thus, we have `mode="intersects"` and `mode="within"`.

We had an unexpected large number of issues related to this, which I tried to tackle together.

Resolves partly #927 (not closed yet)
Resolves #852 
Resolves #583 
Resolves #570 
Resolves #948


geoutils/pointcloud/base.py
```diff
@@ -1014,6 +1031,37 @@ def variogram(
...
+        self: PointCloudBaseType,
+        mask: Any,
+        keep_geom_type: bool = False,
+        sort: bool = False,
+        mp_config: MultiprocConfig | None = None,
+    ) -> PointCloudBaseType | gpd.GeoDataFrame:
+        """
+        Remove points outside an exact clipping geometry.
+
+        :param mask: Clipping geometry, vector, point cloud, raster or bounding box. Georeferenced masks are
```
adehecq: If mask is a geometry or Vector, this is easy. 
If this is a raster, I think one needs to clarify which pixels are used to keep or exclude a point. Do we use non 0 pixel values?
If mask is a point cloud, I don't understand what clip will do...

geoutils/pointcloud/base.py
```diff
@@ -1014,6 +1031,37 @@ def variogram(
...
+            reprojected to this point cloud's CRS.
+        :param keep_geom_type: Whether to remove intersections with a different geometry type than the input.
```
adehecq: I don't understand this argument.

geoutils/pointcloud/base.py
```diff
@@ -1014,6 +1031,37 @@ def variogram(
...
+        :param sort: Whether to sort clipped points by their original index within each partition.
```
adehecq: What is the default sorting, if not the same as the original one?

geoutils/pointcloud/base.py
```diff
@@ -1014,6 +1031,37 @@ def variogram(
...
+        :param mp_config: Worker configuration with an integer number of points per chunk. The output format is
+            inferred from ``outfile`` or selected by ``driver`` (``GPKG``, ``LAS`` or ``LAZ``), defaulting to
```
adehecq: I'm still not very familiar with the multiproc functionalities. But this argument is like a dictionary, which can contain the key "outfile"? If outfile is set to a valid path, then the output is directly saved on disk?

geoutils/raster/base.py
```diff
@@ -1411,6 +1426,30 @@ def crop(
...
+        all_touched: bool = False,
+        mp_config: MultiprocConfig | None = None,
+    ) -> RasterType:
+        """
+        Mask raster cells outside an exact clipping geometry.
+
+        The raster grid and extent stay unchanged. Cells inside the geometry keep their values, while cells outside
+        become masked values for Raster objects and NaN values for Xarray objects.
+
+        :param mask: Clipping geometry, vector, point cloud, raster or bounding box. Georeferenced masks are
```
adehecq: Same comment as for pointcloud, the behavior for mask = raster or pointcloud is ambiguous or unclear.

geoutils/vector/base.py
```diff
@@ -396,62 +415,158 @@ def active_geometry_name(self) -> str:
...
+        sort: bool = False,
+        mp_config: MultiprocConfig | None = None,
+    ) -> VectorBaseType | gpd.GeoDataFrame:
+        """
+        Clip geometries exactly to a mask.
+
+        Vector geometries are intersected with the mask. Point cloud rows outside the mask are removed. Unlike
+        crop(), this operation can change geometry data and therefore reads an unloaded source.
+
+        :param mask: Clipping geometry, vector, point cloud, raster or bounding box. Georeferenced masks are
```
adehecq: Same comments for these 3 arguments.

adehecq (approved):
Happy with the implementation as described in the PR description!
In terms of code, I only looked at the high level functions and documentation. My comments are mostly regarding clarifications.

